PR #12, opened 2020-11-06
Added semver module to prod dependencies for npm

larstobi:
Fixes `Error: Cannot find module 'semver'` when starting the docker built container.


larstobi:
Fixed here  https://github.com/zwave-js/node-zwave-js/commit/79bfc6f2c95aa9dcc86dc24f7c72239977945662

PR #19, opened 2020-11-16
[WIP] Migrate JavaScript to Typescript using ts-migrate

ahochsteger:
## About

This is a proof-of-concept for automatically migrating all JavaScript files to Typescript using ts-migrate.
It is still work in progress and not yet ready to be merged.

## Migration steps

These were the steps done on the original repo to do the migration:

```bash
npm install --save-dev ts-migrate
npx ts-migrate-full .
```

## Run migrated files

To run the migrated files the *.ts files have to be compiled to *.js first:

```bash
tsc # Don't worry about compile errors for now
# Console 1:
npm dev:server
# Console 2:
npm dev
```

## Open issues

* .js files should not be generated next to .ts files
* Fully integrate Typescript into the build system
* ... (many other things I did no think of)

## More information

* <https://dev.to/kojikanao/introduce-typescript-to-react-js-project-with-ts-migrate-190h>
* <https://medium.com/airbnb-engineering/ts-migrate-a-tool-for-migrating-to-typescript-at-scale-cd23bfeb5cc>
* <https://github.com/airbnb/ts-migrate>


robertsLando:
Humm actually the tool hasn't do that much, if I open a class like mqtt client I can convert it to es6 class with a click in vscode and does a better work, also I would expect at least all require to be tranlsated to "import * from"

ahochsteger:
> Humm actually the tool hasn't do that much, if I open a class like mqtt client I can convert it to es6 class with a click in vscode and does a better work, also I would expect at least all require to be tranlsated to "import * from"

@robertsLando Yes, that's right.
I've done some research on how to automate the migration of large code bases and stumbled upon the following resources and tools:

* <https://medium.com/@cpojer/effective-javascript-codemods-5a6686bb46fb>
* <https://github.com/facebook/jscodeshift>
* <https://github.com/rajasegar/awesome-codemods>
* <https://github.com/cpojer/js-codemod/>
* <https://github.com/5to6/5to6-codemod>

Maybe a multi-pass approach would allow us to combine the best of available tools:

1. Extract JavaScript code from .vue files into separate .js files (just to benefit from the migration tools - may be moved back once they are migrated)
2. Convert all JavaScript files to modern ES6 modules using appropriate codemods
3. Migrate JavaScript to TypeScript using ts-migrate
4. Optional: Move TypeScript code back into .vue files

Here are some experiments that showcase how the JavaScript to ES6 migration could be done but I ran into problems that I was not able to resolve due to the lack of knowledge about how the files in 'lib/' are used:

```bash
sudo npm install -g jscodeshift
npm install 5to6-codemod
jscodeshift -t node_modules/5to6-codemod/transforms/cjs.js lib
```

app.ts
```diff
@@ -60,12 +69,12 @@ function startGateway () {
   gw = new Gateway(settings.gateway, zwave, mqtt)
 }
 
-app.startSocket = function (server) {
+app.startSocket = function (server: any) {
```
AlCalzone, line 72: Wow, that's bad. I think you'll get better results by converting the require statements to imports and let VSCode deduce the types.

ahochsteger: @AlCalzone You are right that this is bad if it would already be the finished migration ;-).

I guess this is on purpose for the chosen migration strategy "All-in migration" (see <https://medium.com/airbnb-engineering/ts-migrate-a-tool-for-migrating-to-typescript-at-scale-cd23bfeb5cc>) that states the following advantages:
* Consistency across a project
* Fixing just one type is much easier than fixing the whole file

This way you can migrate the whole code base to typescript without having to support a hybrid JavaScript/TypeScript build system and code base.

The next step would of course be introducing the types, but that is something that might take a bit longer ...

robertsLando:
@ahochsteger I would prefer a cleaner approach, actually as I said this module isn't doing that much IMO, I just see lot annotations and any types everywhere, I thing we could start with the backend migration, file by file, and then move also the  frontend by splitting vue and js content

ahochsteger:
> @ahochsteger I would prefer a cleaner approach, actually as I said this module isn't doing that much IMO, I just see lot annotations and any types everywhere, I think we could start with the backend migration, file by file, and then move also the frontend by splitting vue and js content.
> 
> There are vscode tools that do much better

Yes, the automated approach may be overkill for zwavejs2mqtt since there are not that many JavaScript files to migrate.

robertsLando:
@ahochsteger I would prefer a cleaner approach to this, something like: https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html

PR #25, opened 2020-11-24
feat: multicast/broadcast apis

### LINKED ISSUE #24: [feat] Multicast commands
Ref: https://github.com/zwave-js/zwavejs2mqtt/issues/22#issuecomment-733043629

robertsLando:
Fixes #24


README.md
```diff
@@ -728,6 +729,30 @@ I will set the Heating setpoint of the node with id `4` located in the `office`
 
 `zwave/office/nodeID_4/thermostat_setpoint/heating`
 
+### Multicast
+
+You can send Multicast requests to _all values with a specific suffix_ of a _group_ of nodes in the network.
```
chrisns, line 734: I'm not really clear on this, maybe provide a use case from the outset such as adjusting all your thermostats/lights?

be clearer that this is just a shorthand, it's not an _actual multicast_, behind the scenes zj2m is doing the heavy lifting and iterating

chrisns (changes requested):

README.md
AlCalzone: `zwave-js` does support multicast (if you use the same command and target the same endpoint of all nodes)

AlCalzone: I've raised https://github.com/zwave-js/node-zwave-js/issues/1150 to track that.

AlCalzone: For reference, what you can do already is this:
```ts
const cmd = new BinarySwitchCCSet(driver, {
	nodeId: [2,3,4,5],
	targetValue: true,
});
await driver.sendCommand(cmd);
```

robertsLando: I think we should wait for:  zwave-js/node-zwave-js#1150 Before going on with this?

robertsLando: > For reference, what you can do already is this:

@AlCalzone How could I get the correct command class constructor based on the command class id?

README.md
```diff
@@ -728,6 +729,30 @@ I will set the Heating setpoint of the node with id `4` located in the `office`
...
+
+It works like the set value API without the node name and location properties.
+If the API is correctly called the same payload of the request will be published
+to the topic without `/set` suffix.
+
+Example of multicast command (gateway configured as `named topics`):
+
+`zwave/_CLIENTS/ZWAVE_GATEWAY-test/multicast/thermostat_setpoint/heating/set`
+
+Payload: `{nodes: [5, 7, 9], value: 25.5}`
```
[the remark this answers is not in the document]
robertsLando: Need to check this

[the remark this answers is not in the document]
robertsLando: > But when i send: {"nodes":[2,3],"value": 70.0}(Also value 70 (without the .0) is giving the same error)

THis is json, the other one was js format, I'm converting the payload from string to json so that's why it wasn't working with the first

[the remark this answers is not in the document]
robertsLando: @ODINServ could you try with latest changes?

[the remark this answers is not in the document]
ODINServ: It works with:
```{"nodes":[2,3],"value":99}```

But not with:
```{nodes:[2,3],value: 99}```

With error:
```  
z2m:Mqtt Message received on zwavejs2mqtt/_CLIENTS/ZWAVE_GATEWAY-server/multicast/switch_multilevel/endpoint_1/targetValue/set +34s
z2m:Gateway No nodes found in multicast request to switch_multilevel/endpoint_1/targetValue +0ms```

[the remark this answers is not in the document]
robertsLando: @ODINServ like I said in my previous comment the not working format will never work, it's js non JSON :) now it picks the correct value at least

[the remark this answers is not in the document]
ODINServ: Ah check, should the docs than contain the quotes as well ?

I've just tried with all 12 nodes, it works nice for now, i guess it will be even better with the real multicast!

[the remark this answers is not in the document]
robertsLando: > i guess it will be even better with the real multicast!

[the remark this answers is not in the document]
robertsLando: > Ah check, should the docs than contain the quotes as well ?

You right about docs, fixing them now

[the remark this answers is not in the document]
ODINServ: I'll try a broadcast tomorrow, and see what happens.

Besides that i disabled all power reporting features of the nodes, i don't need that info.
I want to see if it makes sense if the network is less loaded that the unicast messages are send and received better.

robertsLando:
@ODINServ Could you give this a try?

robertsLando:
@ODINServ @AlCalzone I'm still not 100% sure about using this way as it just works for not secure devices. I think the old way was 'dirty' and slow but at least it works

AlCalzone:
> I'm still not 100% sure about using this way as it just works for not secure devices. I think the old way was 'dirty' and slow but at least it works

What about:
1. checking which devices are insecure and using multicast on them
2. doing one-by-one singlecast on the rest

You could later extend this when we have S2 support to create and use S2 Multicast groups.

README.md
AlCalzone: Sorry, I didn't see this...
https://github.com/zwave-js/node-zwave-js/blob/daf0feb9bfe14ae5b9a24c0d0d53206d5bdd41e9/packages/zwave-js/src/lib/commandclass/CommandClass.ts#L1057-L1059

But I'm not sure if that is exported through the main entry point.

lib/ZwaveClient.js
```diff
@@ -1889,6 +1889,43 @@ ZwaveClient.prototype.callApi = async function (apiName, ...args) {
...
+ * @param {number[]} nodes Array of nodes ids
+ * @param {Object} valueId Zwave valueId object
+ * @param {Integer|String} value The value to send
+ */
+ZwaveClient.prototype.writeMulticast = async function (nodes, valueId, value) {
+  if (this.driverReady) {
+    try {
+      const multicastGroup = this.driver.controller.getMulticastGroup(nodes)
+
+      await multicastGroup.setValue(valueId, value)
```
AlCalzone, line 1922: Like I wrote in my comment, I would check for `ZWaveErrorCodes.CC_NotSupported` and fall back to iterating singlecast requests.

AlCalzone (changes requested):

lib/ZwaveClient.js
robertsLando: @AlCalzone Could this cause multiple writes?

AlCalzone: No, the checks happen before anything is sent.

lib/Gateway.js
```diff
@@ -462,32 +463,59 @@ async function onApiRequest (topic, apiName, payload) {
...
-    payload = this.parsePayload(
-      payload,
-      this.topicValues[values[0]],
-      this.topicValues[values[0]].conf
-    )
-    for (let i = 0; i < values.length; i++) {
-      this.zwave.writeValue(this.topicValues[values[i]], payload)
-    }
+  if (payload.value === undefined) {
+    debug('No value found in broadcast request')
```
robertsLando, line 471:
```suggestion
    logger.error('No value found in broadcast request')
```

lib/Gateway.js
```diff
@@ -462,32 +463,59 @@ async function onApiRequest (topic, apiName, payload) {
...
+    return
   }
+
+  const error = utils.isValueId(valueId)
+
+  if (typeof error === 'string') {
+    debug('Invalid valueId: ' + error)
```
robertsLando, line 478:
```suggestion
    logger.error('Invalid valueId: ' + error)
```

lib/Gateway.js
```diff
@@ -462,32 +463,59 @@ async function onApiRequest (topic, apiName, payload) {
...
   }
 }
 
+async function onMulticastRequest (payload) {
+  const nodes = payload.nodes
+  const valueId = payload
+  const value = payload.value
+
+  if (!nodes || nodes.length === 0) {
+    debug('No nodes found in multicast request')
```
robertsLando, line 500:
```suggestion
    logger.error('No nodes found in multicast request')
```

lib/Gateway.js
```diff
@@ -462,32 +463,59 @@ async function onApiRequest (topic, apiName, payload) {
...
+    return
+  }
+
+  const error = utils.isValueId(valueId)
+
+  if (typeof error === 'string') {
+    debug('Invalid valueId: ' + error)
```
robertsLando, line 507:
```suggestion
    logger.error('Invalid valueId: ' + error)
```

lib/Gateway.js
```diff
@@ -462,32 +463,59 @@ async function onApiRequest (topic, apiName, payload) {
...
+    return
+  }
+
+  if (payload.value === undefined) {
+    debug('No value found in multicast request')
```
robertsLando, line 512:
```suggestion
    logger.error('No value found in multicast request')
```

lib/ZwaveClient.js
```diff
@@ -2207,6 +2207,43 @@ ZwaveClient.prototype.callApi = async function (apiName, ...args) {
...
+ * @param {Integer|String} value The value to send
+ */
+ZwaveClient.prototype.writeBroadcast = async function (valueId, value) {
+  if (this.driverReady) {
+    try {
+      const broadcastNode = this.driver.controller.getBroadcastNode()
+
+      await broadcastNode.setValue(valueId, value)
+    } catch (error) {
+      debug(`Error while sending broadcast ${value} to CC ${valueId.commandClass} ${valueId.property} ${valueId.propertyKey || ''}: ${error.message}`)
```
robertsLando, line 2223:
```suggestion
      logger.error(`Error while sending broadcast ${value} to CC ${valueId.commandClass} ${valueId.property} ${valueId.propertyKey || ''}: ${error.message}`)
```

lib/ZwaveClient.js
```diff
@@ -2207,6 +2207,43 @@ ZwaveClient.prototype.callApi = async function (apiName, ...args) {
...
+ * @param {Integer|String} value The value to send
+ */
+ZwaveClient.prototype.writeMulticast = async function (nodes, valueId, value) {
+  if (this.driverReady) {
+    try {
+      const multicastGroup = this.driver.controller.getMulticastGroup(nodes)
+
+      await multicastGroup.setValue(valueId, value)
+    } catch (error) {
+      debug(`Error while sending multicast ${value} to CC ${valueId.commandClass} ${valueId.property} ${valueId.propertyKey || ''}: ${error.message}`)
```
robertsLando, line 2242:
```suggestion
      logger.error(`Error while sending multicast ${value} to CC ${valueId.commandClass} ${valueId.property} ${valueId.propertyKey || ''}: ${error.message}`)
```

robertsLando:
@blhoward2 If you would like to give this a try. I merged it with actual master

blhoward2:
@robertsLando How do I test this? Send an mqtt set command with multiple nodes or something? I’m just not clear how it’s meant to be triggered.

blhoward2:
@robertsLando Can you give me some hint what I'm supposed to send and to where to test this?

blhoward2:
@robertsLando Ok, I figured this out. This does exactly what I'm looking for and I think it would be heavily used if people knew what it could do, but it is buggy. I published the following json to three nodes setting them to 0. Node 3 turned off physically, but remained on in HA. Node 5 dimmed down, but did not turn off completely, and also remained on in HA. (This is odd...nodes 3 and 5 are both GE dimmers, the same model, with the same ramp rate/timing. Publishing a payload to just node 5 with a targetValue of 0 does it make it go fully off.) 

They do go off in HA if I publish a targetValue of 0 directly to them so I'm not sure exactly why they don't go off with multicast. Is it because no response is received, so the currentValue is never updated? @AlCalzone Thoughts?

Topic: homeassistant/_CLIENTS/ZWAVE_GATEWAY-MQTT/multicast/set
```
{
  "commandClass": 38,
  "property": "targetValue",
  "value": 50,
  "nodes": [3, 5, 9]
}
```

[zwavejs_1.log](https://github.com/zwave-js/zwavejs2mqtt/files/5901032/zwavejs_1.log)

[zwavejs2mqtt.log.zip](https://github.com/zwave-js/zwavejs2mqtt/files/5901035/zwavejs2mqtt.log.zip)

robertsLando:
@AlCalzone

AlCalzone:
Multicast SETs don't poll (because that would require one GET for each node, defeating the purpose of a multicast). But we should assume that the command was accepted by all nodes if it succeeds like we already do for singlecasts.

ODINServ:
I was very busy lately and did not have any time to work on my home automation stuff.

So to summarize what i did:
* I re-included all nodes unsecured to reduce my strange network issues. (I dont have any now :))
* The last thing i did was running my setup "the npm way" just after the new broadcast API became available via MQTT.
* * Which i'm using now for 1.5 months without any issues!

From: @blhoward2 
> This does exactly what I'm looking for and I think it would be heavily used if people knew what it could do

Agreed maby this feature requests will solve this: https://github.com/zwave-js/zwavejs2mqtt/issues/70

I'd switched to this branch (without linking the node-zwave-js module) and tested it.

For me with my FGR223 modules this is working as expected with the same results and question that @blhoward2 has.
> Node 3 turned off physically, but remained on in HA.

From @AlCalzone 
> Multicast SETs don't poll (because that would require one GET for each node, defeating the purpose of a multicast). But we should assume that the command was accepted by all nodes if it succeeds like we already do for singlecasts.

So when using the (new) broadcast API, the values are updated when (in my case) the shutters reach there targetValue, not sure if this info helps.

lib/ZwaveClient.js
robertsLando: Done

robertsLando:
@blhoward2 Could you give this a try?

AlCalzone (dismissed):

blhoward2:
@robertsLando So I've tested this around a dozen times on a variety of devices and it seems to work pretty flawlessly. The weirdness I had before with one switch that didn't go off the full way has resolved itself. Some of my switches do in fact call back in their new state, so they update in HA. For others, they do not change as we're still waiting on @AlCalzone for https://github.com/zwave-js/node-zwave-js/issues/1550.

Topic: homeassistant/_CLIENTS/ZWAVE_GATEWAY-MQTT/multicast/set

{
  "commandClass": 32,
  "property": "targetValue",
  "value": 255,
  "nodes": [3, 5, 9, 13, 14, 19, 20, 21, 70, 112, 113, 114, 116, 117, 125]
}

robertsLando:
Thanks for your feedback @blhoward2, based on this seems that it should be good on my side, can I merge this?

docs/guide/mqtt.md
```diff
@@ -169,27 +171,64 @@ I will set the Heating setpoint of the node with id `4` located in the `office`
 
 ## Broadcast
 
-You can send broadcast values to _all values with a specific suffix_ in the network.
+You can send two kind of broadcast requests:
 
-Broadcast API is accessible from:
+1. Send it to values to _all values with a specific suffix_ in the network.
```
robertsLando, line 176:
```suggestion
1. Send it to _all values with a specific suffix_ in the network.
```

PR #28, opened 2020-11-24
Dockerfile.contrib

larstobi:
Dockerfile to build from local node-zwave-js git repo.


robertsLando:
@larstobi Is this actually working? 

Did you checked https://zwave-js.github.io/node-zwave-js/#/development/installing-from-github ?

larstobi:
It works on my machine, at least. I haven't tried on anybody elses. Yes, I used that guide from your link as a starting point.

docker/Dockerfile.contrib
```diff
@@ -0,0 +1,79 @@
...
+
+WORKDIR /root/node-zwave-js
+COPY node-zwave-js .
+
+RUN npm install
+RUN npx lerna init
+RUN npx lerna bootstrap
+RUN npx lerna run build
+RUN npx lerna exec -- npm link
+RUN npm prune --production
```
robertsLando, line 19: @AlCalzone I think this could be improved, also I think that we should add `npm install -g npm@7` before, right?

robertsLando:
@larstobi I think that you copuld merge step 1 and 2 into one

docker/README.contrib.md
```diff
@@ -0,0 +1,27 @@
...
+     "vue-prism-editor": "^1.2.2",
+     "vue-router": "^3.4.3",
+     "vuetify": "^2.3.10",
+-    "vuex": "^3.5.1",
+-    "zwave-js": "^5.5.0"
++    "vuex": "^3.5.1"
+   },
+   "devDependencies": {
+     "@babel/core": "^7.11.5",
+```
```
robertsLando, line 21: Why this ? Also I would suggest to better explain that you should firstly clone both projects in a folder:

```
mkdir -p testing && cd testing
git clone https://github.com/zwave-js/node-zwave-js
git clone https://github.com/zwave-js/zwavejs2mqtt
wget https://raw.githubusercontent.com/zwave-js/zwavejs2mqtt/master/docker/Dockerfile.contrib
docker build -t zwavejs2mqtt .
```

robertsLando:
Also I'm not sure that the links will work in that way

> First, npm link in a package folder will create a symlink in the global folder {prefix}/lib/node_modules/<package> that links to the package where the npm link command was executed. It will also link any bins in the package to {prefix}/bin/{name}. Note that npm link uses the global prefix (see npm prefix -g for its value). Next, in some other location, npm link package-name will create a symbolic link from globally-installed package-name to node_modules/ of the current folder.

So essentially the link will be lost when building the last container (as the sym link is added globally). I suggest do to everything in only one build stage

docker/Dockerfile.contrib
AlCalzone: Uh that seems to be based on the old description. The new one is a bit different:
https://zwave-js.github.io/node-zwave-js/#/development/installing-from-github

Since `npm workspaces`, it is no longer necessary to init or bootstrap lerna.

docker/README.contrib.md
larstobi: I removed the zwave-js dependency so that npm doesn't install it. Otherwise we would have to delete it before overwriting with the master branch version, right?

larstobi: Good idea about the prerequisites explanation!

larstobi:
> @larstobi I think that you copuld merge step 1 and 2 into one

Not sure what you mean by step 1 and 2. Do you mean:
`RUN npm install && npx lerna init`

docker/Dockerfile.contrib
larstobi: I will try to remove the lerna steps.

robertsLando:
I have corrected my comment.

I mean you should not build them in separete build steps but keep only one, firstly build zwavejs then z2m and after building it just run the link command and you don't need to override package.json, the link will override the dep automatically in node modules. 

Also having a one unique step will fix the link problem I mentioned before :)

docker/README.contrib.md
AlCalzone: You could also try changing the `zwave-js` dependency to `"file:/path/to/zwave-js/package"` if you want to avoid the intial installation of zwave-js from npm.
The `zwave-js` workspace gets linked in itself by npm, so that **might** work.

chrisns:
I'm not sure i really follow for the need for this, but can we add to the ci build, even if we don't push somewhere so that we know when we break it?

docker/README.contrib.md
robertsLando: > You could also try changing the zwave-js dependency to "file:/path/to/zwave-js/package" if you want to avoid the intial installation of zwave-js from npm.

Didn't know that was actually possible, is it possible to run somthing like `npm install file:/path/to/zwave-js/package` ?

robertsLando:
> I'm not sure i really follow for the need for this, but can we add to the ci build, even if we don't push somewhere so that we know when we break it?

@chrisns I don't get it, in order this to be broken it means z2m or zwave-js masters are broken someway. I see this a developer tool for the ones that are looking for testing latest builds of both projects using master or specific branches

chrisns:
> > I'm not sure i really follow for the need for this, but can we add to the ci build, even if we don't push somewhere so that we know when we break it?
> 
> @chrisns I don't get it, in order this to be broken it means z2m or zwave-js masters are broken someway. I see this as a developer tool for the ones that are looking for testing latest builds of both projects using master or specific branches (with docker, this is lot easier using nodejs but I see there are lot users that still prefer docker)

agree its a dev tool, I don't see the real usecase of why you'd do it /this/ way. if it does need to exist I feel like it could easily fall out of sync however, so would recommend exercising the use case in a CI job which would double up as documentation on how to use it 😄

robertsLando:
>  I don't see the real usecase of why you'd do it /this/ way

AlCalzone:
> IMO docker isn't really usefull at all now that there isn't OZW lib to compile

Well you do have a confined environment where the libraries are used with a fixed Node.js version without influences from the outside world. I can see the benefit in that.

robertsLando:
> Well you do have a confined environment where the libraries are used with a fixed Node.js version without influences from the outside world. I can see the benefit in that.

Yes that actually is good for every project that you run with docker, it's docker main feature. I mean that if you can use nodejs in your machine it would be much easier and faster, expecially for testing. But as said I understand that someone could still want docker so I'm ok to also provide it as it is already managed :)

chrisns:
if it were me, i'd just `docker run -v ${PWD}:/app zwave-js/zwavejs2mqtt bash` and then run everything in the container starting with `npm i`.
It's a different use case since you'll want dev dependencies etc in there so will be different. but do agree with having a representative version in the same arch, i guess you can probably port forward through to attach to a debugger etc

robertsLando:
>i guess you can probably port forward through to attach to a debugger etc

Yep you can do remote debugging

chrisns:
but your editor+git client is still (presumably) gonna be in the host machine, so wouldn't you want to volume mount the workspace?

robertsLando:
> but your editor+git client is still (presumably) gonna be in the host machine, so wouldn't you want to volume mount the workspace?

:man_shrugging:

chrisns:
> 🤷‍♂️

@larstobi can you explain a bit more about how you're using this and why it's better/different/preferred than mounting in the workspace with `docker run -v` ?
Thanks

larstobi:
Building a self-contained containers instead of running just a node container with mounted source code is a good way of getting immutability for the service. It's the way I prefer to run my production containers. It's also distributable, I can store it in a docker registry, and easily run the exact same container on a different server.

Also, it's a really easy way for new users to get going. Just point them to the container on Docker Hub, and run it. No need to check out the source repo and build it.

There are many benefits to it as I see it, but those mentioned are some. :-)

larstobi:
I've made some changes to the Dockerfile based on input from you. It builds successfully on my machine. I updated it to use the newer npm version along with the new build instructions for zwave-js. I wasn't easily able to get it to work with Alpine Linux, and it doesn't have the more recent node and npm.

I'm getting some new runtime errors from missing dependencies after the updates, though. For instance `triple-beam` and `reflect-metadata`.  Will try to work it out tomorrow.

AlCalzone:
> I'm getting some new runtime errors from missing dependencies after the updates, though. For instance `triple-beam` and `reflect-metadata`. Will try to work it out tomorrow.

Weird... `reflect-metadata` is a dependency of `zwave-js`, `triple-beam` should come with `winston`.

docker/Dockerfile.contrib
```diff
@@ -0,0 +1,31 @@
+FROM node AS build
+
+COPY node-zwave-js /root/node-zwave-js
+COPY zwavejs2mqtt /root/zwavejs2mqtt
+
+WORKDIR /root/node-zwave-js
+RUN npm install
```
robertsLando, line 7:
```suggestion
RUN npm install -g npm@7
RUN npm install
```

docker/Dockerfile.contrib
```diff
@@ -0,0 +1,31 @@
...
+RUN npm run build
+
+WORKDIR /root/zwavejs2mqtt
+RUN npm config set unsafe-perm true
+RUN npm install
+RUN npm run build
+# RUN npm prune --production
+RUN cp -rL /root/node-zwave-js/packages/zwave-js node_modules/
```
robertsLando, line 15:
```suggestion
RUN npm install ../node-zwave-js
```

This will create a symlink to the repository

docker/Dockerfile.contrib
```diff
@@ -0,0 +1,31 @@
...
+    package.sh \
+    src \
+    static \
+    stylesheets
+
+FROM node
+LABEL maintainer="robertsLando"
+# Copy files from build stage
+COPY --from=build /root/zwavejs2mqtt /usr/src/app
+WORKDIR /usr/src/app
```
robertsLando, line 29:
```suggestion
```

Remove this lines, you are creating a new step here

docker/README.contrib.md
robertsLando: This can be deleted with my latest reviews

robertsLando (changes requested):

docker/Dockerfile.contrib
AlCalzone: Wouldn't it make more sense to `npm install ../node-zwave-js/packages/zwave-js`? Not sure though, this whole `npm workspace` link magic is also new to me :D

robertsLando: Shouldn't npm 7 do the job?

robertsLando: Otherwise, try both solutions @larstobi

chrisns:
> Building a self-contained container instead of running just a node container with mounted source code is a good way of getting immutability for the service. It's the way I prefer to run my production containers. It's also distributable, I can store it in a docker registry, and easily run the exact same container on a different server.
> 
> Also, it's a really easy way for new users to get going. Just point them to the container on Docker Hub, and run it. No need to check out the source repo and build it.
> 
> There are many benefits to it as I see it, but those mentioned are some. :-)

I'm with you a million percent on the "why containerize" I just don't really follow to what distinguishes this from the existing, and why we need both https://github.com/zwave-js/zwavejs2mqtt/blob/master/docker/Dockerfile or can we remove one?

robertsLando:
@chrisns The main difference is that this actually allows to use a custom version of zwavejs. We could optionally add an arg on the other one to support this?

chrisns:
Yeah. That's kinda what I was leaning towards, or a super slim one that inherits from the base. 
But given the pace of change of the zwavejs I'm not sure if there's even a strong case for that.
So it's not for sharing and only as a dev tool, again fine but I still don't follow the real 'why'. It feels like an anti pattern or I've just misunderstood the original intention

docker/Dockerfile.contrib
larstobi: This step is to get the size down, and reduce the number of layers. I prefer to keep it. I would prefer to get alpine working, for size, but it wasn't working with the latest code.

larstobi:
> It feels like an anti pattern or I've just misunderstood the original intention

I'm unable to see how this is an anti-pattern. Could you explain?

docker/Dockerfile.contrib
larstobi: I'm pretty newb with the npm tool, but so far, to me it is behaving weirdly. Especially with the dependency vendoring and the symlinking. 

And I'm getting a lot of "npm ERR! EISDIR: illegal operation on a directory, copyfile" messages in a lot of the npm commands.

```
Step 12/15 : RUN npm install ../node-zwave-js/packages/zwave-js
 ---> Running in a68b191e94a7
npm ERR! code EISDIR
npm ERR! syscall copyfile
npm ERR! path /home/node/zwavejs2mqtt/node_modules/agent-base
npm ERR! dest /home/node/zwavejs2mqtt/node_modules/.agent-base-lA1oG2lU
npm ERR! errno -21
npm ERR! EISDIR: illegal operation on a directory, copyfile '/home/node/zwavejs2mqtt/node_modules/agent-base' -> '/home/node/zwavejs2mqtt/node_modules/.agent-base-lA1oG2lU'

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/node/.npm/_logs/2020-11-26T15_46_33_668Z-debug.log
The command '/bin/sh -c npm install ../node-zwave-js/packages/zwave-js' returned a non-zero code: 235
```

I am spending a lot of time on strange npm errors, but hopefully I'm learning lots of new stuff. :-)

docker/Dockerfile.contrib
larstobi:
```
Step 11/18 : RUN npm install -g npm@7
 ---> Running in 7616e7296876
npm ERR! code EISDIR
npm ERR! syscall copyfile
npm ERR! path /usr/local/lib/node_modules/npm
npm ERR! dest /usr/local/lib/node_modules/.npm-i9nnxROI
npm ERR! errno -21
npm ERR! EISDIR: illegal operation on a directory, copyfile '/usr/local/lib/node_modules/npm' -> '/usr/local/lib/node_modules/.npm-i9nnxROI'

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2020-11-26T15_51_43_673Z-debug.log
The command '/bin/sh -c npm install -g npm@7' returned a non-zero code: 235
```

robertsLando: Hummm is there a node image with npm 7 already installed?

chrisns:
> > It feels like an anti pattern or I've just misunderstood the original intention
> 
> I'm unable to see how this is an anti-pattern. Could you explain?

heya
the reasons you mentioned in here https://github.com/zwave-js/zwavejs2mqtt/pull/28#issuecomment-733970144 don't seem to stand up because this surely(?) isn't for running something that is for production or sharing, by definition its a dev enabler that there doesn't appear to be an appetite to maintain by even adding a CI job/step to test or push the image anywhere.

I'm probably totally missing something but I don't really grasp who this is for and why it is needed.
Sorry I don't mean to come across like a ... and it's more likely me than you thats wrong, I just don't get it - I suspect others will be the same. So would be great if you could you could really spell it out :)

I'm on the z2m slack if it's easier to chat about it there rather than bat github comments back and forth?

docker/Dockerfile.contrib
larstobi: So, I figured this one is because zwavejs2mqtt already has agent-base as a dependency, so trying to install it again fails. Of course.

robertsLando:
FYI I'm ok for this to land, personally I would never use such thing but if could help some users with testing I'm totally :+1:

docker/Dockerfile.contrib
robertsLando: Ok but if you use this you also have to copy the zwave-js folder

docker/Dockerfile.contrib
robertsLando: So it's fixed?

larstobi: No, I don't know how to fix it. Will keep working on it, though.

larstobi: Without the dependencies from zwave-js it won't run.

robertsLando: Did you try with `npm install ../node-zwave-js/packages/zwave-js`?

larstobi:
This last version builds and runs on my machine.

robertsLando:
@larstobi Could you delete that Readme and add a section to the already existing readme in `docker` folder with the steps needed to build?  Like: 

```
mkdir -p testing && cd testing
git clone https://github.com/zwave-js/node-zwave-js
git clone https://github.com/zwave-js/zwavejs2mqtt

## Checkout repos to any branch/commit you need to test

cp zwavejs2mqtt/docker/Dockerfile.contrib ./Dockerfile
docker build -t zwavejs2mqtt .
```

robertsLando (approved):
LGTM

robertsLando:
@larstobi Sorry I just saw that lint fails, could you run `npm run lint-fix` ?

chrisns (approved):
would be great to build with actions and maybe push somewhere

docker/Dockerfile.contrib
```diff
@@ -0,0 +1,25 @@
...
+RUN mkdir -p /usr/local/lib/node_modules && chown node /usr/local/lib/node_modules
+USER node
+
+COPY --chown=node node-zwave-js /home/node/node-zwave-js
+COPY --chown=node zwavejs2mqtt /home/node/zwavejs2mqtt
+
+WORKDIR /home/node/node-zwave-js
+RUN npm install
+RUN npm run build
+RUN npm link
```
robertsLando, line 12:
```suggestion
RUN npx lerna exec -- npm link
```

Afeter speaking with @AlCalzone seems this is better as will link all workspaces of zwavejs

robertsLando (changes requested):

docker/README.md
```diff
@@ -209,3 +209,24 @@ docker run --rm -p 8091:8091 --device=/dev/ttyACM0 -it --mount source=zwavejs2mq
 ```bash
 docker run --rm -p 8091:8091 --device=/dev/ttyACM0 -it --mount source=zwavejs2mqtt,target=/dist/pkg zwavejs/zwavejs2mqtt_build sh
 ```
+
+# Building a container using Dockerfile.contrib
```
robertsLando, line 213:
```suggestion
## Building a container using Dockerfile.contrib
```

This is why ci is failing

robertsLando (changes requested):

larstobi:
> would be great to build with actions and maybe push somewhere

There is a Docker repo here https://hub.docker.com/r/zwavejs/zwavejs2mqtt
This can be used to set up Docker Hub Automated Builds with some modifications.

robertsLando:
> would be great to build with actions and maybe push somewhere

chrisns:
> > would be great to build with actions and maybe push somewhere
> 
> @chrisns Where should be pushed? Why? This in inteded to be used just locally

earlier comments suggested it was useful for sharing, eitherway, it needs to be in CI if we're going to be able to maintain it. even if its just a amd64 build that doesn't get pushed

docker/Dockerfile.contrib
robertsLando: @larstobi

robertsLando: @larstobi Is this working?

docker/Dockerfile.contrib
```diff
@@ -1,25 +1,33 @@
...
 
 WORKDIR /home/node/node-zwave-js
-RUN npm install
-RUN npm run build
-RUN npm link
+RUN rm -f package-lock.json
+RUN yarn install
+RUN yarn run build
+RUN yarn install --production --frozen-lockfile
+RUN npx lerna exec -- yarn link
```
AlCalzone, line 15: Interesting combination of commands ;)
grafik

robertsLando: magic

robertsLando:
@larstobi Are you done here?

robertsLando (approved):

PR #33, opened 2020-11-27
chore(ci): skip ci and docker build with [skip ci]

robertsLando:


chrisns (changes requested):
what's the use case?
given we lint the markdown, json, etc theres not much left you wouldn't want to test before merging right?

robertsLando:
> given we lint the markdown, json, etc theres not much left you wouldn't want to test before merging right?

It's expecially for that

PR #35, opened 2020-11-27
fix: meterType devices missing ccSpecific values

billiaz:
When a Command Class is type Meter. In some cases is not a sensor. This metric is a Delta in time, or reset parameter for the Sensor.
There Parameters miss the ccSpecific values, which sets the type of sensor. This change fixes this error

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -1321,7 +1321,11 @@ Gateway.prototype.discoverValue = function (node, vId) {
         } else if (cmdClass === CommandClasses.Meter) {
           // https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/MeterCC.ts
           // https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/meters.json
-          sensor = Constants.meterType(valueId.ccSpecific.meterType)
+          if (valueId.ccSpecific) {
+            sensor = Constants.meterType(valueId.ccSpecific.meterType)
+          } else {
+            return
+          }
         } else if (cmdClass === CommandClasses['Pulse Meter']) {
           sensor = {
             sensor: 'pulse',
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -1317,12 +1317,18 @@ Gateway.prototype.discoverValue = function (node, vId) {
         // https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/MultilevelSensorCC.ts
         if (cmdClass === CommandClasses['Multilevel Sensor']) {
           // https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/sensorTypes.json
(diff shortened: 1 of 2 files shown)


lib/Gateway.js
```diff
@@ -1321,7 +1321,11 @@ Gateway.prototype.discoverValue = function (node, vId) {
         } else if (cmdClass === CommandClasses.Meter) {
           // https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/MeterCC.ts
           // https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/meters.json
-          sensor = Constants.meterType(valueId.ccSpecific.meterType)
+          if (valueId.ccSpecific) {
+            sensor = Constants.meterType(valueId.ccSpecific.meterType)
+          } else {
```
robertsLando, line 1326: Could you add a comment here to explain this? Good catch 👍

robertsLando (changes requested):

lib/Gateway.js
billiaz: Added some comments, can you share your thoughts on this?

robertsLando: I think it's ok, I would make the same check for multilevel sensor

billiaz: this specific issue is on watt and power metering devices. I expect this would not happen on multilevel. But adding it just in case

billiaz:
I am trying to fix hass discovery for multilevel sensors of temperature
The sensor has unit of measurement and this is a problem. I do not know how to do this in the node part. I spent a bit of time but...could not find it

the hass payload is: 
```
{
  "value_template": "{{ value_json.value }}",
  "device_class": "temperature",
  "state_topic": "zwave2mqtt/Garage/sensorBackSide/49/0/Air_temperature",
  "json_attributes_topic": "zwave2mqtt/Garage/sensorBackSide/49/0/Air_temperature",
  "device": {
    "identifiers": [
      "zwavejs2mqtt_0xcddb486e_node35"
    ],
    "manufacturer": "Shenzhen Neo Electronics Co., Ltd.",
    "model": "PIR Motion Sensor with Temperature Sensor (PD03Z)",
    "name": "Garage-sensorBackSide",
    "sw_version": "3.80"
  },
  "name": "Garage-sensorBackSide_temperature_air",
  "unique_id": "zwavejs2mqtt_0xcddb486e_35-49-0-Air temperature"
}
```

The payload is:
```
{
  "id": "35-49-0-Air temperature",
  "nodeId": 35,
  "commandClass": 49,
  "commandClassName": "Multilevel Sensor",
  "endpoint": 0,
  "property": "Air temperature",
  "propertyName": "Air temperature",
  "type": "number",
  "readable": true,
  "writeable": false,
  "label": "Air temperature",
  "ccSpecific": {
    "sensorType": 1,
    "scale": 0
  },
  "genre": "user",
  "unit": "°C",
  "list": false,
  "value": 0,
  "lastUpdate": 1606596275869
}
```

robertsLando (approved):
LGTM

PR #36, opened 2020-11-28
fix: fix multiple parameters for hass Discovery

billiaz:
This pull request applies two changes:
- **Binary Sensor discovery** was not working, fixing it by changing the SensorTypeName source
- Add ability to use **PropertyName** in case **Label is not sent** by driver
- Adds **unit_of_measurement** on **Battery level** devices by default.
- Replace **valueId.units** with  **valueId.unit**
- add README.md on prettierignore list. As some linting standards cannot be followed.

** Binary Sensor discovery**
discovery was not working, as the **sensorTypeName** was a number and not the name of sensor
changed the value used for this from Translated to just valueId.property

**Missing Label**
There are cases where the Label is missing from a Property. We can use the PropertyName in that cases.

I have create a small change in code to make this happen. Because my code was based on hassDiscovery branch, I am also making a pull request with that branch as target.

I have tested a container and I do see only Undefined labels changing to Property Name.
Any property having labels is not impacted.

**Batery Level**
Battery level unit was not discovered. Added a percentage under the payload of this parameter

**valueId.unit**
valueId.units is non existing, fixing the typo with valueId.unit


lib/Gateway.js
```diff
@@ -9,7 +9,7 @@ const path = require('path')
 const reqlib = require('app-root-path').require
 const utils = reqlib('/lib/utils.js')
 const EventEmitter = require('events')
-const { AlarmSensorType, BinarySensorType } = require('zwave-js')
+const AlarmSensorType = require('zwave-js')
```
AlCalzone, line 12: That (default) import is not correct.
```suggestion
const { AlarmSensorType } = require('zwave-js')
```

lib/Gateway.js
```diff
@@ -1265,17 +1265,21 @@ Gateway.prototype.discoverValue = function (node, vId) {
       case CommandClasses['Binary Sensor']:
         // https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/BinarySensorCC.ts#L41
 
-        let sensorTypeName = BinarySensorType[valueId.property]
+        // change the sensorTypeName to use directly valueId.property, as the old way was returning a number
+        // add a comment which shows the old way of achieving this value. This change fixes the Binary Sensor
+        // discovery
+        let sensorTypeName = valueId.property
```
AlCalzone, line 1271: I would try to use the `ccSpecific.sensorType` metadata property (numeric enum value) if it exists and only fall back to the `property`.

lib/Gateway.js
```diff
@@ -1361,9 +1366,13 @@ Gateway.prototype.discoverValue = function (node, vId) {
 
         Object.assign(cfg.discovery_payload, sensor.props || {})
 
-        if (valueId.units) {
+        // Discovered valueId.unit also exists. therefore added both unit and units to be
+        // checked, but only one used
+        if (valueId.unit) {
```
AlCalzone, line 1371: `unit` exists on the metadata, not the valueId itself. Or am I missing some zjs2mqtt magic here?

lib/Gateway.js
billiaz: My understanding this comes after metadata. and is parsed as valueId.

billiaz: My understanding is, valueId is internal of zwavejs2mqtt. and is already translated during the device discovery. this is homeassistant sensor discovery. afterwards

lib/Gateway.js
billiaz: My understanding this comes after metadata. and is parsed as valueId.

lib/Gateway.js
AlCalzone: @robertsLando ?

AlCalzone:
Ok I guess @robertsLando has to chime in here. I don't know the specifics how things are handled on this side.

lib/Gateway.js
robertsLando: @billiaz What does this fixes? I mean you say the old way was returning a number, sholdn't it return a string based on https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/BinarySensorCC.ts#L41 enum? At least that is what I want

lib/Gateway.js
robertsLando: > Or am I missing some zjs2mqtt magic here

I should create some interfaces here to make things more clear but yes the magic happens here: https://github.com/zwave-js/zwavejs2mqtt/blob/1ccab668170275f1a0195a622fe20d921bbd8de2/lib/ZwaveClient.js#L637

robertsLando: What I don't understand here is why you firstly check for `valueId.unit` and then `valueId.units`? [Here](https://github.com/zwave-js/zwavejs2mqtt/blob/1ccab668170275f1a0195a622fe20d921bbd8de2/lib/ZwaveClient.js#L674) I only set `unit`

lib/Gateway.js
```diff
@@ -1361,9 +1366,13 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
           // https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/scales.json
-          cfg.discovery_payload.unit_of_measurement = valueId.units
+          cfg.discovery_payload.unit_of_measurement = valueId.unit
+        } else if (valueId.units) {
+          cfg.discovery_payload.unit_of_measurement = valueId.unit
```
robertsLando, line 1375: This has no sense IMO

lib/ZwaveClient.js
```diff
@@ -650,7 +650,7 @@ function updateValueMetadata (zwaveNode, zwaveValue, zwaveValueMeta) {
     readable: zwaveValueMeta.readable,
     writeable: zwaveValueMeta.writeable,
     description: zwaveValueMeta.description,
-    label: zwaveValueMeta.label,
+    label: zwaveValueMeta.label || zwaveValue.propertyName, // when label is missing, re use propertyName. Usefull for webinterface
```
robertsLando, line 653: @AlCalzone could the label be undefined or is this caused by an error on zwavejs? If it's the second case I would remove this so we can detect undefined labels and fix them on zwavejs

robertsLando (changes requested):

lib/ZwaveClient.js
AlCalzone: The Label can be undefined if we don't set it in zwave-js. Your code should be prepared for that possibility.
But as I wrote in slack, we should find those locations and eliminate them.

lib/Gateway.js
aretakisv: I didn't know if we have units and unit so I started with covering both cases! i can remove it (as I thought in the beginning!

lib/Gateway.js
billiaz: @robertsLando  I have a Motion sensor. this sensor is 0x0c (12 dec) and name is Motion
I set a debug output and tried to find out what the name of sensor was.
This snippet:
```
let sensorTypeName = BinarySensorType[valueId.property]
```
Returns '0x0c' (12)

when printing on my debug output the valueid.property, I do find the string (motion, contact etc)
basically translating the above is: 
```
BinarySensorType['Motion']
```
I think using directly the string is good. 
If we want to cover both cases where we receive a number or string on valueId.property , I could create an isNaN check.
But My understanding reading code, valueId.Property is already parsed and made string

lib/Gateway.js
billiaz: I will fix this to just be unit! wanted to cover both cases, as it was not 100% clear for me if one case exists

lib/Gateway.js
AlCalzone: `property` contains the **name** of the enum member. `ccSpecific.sensorType` is numeric

robertsLando: Ok so this is correct :+1:

lib/Gateway.js
robertsLando: I never set `valueId.units` anywhere so that is useless :)

lib/Gateway.js
robertsLando: Remove it yes :)

lib/ZwaveClient.js
robertsLando: Ok so in poor words should I keep this change or remove this so we can detect undefined labels?

lib/Gateway.js
```diff
@@ -1366,14 +1366,8 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
-        // Discovered valueId.unit also exists. therefore added both unit and units to be
-        // checked, but only one used
-        if (valueId.unit) {
-          // https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/scales.json
-          cfg.discovery_payload.unit_of_measurement = valueId.unit
-        } else if (valueId.units) {
-          cfg.discovery_payload.unit_of_measurement = valueId.unit
-        }
+        // https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/scales.json
+        cfg.discovery_payload.unit_of_measurement = valueId.unit
```
robertsLando, line 1370: @billiaz I would keep the if here, or this could be undefined and hass will throw an error

robertsLando (approved):

lib/ZwaveClient.js
billiaz: we could have a suffix/prefix Property and help us still know what is, but not let it undifined
something like
```
label: zwaveValueMeta.label || ("%s - property", zwaveValue.propertyName),
```

robertsLando: This could be another valid solution

robertsLando:
@billiaz There are some lint issues

billiaz:
my mac has some issues with lint-fix/linting

this issue can probably be solved 
```
        if (valueId.property) {
          let sensorTypeName = valueId.property
        }
        // unsued discovery line. for historical reasons
        // let sensorTypeId = BinarySensorType[valueId.property]
        if (sensorTypeName) {
          sensorTypeName = this.mqtt.cleanName(
            sensorTypeName.toLocaleLowerCase()
          )
        }
        // TODO: Implement all BinarySensorTypes
        cfg = hassCfg['binary_sensor_' + sensorTypeName]
```
by making it:
```
        if (valueId.property) {
          let sensorTypeName = this.mqtt.cleanName(
            valueId.property.toLocaleLowerCase()
          )
          // TODO: Implement all BinarySensorTypes
          cfg = hassCfg['binary_sensor_' + sensorTypeName]
        }
```

lib/Gateway.js
```diff
@@ -1265,17 +1265,23 @@ Gateway.prototype.discoverValue = function (node, vId) {
       case CommandClasses['Binary Sensor']:
         // https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/BinarySensorCC.ts#L41
 
-        let sensorTypeName = BinarySensorType[valueId.property]
+        // change the sensorTypeName to use directly valueId.property, as the old way was returning a number
+        // add a comment which shows the old way of achieving this value. This change fixes the Binary Sensor
+        // discovery
+        if (valueId.property) {
```
robertsLando, line 1271: this if is useless, remove it as `property` is always defined

robertsLando (changes requested):

robertsLando (changes requested):
@billiaz Seems thre are still some lint issues?

billiaz:
> @billiaz Seems thre are still some lint issues?
@robertsLando 
this is an issue also seen in hasDiscovery. Prettierlint fails with readme. as we need to have some double quotes there.
I suggest to add a `.prettierignore` for `README.md` (I am adding it to test if this is the same failure here too! but this is what investigation showed on my laptop)

lib/ZwaveClient.js
billiaz: i think more beautifull would be:
```property: %s```

File: lib/ZwaveClient.js [hunk 8]
```diff
@@ -650,7 +650,7 @@ function updateValueMetadata (zwaveNode, zwaveValue, zwaveValueMeta) {
     readable: zwaveValueMeta.readable,
     writeable: zwaveValueMeta.writeable,
     description: zwaveValueMeta.description,
-    label: zwaveValueMeta.label,
+    label: zwaveValueMeta.label || ('%s (property)', zwaveValue.propertyName), // when label is missing, re use propertyName. Usefull for webinterface
```
robertsLando, line 653:
```suggestion
    label: zwaveValueMeta.label || (zwaveValue.propertyName +  ' (property)'), // when label is missing, re use propertyName. Usefull for webinterface
```

robertsLando (changes requested):

PR #39, opened 2020-11-30
fix topics for 2gig thermostats

chilicheech:
I suspect this may be necessary for other thermostats as well, but i have no way to test the other thermostats


robertsLando (approved):
LGTM. Thanks :)

PR #46, opened 2020-12-02
Dockerfile with optional remote branch support from @roflmao

larstobi:
Based on the idea and work by @roflmao by #41.

This is an alternative solution that gives us both variants in a single Dockerfile.

Also added short sha to package versions.

PR #49, opened 2020-12-04
set yarn install timeout to a large number to prevent timeout issues

chilicheech:


billiaz:
I believe this timeout is too high. This can cause unexpected hours of waiting without result.
I seriously suggest to tripple default timeout but no more  before higher value.

roflmao:
I propose using a build-arg in dockerfile to mitigate this problem. 
Setting --network-timeout to yarn default of 30000 (ms) with an overridable docker --build-arg
#46 is already ongoing so lets reach out to @larstobi.

Ref: https://github.com/yarnpkg/yarn/search?utf8=%E2%9C%93&q=NETWORK_TIMEOUT&type=

larstobi:
Thanks for proposing this change, @chilicheech! I incorporated your change as an optional timeout in #46. Does it look okay to you?

PR #52, opened 2020-12-04
[WIP] fix: discovery missing targetValue when runs first time

billiaz:
During initial discovery, while the setTopic value is undefined there is
an issue which arises with an nill value for command_topic on hass
discovery. This issue comes due to the "split" node compared to previous
driver, which combined both.
By checkinf if both values setTopic and getTopic are set, we make sure
to discover a value once is really complete.


billiaz:
Looking further into the problem. Using vsCode and remote attaching.
I add a break point at lib/Gateway.js #1309
This is the first line in discovery function.

While looping through values of node, what I find is 37-0-currentValue and 37-0-targetValue are undefined:
when first time 37-0-currentValue is found, node.values[vId] line 1286 is:
```
37-0-currentValue:{id: '29-37-0-currentValue', nodeId: 29, commandClass: 37, commandClassName: 'Binary Switch', endpoint: 0, …}
49-0-Power:{id: '29-49-0-Power', nodeId: 29, commandClass: 49, commandClassName: 'Multilevel Sensor', endpoint: 0, …}
__proto__:Object
```
this means 37-0-targetValue is not yet on the list of node.values.vId and therefore this is not an existing value. BUT due to the template... later on we register it as already discovered.
Best is to perform discovery once we have both currentValue and targetValue defined.

next time breakpoint is reached
```
37-0-currentValue:{id: '29-37-0-currentValue', nodeId: 29, commandClass: 37, commandClassName: 'Binary Switch', endpoint: 0, …}
37-0-targetValue:{id: '29-37-0-targetValue', nodeId: 29, commandClass: 37, commandClassName: 'Binary Switch', endpoint: 0, …}
49-0-Power:{id: '29-49-0-Power', nodeId: 29, commandClass: 49, commandClassName: 'Multilevel Sensor', endpoint: 0, …}
__proto__:Object
```
now discovery can be successful as we have get and set topics in the list.
Starting traceback on where these vIds are coming from.

this discovery is called by function onValueChanged.

checking the line
```
function onValueChanged (valueId, node, changed) {
```
I can still see the node contains only 3 values.
Continuing the trace back of this:
```function addValue (zwaveNode, zwaveValue) {``` calls the onValueChanged.. and still here `node` is defined with the same 3 vIds

checking the function onNodeReady
I discover that we loop over the list of values using the addValue function.
Because addValue uses the valueChanged logic. what happens is, valueChanged  executes discovery on specific value, while it has no full list of values. and causes this issue.

A small change in the logic could remove the push I made.. I am thinking how to achieve this.

I wish I didn't confuse everybody.

File: lib/ZwaveClient.js [hunk 1]
```diff
@@ -702,7 +703,23 @@ function updateValueMetadata (zwaveNode, zwaveValue, zwaveValueMeta) {
...
+  } else {
+    const zwaveValueMeta = zwaveNode.getValueMetadata(zwaveValue)
+
+    const valueId = initValue.call(this, zwaveNode, zwaveValue, zwaveValueMeta)
+
+    debug(`Node ${zwaveNode.id}: hass Discovery for ${valueId.id}`)
+
+    this.emit('readyForDiscovery', valueId, node)
+  }
+}
```
robertsLando, line 721: @billiaz Actually this is almost the same as `addValue` function. I suggest to add a bool flag instead to `addValue` and also to the event triggered

robertsLando (changes requested):
Does this actually fixes your problem?

robertsLando:
@AlCalzone I dunno if I'm missing something there or there is a bug somewhere on zwave-js. Actually I discover values when node is ready but seems that at that point not all values are really availables. This is another case like the other posted in zwavejs issue

robertsLando:
Steps: 

1. Bind `node added` event: https://github.com/zwave-js/zwavejs2mqtt/blob/master/lib/ZwaveClient.js#L103
2. When node added is triggered [this](https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L197) function is called, then I [bind events](https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L494) to node and init the node [here](https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L530)
3. Now if the node values are not ion cache I should start getting `value added` events but I don't do anything until the node is ready: https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L384
4. When the node is finally ready I then get all values by using `zwaveNode.getDefinedValueIDs` and I add one value by one.

Seems that on point 4 not all values are there

lib/ZwaveClient.js
aretakisv: I am finishing some changes! and I will post a new function, smaller, simpler

aretakisv:
> Steps:
> 
> 1. Bind `node added` event: https://github.com/zwave-js/zwavejs2mqtt/blob/master/lib/ZwaveClient.js#L103
> 2. When `node added` is triggered [this](https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L197) function is called, then I [bind events](https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L494) to node and init the node [here](https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L530)
> 3. Now if the node values are not ion cache I should start getting `value added` events but I don't do anything until the node is ready: https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L384
> 4. When the node is finally ready I then get all values by using `zwaveNode.getDefinedValueIDs` and I add one value by one: https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L275
> 
> Seems that on point 4 not all values are there

The values list used during discovery. Is not coming from `zwaveNode` but from the `node` variable which is updated by addValue function. And causes this.
The hassDiscovery is using the final state of the `node` variable for that node to discover values! and iterating it after finishing the `node` variable construction is the best way, as it is now

robertsLando:
> The values list used during discovery. Is not coming from zwaveNode but from the node variable which is updated by addValue function. And causes this.

The discovery should happen only when node is `ready` so at that point node values should all be added from [this](https://github.com/zwave-js/zwavejs2mqtt/blob/10143d6ddedf2ffb1e9288f74771fd7a00aefc0a/lib/ZwaveClient.js#L275) loop. Then when `nodeStatus` is called the gateway see that the node is ready and then it discover all values

AlCalzone:
Can someone give me a gist of whats the issue here? Which values are missing and when?

billiaz:
replaced by https://github.com/zwave-js/zwavejs2mqtt/pull/54

PR #54, opened 2020-12-07
fix: discovery missing targetValue when runs first time

billiaz:
During initial discovery, while the setTopic value is undefined there is
an issue which arises with an nill value for command_topic on hass
discovery. This issue comes due to the "split" node compared to previous
driver, which combined both.

Check if node is ready before discovering


robertsLando (approved):
LGTM! Thanks @billiaz :)

PR #56, opened 2020-12-07
fix: Remove Refresh values button

billiaz:
Remove the Refresh values button under Configuration tab of a node.
This function is not working. Just commented out the section


robertsLando (approved):
LGTM

PR #58, opened 2020-12-08
try to fix undefined variable, due to one mising 'e'

billiaz:
A typo, causes no subscription topics to be created in MQTT. this fixes the issue


robertsLando (approved):
LGTM

PR #67, opened 2020-12-11
feat: move to winston logger

ahochsteger:
This PR **unifies** the **logging** across zwavejs2mqtt using **[winston](https://github.com/winstonjs/winston)**. 

**Usage:**

image

```javascript
// Simple logger with module name and default config:
const logger = require('./logger.js').module('Zwave')

// Logger setup with module name and specific config:
const logger = require('./logger.js').module('Zwave').setup(
  { enabled: true, level: 'info', logToFile: true, filename: "zwavejs2mqtt.log" }
)

// Do some logging:
logger.info(`Some log message with a ${value}`)
logger.warn('Some warning')

// Reconfigure an existing logger:
logger.setup({ enabled: true, level: 'info', logToFile: true, filename: "zwavejs2mqtt.log" })

// Reconfigure loggers for all modules (e.g. after config changes in the UI):
const loggers = require('./logger.js')
loggers.setupAll({ enabled: true, level: 'info', logToFile: true, filename: "zwavejs2mqtt.log" })
```

**Conditions & Requirements:**

- [x] The logging solution should be as simple as possible while supporting the other requirements
- [x] Winston should be used the basis for unified logging
- [x] All loggers should be defined on a central place (logger.js)
- [x] The primary logging format should be: `<timestamp> <level> <moduleName/label>: <logMessage>` (e.g. `2020-12-04 09:06:32.568 INFO APP: Zwave2MQTT Version 0.0.0`)
- [x] All loggers should use the same transports
- [x] All loggers should use the same format by default
- [x] Loggers for different modules may use a different logging configuration
- [x] Different log levels should be used to signal event severity (see <https://github.com/winstonjs/winston#logging-levels>)
- [x] Modules/labels should be as short as possible. The following should be used: `ZWAVE` `MQTT` `GATEWAY` `STORE` `HASS` `HTTP` `APP`
- [x] Log statements with multiple arguments using commas should be converted to the template syntax
- [x] The logging must be changeable during runtime from the UI
- [x] It should be able to change the configuration of all loggers at once (to be used by UI for central logging config changes)
- [x] Runtime changes to the logging settings must be persisted to `settings.js`
- [x] The logging must be controlled by `settings.js` using the existing options `logLevel` and `logToFile`
- [x] Add new setting `logEnabled` to `settings.js` and UI

**Required changes to the existing codebase:**

The following files are affected and show the existing logging implementation that should be migrated to the new logging module:

- [x] `lib/ZwaveClient.js` (uses debug) --> winston  `ZWAVE` module
- [x] `app.js` (uses debug & morgan) ---> winston `HTTP` and `APP` modules
- [x] `lib/jsonStore.js` (uses debug) ---> winston `STORE` module (need to think about this, we could use console otherwise)
- [x] `lib/Gateway.js` (uses debug) ---> winston `GATEWAY` and (maybe) `HASS` modules
- [x] `lib/MqttClient.js` (uses debug) --> winston `MQTT` module
- [x] `lib/SocketManager.js` (uses debug) --> winston `SOCKET` module
- [x] `lib/debug.js` (uses debug) ---> Obsolete, replaced with `logger.js`


lib/ZwaveClient.js
```diff
@@ -17,7 +17,7 @@ const { socketEvents } = reqlib('/lib/SocketManager.js')
 const store = reqlib('config/store.js')
 const storeDir = utils.joinPath(true, reqlib('config/app.js').storeDir)
 const debug = reqlib('/lib/debug')('Zwave')
-const logger = require('./logger.js').module('ZWC')
+const logger = require('./logger.js').setup('zwc')
```
robertsLando, line 20: I would prefer to use `module` to get the module and a separete function `setup` to edit the general configuration.

lib/logger.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Format loglevel upper-case:
+const upperCaseLevel = format(info => {
+  info.level = info.level.toUpperCase()
+  return info
+})
+
+// Custom combined logging format:
+const customFormat = config =>
```
robertsLando, line 33: Does this handles errors? I mean if I log an error is the stack trace printed to log output?

ahochsteger: I guess something like `errors({ stack: true })` will have to be added to the format as seen here:
<https://stackoverflow.com/questions/47231677/how-to-log-full-stack-trace-with-winston-3>
I'll add it.

ahochsteger: @robertsLando It's done and I've added 3 example usages in ZwaveClient.js.
The 2nd `error` parameter to the log statements may be removed again, if we don't require a stack trace for these logs.

lib/ZwaveClient.js
ahochsteger: Do you mean something like this?
```javascript
const logger = require('./logger.js').module('Zwave').setup(
  { enabled: true, level: 'info', logToFile: true, filename: "zwavejs2mqtt.log" }
)
```

robertsLando: @ahochsteger Yep and we should decide where and when we will call the setup. It should be called only on application start and when settings change. By default module will return the logger with the last setup configuration.

ahochsteger: @robertsLando it's done now and can be used like this to dynamically reconfigure the logger:

```javascript
let logger = require('./logger.js').module('Zwave')
log.info('This will be shown by default')
logger = logger.setup({ level: 'warn' })
log.info('Now this won't be shown anymore')
```

Since Winston does not seem to support the reconfiguration of an existing logger during runtime I had to do some workarounds which I'm not really happy about yet:
`setup()` closes an existing logger with the given module name and creates a completely new one, thus all references will change and `const logger = ...` will have to be replaced with `let logger = ...`.

Maybe somebody with more Winston knowledge is able to find a better way.

ahochsteger:
@robertsLando the previous debug-based logging is now fully replaced with the new winston-based logging module.

The only thing that is intentionally left is integration into the UI and control using settings.json which requires some discussion about the best approach to propagate logging config changes to the module loggers.

lib/ZwaveClient.js
ahochsteger: Update: Have a look at the section *Open questions* in the description for possible solutions.

ahochsteger: Another update: Winston supports reconfiguration of existing loggers using `logger.configure()` which I was not aware before. That way it was really simple to do it the way we need it.

ahochsteger:
@robertsLando after some more work I was able to resolve all open questions and handle the logging configuration in app.js (see `setupLogging()`).
Changes in the UI now automatically update the logger of all modules.

But there's a side-effect I observed I was not yet able to solve yet:
After changing the logging settings in the UI and restarting the server there are no more logs available from `node-zwave-js`. Maybe there's a conflict somewhere? At least I tried to not touch other Winston loggers as much as possible (e.g. by prefixing logger names with "zwavejs2mqtt." and only handle these in `setupAll()`).

So from my POV the PR is ready for review.

robertsLando:
> After changing the logging settings in the UI and restarting the server there are no more logs available from node-zwave-js. Maybe there's a conflict somewhere

@AlCalzone Do you know if this could cause some conflict with your loggers?

app.js
```diff
@@ -315,10 +328,10 @@ app.use(function (err, req, res) {
 process.removeAllListeners('SIGINT')
 
 process.on('SIGINT', function () {
-  debug('Closing clients...')
+  logger.info('Closing clients...')
   gw.close()
     .catch(err => {
-      debug('Error while closing clients', err)
+      logger.error('Error while closing clients', err)
```
robertsLando, line 334: @ahochsteger How does errors should be logged? I mean can we add error as second parameter and use the first one as a prefix message?

lib/Gateway.js
```diff
@@ -1146,11 +1144,8 @@ Gateway.prototype.discoverDevice = function (node, hassDevice) {
       }
     }
   } catch (error) {
-    debug(
-      'Error while discovering device %s of node %d: %s',
-      hassID,
-      node.id,
-      error
+    logger.error(
+      `Error while discovering device ${hassID} of node ${node.id}: ${error}`
```
robertsLando, line 1148: Print also the error stack, this could make easier to undestand where the discovery is broken

lib/Gateway.js
```diff
@@ -1611,11 +1606,8 @@ Gateway.prototype.discoverValue = function (node, vId) {
 
     this.publishDiscovery(cfg, node.id)
   } catch (error) {
-    debug(
-      'Error while discovering value %s of node %d: %s',
-      valueId.id,
-      node.id,
-      error.message
+    logger.error(
+      `Error while discovering value ${valueId.id} of node ${node.id}: ${error.message}`
```
robertsLando, line 1610: Add stack trace here

lib/Gateway.js
```diff
@@ -1274,11 +1269,11 @@ Gateway.prototype.discoverClimates = function (node) {
     // discovered later when we call `discoverDevice`
     nodeDevices.push(config)
 
-    debug('New climate device discovered: ' + JSON.stringify(config))
+    logger.info(`New climate device discovered: ${JSON.stringify(config)}`)
 
     allDevices[node.deviceId] = nodeDevices
   } catch (error) {
-    debug('Unable to discover climate device: ' + error.message)
+    logger.error(`Unable to discover climate device: ${error.message}`)
```
robertsLando, line 1276: Print also stack trace here

robertsLando (changes requested):
@ahochsteger Well done! Actually you are using zwave configuration for logging but I would use separete settings to configure z2m loggger (some user may be interested just on one or the other). Did you also checked if the interceptor works? The method used in app.js to intercept logs and send them to socket, I think we could create a custom transport for that instead

AlCalzone:
> Do you know if this could cause some conflict with your loggers?

Not sure... you're also using `winston.loggers`, maybe if you're overwriting a logger name that zwave-js uses, e.g.
https://github.com/zwave-js/node-zwave-js/blob/daf0feb9bfe14ae5b9a24c0d0d53206d5bdd41e9/packages/zwave-js/src/lib/log/Driver.ts#L33

lib/logger.js
```diff
@@ -0,0 +1,85 @@
...
+      })
+    )
+  }
+  return transportsList
+}
+
+// Helper function to setup a logger:
+const setupLogger = (container, module, config) => {
+  config = sanitizedConfig(module, config)
+  const logger = container.add(loggerPrefix + module) // NOTE: Winston automatically reuses an existing module logger
```
robertsLando, line 58: @AlCalzone We are using a custom prefix for loggers so we shouldn't touch yours

robertsLando: Also we are using a custom containers so I even don't think we have any access to your loggers

AlCalzone: Then 🤷🏻‍♂️

lib/Gateway.js
ahochsteger: done.

lib/Gateway.js
ahochsteger: done.

lib/Gateway.js
ahochsteger: done.

lib/logger.js
ahochsteger: FYI: In the last iteration I changed `module()` to directly return `winston.loggers`, since I prefixed the loggers anyway to separate them and the unwanted behavior was still the same using `new winston.Container()`.

@AlCalzone, do you have any pointers for me where I could best hook into the logging setup of `node-zwave-js`? I want to find out the cause of this, since it may bite us somewhere else too ;-).

ahochsteger:
> @ahochsteger Well done! Actually you are using zwave configuration for logging but I would use separete settings to configure z2m loggger (some user may be interested just on one or the other).

Yes, I'm aware, that this is not yet the final way but I wanted to take small working steps and the current solution is minimal invasive concerning the handling of `settings.js`.
Should there be a `logging` section in `settings.js` on the same level as `mqtt` and `zwave`? Or do you think of something else?

> Did you also checked if the interceptor works?

Not yet, but looks interesting.
I'll have a look ...

> Does the configure method still creates problems with zwavejs? Even with prefix?

Unfortunately yes (see above).
IMO I'd like to solve this issue and then merge the changes ASAP, since the rest of functionality it could be added independently.
In theory this could also be the case for changing `settings.js` for logging.

lib/logger.js
ahochsteger: I found the reason of this behavior and it has nothing to do with this PR since it is already behaving that way on master branch.
The responsible lines are in `ZwaveClient.js` which seem to intentionally control the logging behavior of zwave-js using environment variables:
```javascript
  if (process.env.LOGLEVEL === 'undefined') process.env.LOGLEVEL = null
  process.env.LOGLEVEL = process.env.LOGLEVEL || config.logLevel
  if (process.env.LOGTOFILE || config.logToFile) {
    process.env.LOGTOFILE = 'true'
  }
```
@robertsLando I'm not really sure, what your goal is for this with respect to the configuration in `settings.js`?
Currently there's a logging configuration in the `zwave` config, that controls the logging of zwave-js. Should this be kept as is or fully replaced with the central logging configuration you mentioned to be placed into the `gateway` section?

robertsLando: > The responsible lines are in ZwaveClient.js which seem to intentionally control the logging behavior of zwave-js using environment variables:

robertsLando (dismissed):
LGTM

robertsLando:
@ahochsteger Review done. Feel free to squash and merge if you are done here :smile:

src/components/Settings.vue
```diff
@@ -320,6 +320,27 @@
...
                         ></v-switch>
                       </v-flex>
+                      <v-flex xs12 sm6>
+                        <v-switch
+                          hint="Enable logging"
+                          persistent-hint
+                          label="Log enabled"
+                          v-model="gateway.logEnabled"
+                        ></v-switch>
+                        <v-switch
```
robertsLando, line 330: This should be visible just when logging is enabled

src/components/Settings.vue
```diff
@@ -320,6 +320,27 @@
...
+                          hint="Store zwave logs in a file"
+                          persistent-hint
+                          label="Log to file"
+                          v-model="gateway.logToFile"
+                        ></v-switch>
+                      </v-flex>
+                      <v-flex xs12 sm6>
```
robertsLando, line 337: This should be visible just when logging is enabled

robertsLando (changes requested):

src/components/Settings.vue
ahochsteger: done.

src/components/Settings.vue
ahochsteger: done.

lib/logger.js
ahochsteger: @robertsLando any feedback on how to proceed with the zwave-js logging, since I saw that there are TODOs with a PR mentioned (not sure, if you were referring to this PR here)?
Here's the complete code:
```javascript
function ZwaveClient (config, socket) {
  ...
  // TODO: Handle this by using `logOptions` (PR waiting to be merged)
  if (process.env.LOGLEVEL === 'undefined') process.env.LOGLEVEL = null

  // https://github.com/zwave-js/node-zwave-js/blob/master/packages/core/src/log/shared.ts#L13
  // https://github.com/winstonjs/triple-beam/blob/master/config/npm.js#L14-L15
  process.env.LOGLEVEL = process.env.LOGLEVEL || config.logLevel

  if (process.env.LOGTOFILE || config.logToFile) {
    process.env.LOGTOFILE = 'true'
  }
  ...
```

robertsLando: Nope, we need to wait zwave-js 6.0.0 for that. Actually that 'workaround' is the only way to configure zwave-js logging

src/components/Settings.vue
```diff
@@ -320,6 +320,29 @@
...
+                          v-if="gateway.logEnabled"
+                          hint="Store logs in a file"
+                          persistent-hint
+                          label="Log to file"
+                          v-model="gateway.logToFile"
+                        ></v-switch>
+                      </v-flex>
+                      <v-flex xs12 sm6>
+                        <v-select
+                          v-if="gateway.logEnabled"
```
robertsLando, line 340: move the v-if to the parent v-flex

src/components/Settings.vue
```diff
@@ -320,6 +320,29 @@
...
                         ></v-switch>
                       </v-flex>
+                      <v-flex xs12 sm6>
+                        <v-switch
+                          hint="Enable gateway logging"
+                          persistent-hint
+                          label="Log enabled"
+                          v-model="gateway.logEnabled"
+                        ></v-switch>
+                        <v-switch
```
robertsLando, line 330: This should be moved to another v-flex container

robertsLando (changes requested):

src/components/Settings.vue
ahochsteger: done.

src/components/Settings.vue
ahochsteger: done.

robertsLando (commented):
@ahochsteger I would like a cleaner output format. The color should be applied only to level and the timestamp could be gray. It is much pretty to read

robertsLando:
I also think you have missed `stderrLevels: ['error']` in console transport. If user is using pm2 that will redirect errors to stderror

robertsLando:
My changes: 

```js
const winston = require('winston')
const { format, transports, addColors } = winston
const { combine, timestamp, label, printf, errors, colorize } = format
const loggerPrefix = 'zwavejs2mqtt.'

const colorizer = colorize()

// Make sure every config value is defined
const sanitizedConfig = (module, config) => {
  config = config || {}
  return {
    module: module !== undefined ? module : '-',
    enabled: config.enabled !== undefined ? config.enabled : true,
    level: config.level !== undefined ? config.level : 'info',
    logToFile: config.logToFile !== undefined ? config.logToFile : false,
    filename:
      config.filename !== undefined ? config.filename : 'zwavejs2mqtt.log'
  }
}

addColors({
  time: 'grey',
  module: 'bold'
})

// Custom logging format:
const customFormat = config =>
  combine(
    errors({ stack: true }),
    timestamp({ format: 'YYYY-MM-DD HH:mm:ss.SSS' }),
    format(info => {
      info.level = info.level.toUpperCase()
      return info
    })(),
    label({ label: config.module.toUpperCase() }),
    colorize({ level: true }),
    printf(
      info =>
        `${colorizer.colorize('time', info.timestamp)} ${info.level} ${colorizer.colorize('module', (info.label || '-'))}: ${info.message}`
    )
  )

// Custom transports:
const customTransports = config => {
  const transportsList = [
    new transports.Console({
      format: format.colorize({ level: true }),
      level: config.level,
      stderrLevels: ['error']
    })
  ]
  if (config.logToFile) {
    transportsList.push(
      new transports.File({
        format: customFormat(config),
        filename: config.filename,
        level: config.level
      })
    )
  }
  return transportsList
}
```

This creates a pretty print: 

Schermata da 2020-12-15 15-01-43

lib/logger.js
```diff
@@ -0,0 +1,98 @@
...
+// Setup a logger for a certain module:
+logContainer.loggers.module = module => {
+  return setupLogger(logContainer, module)
+}
+
+// Setup loggers for all modules
+logContainer.loggers.setupAll = config => {
+  for (const [key, logger] of logContainer.loggers) {
+    // Make sure, only winston loggers created from this logger module are modified:
+    if (key.startsWith(loggerPrefix)) {
```
robertsLando, line 92: I think we can remove this now, it's useless

src/components/Settings.vue
```diff
@@ -320,6 +320,27 @@
...
+                          v-model="gateway.logLevel"
+                          label="Log Level"
+                        ></v-select>
+                        <v-switch
+                          hint="Store logs in a file"
+                          persistent-hint
+                          label="Log to file"
+                          v-model="gateway.logToFile"
+                        ></v-switch>
+                      </v-flex>
```
robertsLando, line 343:
```suggestion
                      <v-flex xs12 sm6 v-if="gateway.logEnabled">
                        <v-select
                          :items="logLevels"
                          v-model="gateway.logLevel"
                          label="Log Level"
                        ></v-select>
                       </v-flex>
                       <v-flex xs12 sm6 v-if="gateway.logEnabled">
                        <v-switch
                          hint="Store logs in a file"
                          persistent-hint
                          label="Log to file"
                          v-model="gateway.logToFile"
                        ></v-switch>
                      </v-flex>
```

robertsLando (changes requested):

robertsLando (approved):
LGTM

PR #76, opened 2020-12-15
Feature: Add Export button to Nodes for Debug reasons

billiaz:
I added an Export option for nodes.. Currently this export button can help on debugging issues.

Info:

- Per Node Export
- Discreet location no right side, next to node ID
- Output is Json format

Export Screenshot:
exportNode


robertsLando (approved):
LGTM. Thanks @billiaz :)

PR #86, opened 2020-12-18
[feat] add ability to update devices of driver during build

billiaz:
Nice to have posibility to update on docker image build time devices!


robertsLando:
@billiaz I think this is something that should be implemented on zwavejs side first

billiaz:
@robertsLando i do not disagree,  till then, having a docker file, able with an argument to update devices is nice feature
By default is disabled, u les you set a value on build argukent deviceUpdate.

AlCalzone:
I'm still planning to auto-update `@zwave-js/config`, but until that is implemented, I think we can do it during the build.

billiaz:
The change I propose is not documented well.

More details:
- Following the normal build, without modifying any argument. It will build a container without updating the database
- Adding a `--build-arg updateDevices=on` with any value instead of `on` it will trigger the cloning of the repository under the zwavejs argument
- We can override the target repository by also adding the argument `--build-arg zwavejs=<git endpoint>`. This solves the reason I though of adding this option. I can this way Debug device additions without the need of compiling the zwave-js module.

I find it helpful for people (with or without implemented the auto-update happening through driver, to give the ability on people to participate in these projects.

robertsLando (approved):
LGTM

Issue #94, opened 2020-12-20
[bug] 113-0-Access Control-Door state not sending MQTT message (spaces in topic?)


Lopton:
**Version**
Build/Run method
- [ X ] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

zwavejs2mqtt version: 1.0.0-alpha 

**Describe the bug**
When my Aeotec door sensor reports a notification that the door is open, no MQTT message appears to be generated.

**To Reproduce**
1. Add Aeotec Door Sensor 7
2. Open door, observe debug log message come in "INFO ZWAVE: Node 17: value updated: 113-0-Access Control-Door state", while monitoring MQTT Explorer you will see no MQTT message generated.

**Expected behavior**
If 113-0-Access Control-Door State is changed I would expect an MQTT message to be produced.

**Additional context**
This may be partially related to #91, because originally the door notification would not be discovered. So I manually modified the payload to allow for discovery. It was only then that I figured out that the mqtt message for the door opening also didn't seem to be being generated. However I may be misunderstanding how this works. Let me know if I can provide more information. 
When I monitor the 133->0 topics I only see "Power_Management" and "System". "Access Control" never appears.

robertsLando:
Thanks for the issue @Lopton ! Will fix it tomorrow :)

billiaz:
I am already testing a small fix @robertsLando please check my pull request once i post it for comments

Issue #91, opened 2020-12-18
[bug] Entities missing in Home Assistant


oxgl:
Build/Run method
- [X] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

zwavejs2mqtt version: 1.0.0-alpha.1

**Describe the bug**
Some entities are missing in Home Assistant. The problem is that zwavejs2mqtt is creating topics in MQTT with spaces in topic name. As an example, one of sensor configurations is under topic
`homeassistant/sensor/n015_ws_boilerw/notification_Water Alarm/config` (space between Water and Alarm). 
If I replace the space with underscore in the topic name: `homeassistant/sensor/n015_ws_boilerw/notification_Water_Alarm/config`, the entity shows up in home assistant. It would be also nice to have topic name all lower case.
Thank you!

billiaz:
I will test this

Issue #60, opened 2020-12-08
[bug] Hass Discovery Motion Sensor & Cover Status name issues


chilicheech:
**Version**

Build/Run method
- [x] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

zwavejs2mqtt version: master

**Describe the bug**
Some devices create `113-0-Home_Security-Motion_sensor_status` and `113-0-Home_Security-Cover_status` values. Hass discovery creates `sensor_notification_Home_Security` for motion and `sensor_notification_Home_Security_Home Security` for cover. These names could be better represented. Maybe as `sensor_notification_home_security_motion` and `sensor_notification_home_security_cover` or something in those lines. I'm not sure what the appropriate name should be. I can look at what z2m/ozw creates and post them here for reference.

![Screen Shot 2020-12-08 at 11 12 14 AM](https://user-images.githubusercontent.com/732659/101530546-e8bb1200-3946-11eb-8920-fa15360c9847.png)

The other issue is that the payload also has those names and they're not good names. They should also be improved:
Motion sensor:
```
{
  "type": "sensor",
  "object_id": "notification_Home Security",
  "discovery_payload": {
    "value_template": "{{ value_json.value }}",
    "icon": "mdi:alarm-light",
    "state_topic": "z2m/24/113/0/Home_Security/Motion_sensor_status",
    "json_attributes_topic": "z2m/24/113/0/Home_Security/Motion_sensor_status",
    "device": {
      "identifiers": [
        "zwavejs2mqtt_0xc5b5c9e5_node24"
      ],
      "manufacturer": "Vision Security",
      "model": "Zooz 4-in-one motion/temperature/humidity/luminance sensor (ZSE40)",
      "name": "master",
      "sw_version": "5.1"
    },
    "name": "master_notification_Home Security",
    "unique_id": "zwavejs2mqtt_0xc5b5c9e5_24-113-0-Home Security-Motion sensor status"
  },
  "discoveryTopic": "sensor/master/notification_Home Security/config",
  "values": [
    "113-0-Home Security-Motion sensor status"
  ],
  "persistent": false,
  "ignoreDiscovery": false,
  "id": "sensor_notification_Home Security"
}
```
Cover sensor:
```
{
  "type": "sensor",
  "object_id": "notification_Home Security_Home Security",
  "discovery_payload": {
    "value_template": "{{ value_json.value }}",
    "icon": "mdi:alarm-light",
[...]

PR #97, opened 2020-12-20
[fix] HassDiscovery payload

billiaz:
- Convert to lower case
- Replace spaces with underscore

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -566,11 +566,7 @@ function deviceInfo (node, nodeName) {
  * @returns The topic string for this device discovery
  */
 function getDiscoveryTopic (hassDevice, nodeName) {
-  return `${
-    hassDevice.type
-  }/${nodeName}/${hassDevice.object_id
-    .toLocaleLowerCase()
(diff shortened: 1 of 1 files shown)


billiaz:
fixes issues 
https://github.com/zwave-js/zwavejs2mqtt/issues/94
https://github.com/zwave-js/zwavejs2mqtt/issues/91
https://github.com/zwave-js/zwavejs2mqtt/issues/60

robertsLando:
@billiaz I would instead fix the object id. Let me show you

billiaz:
> @billiaz I would instead fix the object id. Let me show you

Thinking of it.
If we check object_id after is defined and change it, should work right? for instance 

To be clear:
1. create a function which converts any string passed to lowercase and replace spaces with underscores.
2. Update object_id earlier in code, where object_id is defined, using the function
3. have a function for other uses too (the 1 point)

@robertsLando something like that?

robertsLando:
@billiaz FIxed in #99 . Thanks for pointing out the origin of the fix :smile:

PR #98, opened 2020-12-20
[fix] Detail on naming - Notifications  Names

billiaz:
Multiple times happens on notification names during discovery to have a name like:
```
sensor_notification_Power Management_Power Management
```
It would be better if these names where:
```
sensor_notification_Power Management_Motion
```
This tries to avoid using the "duplicate" name by setting on Notifications the propertyKey

Also this happens with betteries, and other endpoints. when duplication happens.  It is good idea to use endpoint number as extension instead of "repeating" the same property name.


robertsLando (approved):
LGTM. Thanks @billiaz :tada:

PR #99, opened 2020-12-21
fix: discovery issues caused by spaces in topic

### LINKED ISSUE #91: [bug] Entities missing in Home Assistant

robertsLando:
Fixes #91 #60 #94

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -570,6 +570,16 @@ function getDiscoveryTopic (hassDevice, nodeName) {
 }
 
 /**
+ * Sanitize ids removing chars that could break discovery
+ *
+ * @param {String} id The id string
+ * @returns The sanitized id, lower cases and without spaces
+ */
+function sanitizeId (id) {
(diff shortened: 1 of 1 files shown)


robertsLando:
@billiaz Check this

robertsLando:
This is a breaking change for hass users, many entities will have a different unique id

billiaz:
tested works like charm

PR #100, opened 2020-12-21
feat(hass): entities name template configuration

billiaz:
I always struggled with the idea of having underscores and modifying the names of my devices through Hass.

Looking at the current options, I thought it might be possible for the user to decide the pattern of each entity which is generated by zj2m

What I did till now is:
1. add option when Hass Discovery is enabled to choose between 4 patterns (default is always the current, even if value is undef)
2. create a function for this option, to return the pattern. As this is used in two places
3. replace payload.name on both places to use this function.

What I miss:
1. probably there is a better way to pass the values on my function (especially the `this.` value) 
2. Possible new combinations useful for users
3. possible nitty gritty stuff i haven't thought of
4. Documentation once we are sure we like the feature and we have some standards

A device with
Node Location: LivingRoom
Node Name: Sensor
object_id: Battery Level

Current options:
- NodeName_object_id (value: LivingRoom-Sensor_battery_level)
- node.Loc node.name object_id(spaces) (value: LivingRoom Sensor battery level)
- node.name node.loc object_id(spaces) (value: Sensor LivingRoom battery level)
- object_id(spaces) (value: battery level)

More to add:
- [wip] node.propertyName node.propertyKey(if exists) (value example: Motion Home Security)
- [wip] node.propertyKey node.propertyName (value example: Home Security Motion status)


src/components/Settings.vue
```diff
@@ -549,6 +558,24 @@ export default {
           value: 2
         }
       ],
+      hfn_types: [
+        {
+          text: '<Location>-<name>_<object_id>',
```
robertsLando, line 563:
```suggestion
          text: '<location>-<name>_<object_id>',
```

src/components/Settings.vue
```diff
@@ -549,6 +558,24 @@ export default {
...
+          value: 0
+        },
+        {
+          text: '<Location> <name> <object_id(spaces)>',
```
robertsLando, line 567:
```suggestion
          text: '<location> <name> <object_id(spaces)>',
```

src/components/Settings.vue
```diff
@@ -549,6 +558,24 @@ export default {
...
+          value: 1
+        },
+        {
+          text: '<name> <Location> <object_id(spaces)>',
```
robertsLando, line 571:
```suggestion
          text: '<name> <location> <object_id(spaces)>',
```

robertsLando (changes requested):
By default now all object ids are sanitized, so it's useless to specify spaces/no spaces

Also in options you should add a default one (value `0`) that equals to the current behaviour

billiaz:
> By default now all object ids are sanitized, so it's useless to specify spaces/no spaces
> 
> Also in options you should add a default one (value `0`) that equals to the current behaviour

I do not know how to do this in Vue :) struggling

robertsLando:
> I do not know how to do this in Vue :) struggling

Just add a new option to `hfn_items` with value `undefined` and text `Default`

robertsLando:
@billiaz I agree that it looks much cleaner now, we should so replace _ with spaces there

robertsLando:
@billiaz I agree it's much cleaner now, but I would keep the object id with `_` just to don't do too much replaces. Anyway see `battery_level` there would be fine too IMO

billiaz:
> @billiaz I agree it's much cleaner now, but I would keep the object id with `_` just to don't do too much replaces. Anyway see `battery_level` there would be fine too IMO

I still would like as user to have an option to get read of spaces, or even have more combinations.
I am ok with the default, but generally if we keep `_` on object. I will choose the space.
Also on my suggestion, We leave to the User the option to use spaces or not. maybe adding more options on list people can choose if they want to "overdo" the replacesments or not.

The discovery runs once on startup, or when changes happen and this should not cause so much load.

I suggest the following list:
- NodeName object_id (value: LivingRoom-Sensor battery_level) (default, remove underscoer between nodeName and object_id)
- `node.Loc node.name object_id` (value: LivingRoom Sensor battery_level) # same as default but with space on loc/name
- `node.name node.loc object_id` (value: Sensor LivingRoom battery_level) # (reverse)
- `node.Loc node.name object_id(noUnderscores)`  # (value: LivingRoom Sensor battery level) 
- `node.name node.loc object_id(noUnderscores)` # (value: Sensor LivingRoom battery level) (reverse)
- `object_id # (value: battery_level)
- `object_id(noUnderscores) # (value: battery level)
- `node.propertyName node.propertyKey(if exists)` (value example: Motion Home Security)
- `node.propertyKey node.propertyName` (value example: Home Security Motion status)

Probably i can use the propertyName/Key to replace object_id on the non underscored ones!

This is the thinking from user prospective. I know might be to many options. But makes it more attractive

robertsLando:
@billiaz How do you handle nodes that doesn't have a name/location or both? 

I would like to avoid adding configurations when it's not needed, them create confusion to users, in this case I would prefer to create a default one instead that would be fine for most users

billiaz:
> @billiaz How do you handle nodes that doesn't have a name/location or both?
> 
> I would like to avoid adding configurations when it's not needed, them create confusion to users, in this case I would prefer to create a default one instead that would be fine for most users

I would use in this case the NodeID as it is on the nodeName. I get your point.
I suspect limiting the list to 2-3 options "unbreakable" would be nice
Though living to users some options is a good idea. Additionaly instead of using object_id we could use the PropertyKey and PropertyName which are the real nice looking value.


looking valueId:
```ccSpecific:{sensorType: 4, scale: 0}
commandClass:49
commandClassName:'Multilevel Sensor'
default:undefined
description:undefined
endpoint:0
genre:'user'
id:'64-49-0-Power'
label:'Power'
lastUpdate:1608547182790
list:false
max:undefined
min:undefined
nodeId:64
property:'Power'
propertyKey:undefined
propertyName:'Power'
readable:true
step:undefined
type:'number'
unit:'W'
value:85.4
writeable:false
__proto__:Object```

- `NodeName object_id` (value: LivingRoom-Sensor battery_level) (default, remove underscoer between nodeName and object_id)
- `NodeName(noHyphen) object_id(noUnderscores) # (value: LivingRoom Sensor battery level, OR Node_88 Sensor battery level)
- `object_id` # (value: battery_level)
- `propertyKey or propertyName(if key undefined)` # (value when PropertyKey is missing: battery level) (value when propertyName is there: Motion sensor Status)
- `NodeName propertyKey or NodeName propertyName(if key undefined)` # (value when PropertyKey is missing: battery level) (value when propertyName is there: LivingRoom-Sensor Motion sensor Status)
- previous with no hyphen

I understand can be confusing. I believe the above covers more cases to avoid undefined names. what you think?

File: lib/Gateway.js [hunk 4]
```diff
@@ -621,6 +621,36 @@ function setDiscoveryValue (payload, prop, node) {
...
+      return `${nodeName} ${
+        valueId.propertyKey ? valueId.propertyKey : valueId.propertyName
+      }`
+    case 4:
+      // Use PropertyKey, if not exists fall back to propertyName
+      return `${nodeName.replace('-', ' ')} ${
+        valueId.propertyKey ? valueId.propertyKey : valueId.propertyName
+      }`
+    default:
+      return `${nodeName} ${cfg.object_id}`
```
robertsLando, line 647: Default should be with `_`

lib/Gateway.js
```diff
@@ -1130,7 +1160,12 @@ Gateway.prototype.discoverDevice = function (node, hassDevice) {
         hassDevice.object_id = sanitizeId(hassDevice.object_id)
 
         // Set a friendly name for this component
-        payload.name = `${nodeName}_${hassDevice.object_id}`
+        payload.name = setHassFriendlyName(
+          this.config.hassFriendlyName,
+          hassDevice,
+          valueId,
```
robertsLando, line 1162: `valueId` doesn't exists

robertsLando (changes requested):
I'm still not 100% sure about this. Couldn't you find a unique friendly name that will work for most users out of the box?

larstobi:
For me, I don't like prefixing the device name with location. I have renamed my devices in zj2m, but I have explicitly not set locations to any devices in zj2m, since they get prefixed and I can't see an option to remove that.

Also, in order to not have to specify room twice, since Home Assistant isn't able to detect the location/room, it would be cool if Home Assistant could read location from the JSON payload and just use that.

I don't personally have an issue with underscores in the naming, but I can see that it looks prettier without. However, if it gets really long it also starts to look like a sentence instead of a name.

billiaz:
> For me, I don't like prefixing the device name with location. I have renamed my devices in zj2m, but I have explicitly not set locations to any devices in zj2m, since they get prefixed and I can't see an option to remove that.
> 
> Also, in order to not have to specify room twice, since Home Assistant isn't able to detect the location/room, it would be cool if Home Assistant could read location from the JSON payload and just use that.
> 
> I don't personally have an issue with underscores in the naming, but I can see that it looks prettier without. However, if it gets really long it also starts to look like a sentence instead of a name.

I agree with you, although hass has some plugins to setup locations
The proposal will provide with a dropdown, which will allow you to "shorten names" Also some options try to use the shortest path. 

After talking with @robertsLando and Oxgl, but also taking into consideration your thought

btw payload has nodeLocation and nodeName exposed in json these days (previous patch)

A smaller list of targetted options could be:

1. Default
2. PropertyKey else PropertyName
3. Name with PropertyKey (else PropertyName)
4. NodeName with PropertyKey (else PropertyName) 


The above woud look like:
Binary sensor:
1. garage-sensor_battery_level
2. battery level
3. Sensor battery level
4. Garage-Sensor Battery level

Notification CC  Sensor:
1. livingroom-multisensor_home_security_motion_sensor_status
2. Motion sensor status
3. MultiSensor Motion sensor status
4. Livingroom-MultiSensor Motion sensor status
 
In case Name is empty, we will use the nodeId_<num>

@robertsLando what you think?

robertsLando:
Let's see also what @chilicheech thinks about this

robertsLando:
I'm ok to whatever makes most users happy in general, this is why I would like to collect some users opinions about this to find the best solution. As @larstobi said too long names wouldn't be so good

billiaz:
> I'm ok to whatever makes most users happy in general, this is why I would like to collect some users opinions about this to find the best solution. As @larstobi said too long names wouldn't be so good

To reduce confusion would be nice to have 2 options
Configurable Node Name:
- Location-Name (fallback to NodeId)
- Name only (fallback to NodeId)

Above:
- Default
- PropertyKey (if undefined PropertyName)
- NodeName with PropertyKey (else PropertyName)

This way users can choose to expose or not location to name and then choose only namings, which makes it super short! :)
and pretty flexible but not confusing

robertsLando:
@billiaz Ok go for it :)

lib/Gateway.js
```diff
@@ -621,6 +621,33 @@ function setDiscoveryValue (payload, prop, node) {
...
+ */
+function setHassFriendlyName (value, cfg, nodeName, valueId) {
+  const id = valueId ? valueId.label : cfg.object_id
+  switch (value) {
+    case 0:
+      return `${nodeName.replace('-', ' ')} ${cfg.object_id.replace(/_/g, ' ')}`
+    case 1:
+      return cfg.object_id
+    case 2:
+      // Use PropertyKey, if not exists fall back to propertyName
```
robertsLando, line 639:
```suggestion
      // Use propertyKey, if not exists fall back to propertyName
```

lib/Gateway.js
```diff
@@ -621,6 +621,33 @@ function setDiscoveryValue (payload, prop, node) {
...
+      return id
+    case 3:
+      // Use PropertyKey, if not exists fall back to propertyName
```
robertsLando, line 642:
```suggestion
      // Use propertyKey, if not exists fall back to propertyName
```

lib/Gateway.js
```diff
@@ -621,6 +621,33 @@ function setDiscoveryValue (payload, prop, node) {
...
+      return `${nodeName} ${id}`
+    case 4:
+      // Use PropertyKey, if not exists fall back to propertyName
```
robertsLando, line 645:
```suggestion
      // Use propertyKey, if not exists fall back to propertyName
```

robertsLando (changes requested):
Just some typos, the code is good :)

aretakisv:
> @billiaz Ok go for it :)

you mean to create two switches
1. for nodeName
2. for the payload.name?

robertsLando:
> for the payload.name?

What? 

I'm ok with this: https://github.com/zwave-js/zwavejs2mqtt/pull/100#issuecomment-749409611

aretakisv:
> > for the payload.name?
> 
> What?
> 
> I'm ok with this: [#100 (comment)](https://github.com/zwave-js/zwavejs2mqtt/pull/100#issuecomment-749409611)
This proposes to have two options under Gateway configuration:

1. Configurable Node Name:

  Location-Name (fallback to NodeId)
  Name only (fallback to NodeId)

2. Configure friendly_name:
  Default
  PropertyKey (if undefined PropertyName)
  NodeName with PropertyKey (else PropertyName)

robertsLando:
We could also think about using a placeholder method... Like:

Entity name: "%n %l %pk %pn" 

And replace: 

- `%n` with Node Name (fallback to node id)
- `%l` with node Location (empty if undefined)
- `%pk` propertyKey (propertyName if undefined)
- `%pn` propertyName 

@roflmao @larstobi @billiaz let me know your thoughts

aretakisv:
> We could also think about using a placeholders. In this case we will have just one configuration field:
> 
> Entity name: "%n %l %pk %pn"
> 
> And replace:
> 
> * `%n` with Node Name (fallback to node id)
> * `%l` with node Location (empty if undefined)
> * `%pk` propertyKey (propertyName if undefined)
> * `%pn` propertyName
> 
> @roflmao @larstobi @billiaz let me know your thoughts

like a templating system.
And how do you make nodeName configurable to include or not Location? with a switch?

robertsLando:
> And how do you make nodeName configurable to include or not Location? with a switch?

If you have placeholders you don't need it, give the user the possibility to choose how it's node name is composed using placeholders

aretakisv:
> > And how do you make nodeName configurable to include or not Location? with a switch?
> 
> If you have placeholders you don't need it, give the user the possibility to choose how it's node name is composed using placeholders. We should just decide what will be the default placeholder

that what I was thinking in the beginning! if you show me an example, I can change it to that :) 

As this is something like template! which is cool!

robertsLando:
@billiaz Essentially you should add an input text as configuration option, add a little legend below it that explains what are the available placeholders (check my previous comment and feel free to change them or add others). Then on the backend instead of calling your function that uses a switch-case to get the entity name just create a new one that does something like:

```
function gettEntityName(node, valueId) {

return this.config.entityTemplate.replace(/%n/g, node.name)
                                                   .replace(/%l/g, node.loc || '')
                                                   .replace(/%pk/g, valueId.propertyKey || valueId.propertyName)
                                                   .replace(/%pn/g, valueId.propertyName)

}

```

chilicheech:
I like this idea: https://github.com/zwave-js/zwavejs2mqtt/pull/100#issuecomment-749534757

lib/Gateway.js
```diff
@@ -621,6 +621,26 @@ function setDiscoveryValue (payload, prop, node) {
   }
 }
 
+/**
+ *
+ * @param {*} node node payload
+ * @param {*} valueId the valueId or CFG
+ * @param {*} nodeName compiled nodeName
+ * @param {*} valueId node value info
```
robertsLando, line 629: Add types

src/components/Settings.vue
```diff
@@ -320,6 +320,14 @@
                           persistent-hint
                         ></v-switch>
                       </v-flex>
+                      <v-flex xs12 v-if="gateway.hassDiscovery">
+                        <v-text-field
+                          v-model="gateway.entityTemplate"
+                          label="Entity name template"
+                          persistent-hint
+                          hint="Template which generates entity names<br>1"
+                        ></v-text-field>
```
robertsLando, line 329: Add a legend here, you can use a `<p>` tag

Also this should be set to default or the default should be specified somewhere

robertsLando (changes requested):

src/components/Settings.vue
billiaz: I have taken care the default value is set in function if this is undefined, i did not know how to set default here
still making function to cover both existense of valueId and when undefined. once I test all possible scenarios I will push my change and then will look into this!

lib/Gateway.js
billiaz: Will do, still working on function :) thnx

src/components/Settings.vue
```diff
@@ -321,12 +321,20 @@
...
-                          persistent-hint
-                          hint="Template which generates entity names<br>1"
-                        ></v-text-field>
+                          <v-text-field
+                            v-model="gateway.entityTemplate"
+                            label="Entity name template"
+                            persistent-hint
+                            hint="Template which generates entity names"
+                          ></v-text-field>
+                          <v-tooltip bottom>
```
robertsLando: @billiaz tooltip doesn't work in that way :) Check it here: https://vuetifyjs.com/en/components/tooltips/

robertsLando (changes requested):

robertsLando:
Schermata da 2020-12-23 11-01-33

lib/Gateway.js
```diff
@@ -650,10 +650,10 @@ function setDiscoveryValue (payload, prop, node) {
...
 /**
  *
- * @param {*} node any node payload
- * @param {*} valueId valueId object
- * @param {*} cfg configuration object
- * @param {*} entityTemplate gateway.entityTemplate string
+ * @param {Object} node any node payload
+ * @param {Object} valueId valueId object
+ * @param {Object} cfg configuration object
+ * @param {String} entityTemplate gateway.entityTemplate string
```
robertsLando, line 656:
```suggestion
 * @param {any} node node object
 * @param {any} valueId valueId object
 * @param {any} cfg configuration object
 * @param {string} entityTemplate the entity template from configuration
```

robertsLando (changes requested):

lib/Gateway.js
```diff
@@ -1157,7 +1183,12 @@ Gateway.prototype.discoverDevice = function (node, hassDevice) {
         hassDevice.object_id = sanitizeId(hassDevice.object_id)
 
         // Set a friendly name for this component
-        payload.name = `${nodeName}_${hassDevice.object_id}`
+        payload.name = getEntityName(
+          node,
+          undefined,
```
robertsLando, line 1188: I think that instead of undefined here you could use the zwaveNode deviceClass? Or something else, othetwise when using devices there could be repeated object_id in entity name, maybe this should be specified in the UI

```
{label: , propertyKey: , propertyName: }
```

robertsLando (changes requested):

lib/Gateway.js
billiaz: thank you @robertsLando for fixing this

lib/Gateway.js
billiaz: This part concerns only the customDevices.js(on)

I'm trying to find a case of duplicate. as custom devices are limiting you to specific values. I cannot think of any
in the other hand, you cannot have duplicates between different devices.

lib/Gateway.js
```diff
@@ -666,7 +666,8 @@ function getEntityName (node, valueId, cfg, entityTemplate) {
     label = valueId.label
   }
   return entityTemplate
-    .replace(/%n/g, node.name)
+    .replace(/%ann/g, nodeName)
```
robertsLando, line 669: I prefer the old way, this introduce just confusiuon about the difference between the auto generated and the default node name, it's clear to us but not to the end user

robertsLando (changes requested):

src/components/Settings.vue
```diff
@@ -320,6 +320,26 @@
...
+                      </v-flex>
+                      <v-flex xs6 v-if="gateway.hassDiscovery">
+                        <div>
+                          Default: <code>%loc-%n_%ob</code> <br />-
+                          <code>%n</code>: Node Name <br />- <code>%loc</code>:
+                          Node Location <br />- <code>%pk</code>: valueId
+                          property key (uses property name if undefined) <br />-
+                          <code>%pn</code>: valueId property name <br />-
+                          <code>%ob</code>: HASS object_id <br />-
+                          <code>%lbl</code>: valueId label
```
robertsLando, line 340: I would remove label from available placeholders, it could be very long in some values

src/components/Settings.vue
```diff
@@ -320,6 +320,26 @@
...
+                          v-model="gateway.entityTemplate"
+                          label="Entity name template"
+                          persistent-hint
+                          hint="Template which generates entity names"
+                        ></v-text-field>
+                      </v-flex>
+                      <v-flex xs6 v-if="gateway.hassDiscovery">
+                        <div>
+                          Default: <code>%loc-%n_%ob</code> <br />-
+                          <code>%n</code>: Node Name <br />- <code>%loc</code>:
```
robertsLando, line 335: Maybe we could also add the node id as placeholder `%nid`

src/components/Settings.vue
```diff
@@ -320,6 +320,26 @@
...
+                          Node Location <br />- <code>%pk</code>: valueId
+                          property key (uses property name if undefined) <br />-
+                          <code>%pn</code>: valueId property name <br />-
+                          <code>%ob</code>: HASS object_id <br />-
```
robertsLando, line 339: I would use `%o`

robertsLando (changes requested):

src/components/Settings.vue
billiaz: i agree! good one! will do

src/components/Settings.vue
billiaz: I do not agree, Label is really useful as it is better than other names for presenting short names, like
`Motion Detection Status`
`Battery Level`

an example of the longest i found! but makes sense
```json
{
  "id": "86-112-0-33",
  "nodeId": 86,
  "commandClass": 112,
  "commandClassName": "Configuration",
  "property": 33,
  "propertyName": "Sending Basic Set to nodes in Group 3 .",
  "type": "number",
  "readable": true,
  "writeable": true,
  "description": "Sending Basic Set to nodes in Group 3 .",
  "label": "Sending Basic Set to nodes in Group 3 .",
  "default": 3,
  "min": 0,
  "max": 4,
  "list": true,
  "states": [
    {
      "text": "Don’t send Basic Set.",
      "value": 0
    },
    {
      "text": "Option 1",
      "value": 1
    },
    {
      "text": "Option 2",
      "value": 2
    },
    {
      "text": "Option 3 (default)",
      "value": 3
    },
    {
      "text": "Option 4",
      "value": 4
    }
  ],
  "value": 3,
  "lastUpdate": 1608813167188,
  "nodeName": "Doorbell",
  "nodeLocation": "MetersBox"
}
```

Where typically is: 
```json

{
  "id": "87-128-0-level",
  "nodeId": 87,
  "commandClass": 128,
  "commandClassName": "Battery",
  "endpoint": 0,
  "property": "level",
  "propertyName": "level",
  "type": "number",
  "readable": true,
  "writeable": false,
  "label": "Battery level",
  "min": 0,
  "max": 100,
  "list": false,
  "value": 100,
  "lastUpdate": 1609086123492,
  "nodeName": "Button",
  "nodeLocation": "Kitchen"
}
```
in the second example, label makes more sense than level, except if we allow also Command Clases to join the list of placholders.
another example label fits the purpose:
```json
{
  "id": "97-48-0-Any",
  "nodeId": 97,
  "commandClass": 48,
  "commandClassName": "Binary Sensor",
  "endpoint": 0,
  "property": "Any",
  "propertyName": "Any",
  "type": "boolean",
  "readable": true,
  "writeable": false,
  "label": "Any",
  "ccSpecific": {
    "sensorType": 255
  },
  "list": false,
  "value": false,
  "lastUpdate": 1609088758510,
  "nodeName": "Sensor",
  "nodeLocation": "Hallway"
}
```

src/components/Settings.vue
billiaz: the purpose of using at least 2 letters is to avoid breaking changes in the future. if we introduce another placeholder starting with o. we will have to think priorities of replacement etc.

lib/Gateway.js
billiaz: this is removed

src/components/Settings.vue
robertsLando: Ok keep the label so :+1:

src/components/Settings.vue
robertsLando: I agree! Good point :)

src/components/Settings.vue
billiaz: changing also the node placeholder to %nn! as this fits on the scenario of conflict

src/components/Settings.vue
robertsLando: @billiaz We could make this more strict by adding \b at the end of the regex so this will not be a problem

billiaz: makes sense, I do not know if \b reads _ and - .. 
I checked with a regex and seems to work I will update patterns 


For instance from `%nn` to `%\bnn\b`

robertsLando: humm ok keep this for now so, no problem. Additionally we could firstly replace the longer occurance and then the smaller ones. For example if we would have `%n` and `%nn` we can firstly replace `%nn` and then `%n`

billiaz: I will revert the %nn to %n, I will add a comment to be careful with conflicting changes and order of replacement

this way I can revert the label to %l too

billiaz: @robertsLando check the new shortened ones. I could replace the l and make it location which makes it look: `%l-%n_%o` and keep label as `lbl` instead of `l`

lib/Gateway.js
robertsLando: What about simply don't print values if them are missing? I mean, if valueId is missing we should not fallback to object_id IMO

robertsLando (changes requested):

robertsLando (changes requested):
@billiaz What is missing now is just to add the entity name description in readme docs :)

robertsLando (approved):
LGTM

PR #105, opened 2020-12-21
[feat] Hass: Add Translation of states list to value_template map and Notification Command Class first to use this

billiaz:
Make Sensors of Notification CC provide translated values to Home assistant.

An example is Motion sensors. While the state is 0 or 8, this is not a nice to look into.
The real values per z-wave specification are idle and 'Motion detection'.

This patch converts the sensor values showin in HASS in human readable form.

Screenshot 2020-12-22 at 09 01 57

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -1520,7 +1520,7 @@ Gateway.prototype.discoverValue = function (node, vId) {
         }
         switch (valueId.propertyKey) {
           case 'Motion sensor status':
-            cfg.discovery_payload.icon = 'hass:walk'
+            cfg.discovery_payload.icon = 'mdi:motion-sensor'
             break
           default:
             cfg.discovery_payload.icon = 'mdi:alarm-light'
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -631,7 +631,7 @@ function getMappedStateTemplate (state, defaultValueKey) {
 
   return `{{ {${map.join(
     ','
-  )}}[value_json.value] | default(${defaultValue}}) }}`
+  )}}[value_json.value] | default(${defaultValue}) }}`
 }
 
 /**
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -1518,7 +1518,13 @@ Gateway.prototype.discoverValue = function (node, vId) {
         } else {
           cfg.object_id = 'notification_' + valueId.property
         }
-        cfg.discovery_payload.icon = 'mdi:alarm-light'
+        switch (valueId.propertyKey) {
+          case 'Motion sensor status':
+            cfg.discovery_payload.icon = 'hass:walk'
+            break
+          default:
+            cfg.discovery_payload.icon = 'mdi:alarm-light'
+        }
         if (valueId.list) {
           cfg.discovery_payload.value_template = getMappedStateTemplate(
             valueId.states,
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -615,8 +615,7 @@ function getMappedValuesTemplate (modeMap, defaultValue) {
  */
 function getMappedStateTemplate (state, defaultValueKey) {
   const map = []
-  let defaultValue = 'clear'
-
+  let defaultValue = 'value_json.value'
   for (const listKey in state) {
     map.push(
       `${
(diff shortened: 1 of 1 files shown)


lib/Gateway.js
```diff
@@ -1488,7 +1518,19 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
+          case 'Motion sensor status':
+            cfg.discovery_payload.icon = 'mdi:motion-sensor'
+            break
+          default:
+            cfg.discovery_payload.icon = 'mdi:alarm-light'
+        }
+        if (valueId.list) {
+          cfg.discovery_payload.value_template = getMappedStateTemplate(
+            valueId.states,
+            0
```
robertsLando, line 1531: could this be `valueId.default` instead?

robertsLando (changes requested):

lib/Gateway.js
aretakisv: I did not see this! and i was thinking about it! cool! let me check the payload and fix this! :)

lib/Gateway.js
```diff
@@ -1488,7 +1518,19 @@ Gateway.prototype.discoverValue = function (node, vId) {
         } else {
           cfg.object_id = 'notification_' + valueId.property
         }
-        cfg.discovery_payload.icon = 'mdi:alarm-light'
+        switch (valueId.propertyKey) {
```
robertsLando, line 1521: Add a TODO here to support other notification types, we could do this in another PR if you want (create a feature issue to keep track of this)

lib/Gateway.js
aretakisv:
```ccSpecific:{notificationType: 7}
commandClass:113
commandClassName:'Notification'
default:undefined
description:undefined
endpoint:0
id:'95-113-0-Home Security-Motion sensor status'
label:'Motion sensor status'
lastUpdate:1608645438998
list:true
max:255
min:0
nodeId:95
property:'Home Security'
propertyKey:'Motion sensor status'
propertyName:'Home Security'
readable:true
states:(2) [{…}, {…}]
0:{text: 'idle', value: 0}
1:{text: 'Motion detection', value: 8}
length:2
__proto__:Array(0)
step:undefined
type:'number'
unit:undefined
value:0
writeable:false
__proto__:Object
```
default is undefined. I think i need to make something like
`valueId.default ? valueId.default : 0`

robertsLando: @billiaz I think both are wrong, IMO default should not be a valid state, I mean if the value is not in the list of states with a default we will never know this is happening, we should show something like `Unknown` when the state is unknown

aretakisv: in that case home assistant will return the number! and user will see it as a broken setup which is acceptable.
when i post the template with out the `8:` i get 8 instead of motion detection.

The default value on states is always (per documentation) 0 (excel)

lib/Gateway.js
aretakisv: done this!

lib/Gateway.js
robertsLando: Ok so I think that is good, not to use a default one that exists

lib/Gateway.js
```diff
@@ -1528,7 +1530,7 @@ Gateway.prototype.discoverValue = function (node, vId) {
         if (valueId.list) {
           cfg.discovery_payload.value_template = getMappedStateTemplate(
             valueId.states,
-            0
+            valueId.default ? valueId.default : 0
```
robertsLando, line 1533: Just to be sure I understood this correctly, IMO when the value is not in the map the value shown should be the one coming from json payload. Id this what's happening here?

billiaz: I stand corrected. I did some nice patching on the payload, after manual testing

If `valueId.default` is undefinied the default value of payload is the `value_json.value`
this way it really presents the value numeric or other on UI

lib/Gateway.js
```diff
@@ -605,6 +605,33 @@ function getMappedValuesTemplate (modeMap, defaultValue) {
...
+          ? state[listKey].value
+          : '"' + state[listKey].value + '"'
+      }: "${state[listKey].text}"`
+    )
+    if (state[listKey].value === defaultValueKey) {
+      defaultValue = `'${state[listKey].text}'`
+    }
+  }
+
+  return `{{ {${map.join(',')}}[value_json.value] | default(${defaultValue}) }}`
```
robertsLando, line 632: Shouldn't this be without "default()"?

billiaz: No. if this is not having a default(), it returns no value.. aka empty at all, where in this case if there is no defined default in payload... unmapped value will return the number or string.

Basically the default here is if the map translation fails, what to show

robertsLando (approved):
LGTM

PR #111, opened 2020-12-25
[WIP] Correct the way value addition works

### LINKED ISSUE #110: [bug] When refreshing a node, values are empty
After a Refresh Info command, values are empty

trying to fix this as also
1. UI has empty values
2. Log shows no values in console

billiaz:
Trying to fix missing value when valueAdded

- when newValue is present and value is not, update node value during addValue execution
fixes #110

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -375,7 +375,6 @@ function onNodeDead (zwaveNode, oldStatus) {
 function onNodeValueAdded (zwaveNode, args) {
   logger.info(
     `Node ${zwaveNode.id}: adding value ${getValueID(args)} => ${args.newValue}`
-
   )
   // handle node values added 'on fly'
   if (zwaveNode.ready || args.newValue) {
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -380,6 +380,9 @@ function onNodeValueAdded (zwaveNode, args) {
 
   // handle node values added 'on fly'
   if (zwaveNode.ready || args.newValue) {
+    if (args.newValue) {
+      args.value = args.newValue
+    }
     addValue.call(this, zwaveNode, args)
   }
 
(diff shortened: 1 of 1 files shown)


lib/ZwaveClient.js
```diff
@@ -374,12 +374,13 @@ function onNodeDead (zwaveNode, oldStatus) {
...
 function onNodeValueAdded (zwaveNode, args) {
   logger.info(
-    `Node ${zwaveNode.id}: value added: ${getValueID(args)} =>`,
-    args.newValue
+    `Node ${zwaveNode.id}: adding value ${getValueID(args)} => ${args.newValue}`
   )
-
   // handle node values added 'on fly'
-  if (zwaveNode.ready) {
+  if (zwaveNode.ready || args.newValue) {
```
robertsLando, line 380: This could be break when `newValue` is `0`.

robertsLando (changes requested):

lib/ZwaveClient.js
```diff
@@ -374,12 +374,13 @@ function onNodeDead (zwaveNode, oldStatus) {
...
+    if (args.newValue) {
+      args.value = args.newValue
```
robertsLando, line 382: Also I don't understand why and how this could fix the problem, I use the zwave-js api getValue to get the actual value later

robertsLando (changes requested):

lib/ZwaveClient.js
billiaz: when you refresh (refresh info) a node, this never updates the values of the node, and reverts them to undefined. 
Because node is not ready, this never been executed. Refresh Info send newValue instead of value.
So newValue is parsed later? (just did not find it)

lib/ZwaveClient.js
billiaz: if i change it to `args.newvalue !== 'undefined'` would be better?

lib/ZwaveClient.js
robertsLando: > So newValue is parsed later? (just did not find it)

I never use that, I set the value in the `parseValue` function here: https://github.com/zwave-js/zwavejs2mqtt/blob/master/lib/ZwaveClient.js#L712

robertsLando:
Fixed here: https://github.com/zwave-js/zwavejs2mqtt/pull/113

Issue #109, opened 2020-12-24
[bug] Wrong topics for autodiscovered devices of Greenwave PowerNode 6-plug


katiuskt:
**Version**

Build/Run method
- [X] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

zwavejs2mqtt version: I'm not sure how to get this. I'm using zwavejs/zwavejs2mqtt:dev pulled on 24th Dec

**Describe the bug**
Autodiscovered devices for Greeenwave PowerNode 6-plugs are using wrong topics
Manufacturer ID: 153
Product Type: 3
Product ID: 4
![image](https://user-images.githubusercontent.com/75622100/103099527-1a063400-460f-11eb-9d86-615b861b3d1d.png)
![image](https://user-images.githubusercontent.com/75622100/103099734-ebd52400-460f-11eb-9e13-17ede8d12a0a.png)

* kwh_meter sensors unit_of_measurement should be kWh, not seconds. The point out to [node]/X/deltaTime/65537 instead to [node]/X/value/65537 (with X=1 to 6) 
* w_meter sensors unit_of_measurement should be W, not seconds. They point out to [node]/X/deltaTime/66049 instead to [node]/X/value/66049  (with X=1 to 6) 

![image](https://user-images.githubusercontent.com/75622100/103099881-7a49a580-4610-11eb-8e7b-cc29e45e418c.png)

I'm pretty confident in earlier versions of ZWaveJS2MQTT the autodiscovered devices were right (kwh sensors pointed out towards the right kwh topics, the same with the "w". In fact, even "previous value" and "delta time" devices were autodiscovered (and they don't now) so I'm not sure if this could be some kind of regression.

I've tried "rediscovering" the node and also "Refreshing Node Info" with no luck. For now, you need to edit the HASS device JSON and set the proper proper topics manually.

robertsLando:
@billiaz I think this has been broken by one of the last changes, could you investigate?

robertsLando:
Or maybe @AlCalzone Did you changed something on your side that could have messed up units?

billiaz:
This behaviour was observed also on 5.5.0 driver.

But I ignored it for a bit, as we where still focusing on other bits

AlCalzone:
> Or maybe @AlCalzone Did you changed something on your side that could have messed up units?

Nothing that comes to mind. I'll need to see the usual files to investigate (log and cache)

[...]

Issue #108, opened 2020-12-24
[bug] Missing sensors for three (of four) of the Octan NodOn Remote Controller scenes


katiuskt:
**Version**

Build/Run method
- [X] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

zwavejs2mqtt version: I'm not sure how to get this. I'm using zwavejs/zwavejs2mqtt:dev pulled on 24th Dec

**Describe the bug**
Missing HASS devices for NodOn Octan Remote Controller CRC-3100.
Manufacturer ID: 357
Product Type: 2
Product ID: 1

Once you hace included your Noctan Octan remote controller into the ZWave network, with this configuration (Central Scene):
![image](https://user-images.githubusercontent.com/75622100/103091423-6c862700-45f4-11eb-8a2c-dca12291fd66.png)

Each botton is identified like this into ZWaveJS2MQTT GUI:
![image](https://user-images.githubusercontent.com/75622100/103091459-8b84b900-45f4-11eb-87df-d6e1ede3def9.png)

Meaning, each one is identified by [node]-91-0-scene-00X with X=1, 2, 3, 4

However, not all of them are "converted" to devices for HASS. 

"sensor_scene_state" is for topic "91/0/slowRefresh"
![image](https://user-images.githubusercontent.com/75622100/103091584-e28a8e00-45f4-11eb-9e76-9adc80cfc75e.png)

And the relevant one "sensor_scene_state_0" is created for topic "91/0/scene/002":
![image](https://user-images.githubusercontent.com/75622100/103091629-06e66a80-45f5-11eb-9aa8-fe0f88833a13.png)

There are no devices for scenes 001, 003 or 004 hence HASS isn't aware when you click (or double click or hold,...) the other three buttons.

I've defined these devices manually and it did the trick. 
Note: The "sensor_scene_state_02" is basically the same than the automatically discovered "sensor_scene_state_0" but just assigning a name with the same pattern than the 01/03/04 to make it a little bit more consistent:

```
{
  "type": "sensor",
  "object_id": "scene_state_01",
  "discovery_payload": {
    "state_topic": "zwavejs/LR/RemoteNodOn/91/0/scene/001",
    "value_template": "{{ value_json.value}}",
    "json_attributes_topic": "zwavejs/LR/RemoteNodOn/91/0/scene/001",
    "device": {
      "identifiers": [
        "zwavejs2mqtt_0xf77d3d75_node9"
[...]

PR #112, opened 2020-12-26
[fix] Try to fix bugs 109 and 108

billiaz:
Add support for propertyKey and property on bugs 109 and 108.

I am looking into a more generic place for this to be applied. As it seems the same issue we faced with notifications exists in multiple places.

in some cases it makes sense to have a reverse order of the propertyKey and property, in order to look better and make sense.

Bugs: #108 #109 

example meter

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -1600,8 +1600,7 @@ Gateway.prototype.discoverValue = function (node, vId) {
         cfg.object_id =
           sensor.sensor + (sensor.objectId ? '_' + sensor.objectId : '')
         if (valueId.propertyKey) {
-          cfg.object_id =
-            cfg.object_id +
+          cfg.object_id +=
             (valueId.propertyKey ? '_' + valueId.propertyKey : '') +
             (valueId.property ? '_' + valueId.property : '')
         }
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -1615,6 +1615,7 @@ Gateway.prototype.discoverValue = function (node, vId) {
             sensor.sensor,
             sensor.objectId
           ]
+            .filter(Boolean)
             .join('_')
         }
 
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -1463,11 +1463,7 @@ Gateway.prototype.discoverValue = function (node, vId) {
       case CommandClasses['Central Scene']:
       case CommandClasses['Scene Activation']:
         cfg = copy(hassCfg.central_scene)
-        cfg.object_id = [
-          cfg.object_id,
-          valueId.propertyKey,
-          valueId.property
-        ]
+        cfg.object_id = [cfg.object_id, valueId.propertyKey, valueId.property]
           .filter(Boolean)
           .join('_')
         break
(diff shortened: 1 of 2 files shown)


lib/Gateway.js
```diff
@@ -1463,6 +1463,9 @@ Gateway.prototype.discoverValue = function (node, vId) {
       case CommandClasses['Central Scene']:
       case CommandClasses['Scene Activation']:
         cfg = copy(hassCfg.central_scene)
+        cfg.object_id = [cfg.object_id, valueId.propertyKey, valueId.property]
+          .filter(Boolean)
```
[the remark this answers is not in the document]
robertsLando: What would you like to do with that?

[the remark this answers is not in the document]
aretakisv: this is taking an Array join it and use _ by joining it.
.filter(Boolean) removes the undefined params.
This object_id is used for:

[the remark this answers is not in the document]
robertsLando: > removes the undefined params.

I prefer `.filter(v => !!v)` so

lib/Gateway.js
```diff
@@ -1592,8 +1595,22 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
+        // Assemble an object ID including propertyKey and property.
+        // PropertyKey is set before property as propertyKey can have multiple relevant properties.
+        if (valueId.propertyKey) {
+          cfg.object_id = [
+            sensor.sensor,
+            sensor.objectId,
+            valueId.propertyKey,
+            valueId.property
+          ]
+            .filter(Boolean)
```
robertsLando, line 1607: Instead of doing this twice, do it only after the if-else

robertsLando (changes requested):

lib/Gateway.js
billiaz: to reduce the confusion i think this is a better approach:
```
cfg.object_id =
  sensor.sensor + (sensor.objectId ? '_' + sensor.objectId : '')
if (valueId.propertyKey) {
  cfg.object_id = 
    cfg.object_id +
    (valueId.propertyKey ? '_' + valueId.propertyKey : '') +
    (valueId.property ? '_' + valueId.property : '')
}
```

robertsLando: Ok :+1:

aretakisv: pushed fixes for this and the filter proposal on previous

lib/Gateway.js
```diff
@@ -1597,19 +1597,13 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
-            valueId.property
-          ]
-            .filter(Boolean)
-            .join('_')
-        } else {
-          cfg.object_id = [sensor.sensor, sensor.objectId]
-            .filter(Boolean)
-            .join('_')
+          cfg.object_id =
+            cfg.object_id +
```
robertsLando, line 1604:
```suggestion
          cfg.object_id +=
```

robertsLando (changes requested):

lib/Gateway.js
billiaz: did not know this exists! cool

robertsLando (approved):
LGTM

PR #126, opened 2020-12-30
Add support for Battery isLow binary sensor

billiaz:
Low battery level is a binary sensor, but it is discovered through multilevel/sensors

Added some small changes, which can support cases like this in the future


robertsLando (approved):
LGTM

PR #151, opened 2021-01-05
fix: use buttons for target value #133

### LINKED ISSUE #133: [feat] Target Value should be selectable by user
Currently the targetValue is a toggle switch. This toggle is (by design?) out of sync with the current state of the switch. So what happens is that the "current value" will show true, but the target value will show false. Now to turn off the switch I have to issue two commands, once to toggle the target value on then, then another to toggle the target value off.

TargetValue

I would like to have the option to select which target value I want to send. 

You could ignore this request :)

**Additional context**

robertsLando:
Fixes #133 

Now switches are replaced by two buttons:

Schermata da 2021-01-05 09-59-11


aretakisv:
This is nice! 
What I wanted to also mention, although not covered here!

Level up and Level down buttons should be instant buttons and not on/off as it is press/release action
We should think it separately. but this solution is nice for true/false events

to fit with your ui colors, would different softer colors be an option?

robertsLando:
> to fit with your ui colors, would different softer colors be an option?

Suggestion?  As @AlCalzone said bool values are not stateless (at least not all) and to show the actual state I use the `blue` color.

If the actual state is `true` the `ON` will be `blue` and the `OFF` red.  
If the actual state is `false` the `ON` will be `green` and the `OFF` blue.

robertsLando:
I have fixed the style now. I prefer this: 

Schermata da 2021-01-05 12-11-28

PR #154, opened 2021-01-05
fix: run new build:full command to build zwave-js

AlCalzone:
Yesterday I did some changes to speed up CI runs in zwave-js. One big offender was building the TS sources which essentially built the same packages over and over.
As a result I split the build step into two - one for packaging etc. (`build:full`) and one for small updates (`build`).


robertsLando (approved):
LGTM

PR #208, opened 2021-01-11
fix: HASS Discovery - Name - First character fix

### LINKED ISSUE #162: [bug] hass entity names are now being prefixed with `-`
**Version**

Build/Run method
- [x] Docker
- [ ] PKG

zwavejs2mqtt version: 1.0.0.alpha.2

**Describe the bug**
Before this version HASS entity names were being reported correctly, for instance, `patio_lights_dimmer`. In this version the name is being prefixed with a dash, `-patio_lights_dimmer`. This is probably due to the introduction of the entity name template feature, that has a default template of `%loc-%n_%o`. However, this "breaks" the name when there's no location present. When the location is empty/null, the dash should be removed.

**Expected behavior**
Using the default entity name template, when the node's location is NOT empty/null, the effective template should be what it is now: `%loc-%n_%o`. However, when node's location IS empty/null, the effective template should be `%n_%o` (without a dash for a prefix).

**Additional context**
Here's a picture of my config:
Screen Shot 2021-01-06 at 11 49 31 AM
Here's a picture of the config of the patio_lights device:
Screen Shot 2021-01-06 at 11 59 05 AM

And here's the generated HASS discovery:
```
{
  "type": "light",
  "object_id": "dimmer",
  "discovery_payload": {
    "schema": "template",
    "brightness_template": "{{ (value_json.value / 99 * 255) | round(0) }}",
    "state_topic": "z2m/2/38/0/currentValue",
    "state_template": "{{ \"off\" if value_json.value == 0 else \"on\" }}",
    "command_topic": "z2m/2/38/0/targetValue/set",
    "command_on_template": "{{ ((brightness / 255 * 99) | round(0)) if brightness is defined else 255 }}",
    "command_off_template": "0",
    "device": {
      "identifiers": [
        "zwavejs2mqtt_0xc5b5c9e5_node2"
      ],
      "manufacturer": "Zooz",
      "model": "Dimmer Paddle Switch (ZEN22)",
      "name": "patio_lights",
      "sw_version": "3.5"
    },
    "name": "-patio_lights_dimmer",
    "unique_id": "zwavejs2mqtt_0xc5b5c9e5_2-38-0-currentValue"
  },
  "discoveryTopic": "light/patio_lights/dimmer/config",
  "values": [
    "38-0-currentValue",
    "38-0-targetValue"
  ],
  "persistent": false,
  "ignoreDiscovery": false,
  "id": "light_dimmer"
}
```
Notice how it now sets `"name": "-patio_lights_dimmer"` whereas before the entity name template feature was introduced it would correctly set it to `"name": "patio_lights_dimmer"`.

billiaz:
Apply a fix for better Naming in Home Assistant. Add a replace of the first character if it is a hyphen `-`, understore `_` or space ` `
by cleaning up the first character, in case of location is empty or unused, like in default payload. Will then clean up the this characted.

fixes #162

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -688,6 +688,7 @@ function getEntityName (node, valueId, cfg, entityTemplate) {
     .replace(/%o/g, cfg.object_id)
     .replace(/%n/g, node.name || NODE_PREFIX + node.id)
     .replace(/%l/g, label)
+    .replace(/^[-_ ]/, '')
 }
 
 /**


robertsLando:
@billiaz This fixes the problem but IMO it's not the best way. You are replacing `_-` chars but them could be also `.,;` or any other as templates have not a strict form, it works for @chilicheech but not in all others templates.

@chilicheech If you are not using location why don't just remove it from the template?

robertsLando:
@billiaz Ok I have the solution to this. Add a new placeholder `%nn` that stay for `nodeName` and use the function `getNodeName()` to retrive it. That was the function used before and essentially will return `location-name` or `name` if location is not defined. Also the default template then will become `%nn_%o`

robertsLando:
Supersided by #218

PR #209, opened 2021-01-11
Fix commands in Dockerfile.contrib steps

LordMike:
First fix: If you cd into "docker", the Dockerfile is present in the same directory. 

Second fix: The branch `feat#startStop` doesn't exist, and most people must want to build master branches.


robertsLando (approved):
LGTM

Issue #210, opened 2021-01-11
[bug] Upgrade to 6.0.0-alpha.3 broke HASS scene sensor states


joeyb:
Build/Run method
- [x] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

zwavejs2mqtt version: Working pre-`a75c112ce6bb52d85f75829b36e5f8fce99899a2`, broken since.

**Describe the bug**
After the zwave-js upgrade in `a75c112ce6bb52d85f75829b36e5f8fce99899a2`, the MQTT scene topics no longer contain the actual scene state, just the `time`, `nodeName`, and `nodeLocation`. For example, I have some HomeSeer HS-WD100+ switches that should set the 001 scene's state to `3` when the top paddle is double tapped. Prior to `a75c112` that worked, but after the upgrade the entire state value is missing from the payload.

**To Reproduce**
1. Run a version of zwavejs2mqtt post-`a75c112ce6bb52d85f75829b36e5f8fce99899a2`.
2. In MQTT Explorer (or some other MQTT browser), pull up the `central_scene/endpoint_0/scene/001` topic for your device.
3. Trigger that scene (for example, with a double tap on the HS-WD100+ switches).
4. In MQTT Explorer, you should see that the payload in that topic is missing the state value.

**Expected behavior**
The state value should be set in the scene's topic.

blhoward2:
Thanks for reporting this. We’ve reproduced it on another switch.

jcam:
The cause of this is covered here:
https://zwave-js.github.io/node-zwave-js/#/getting-started/migrating-to-v6?id=new-quotvalue-notificationquot-event

The new event handler that was added did not account for the value being sent in a different argument, and it also did not account for the shift of CCs like Central Scene from stateful to stateless.

I have implemented an interim solution here which sends the updated value in the 'state' field to MQTT so it is received by the sensor, which should match the old functionality.

IMO The correct fix is to stop using the state_changed event entirely and shift to a stateless event.

As implemented here, an automation that triggers on "state" will only be triggered the first time a Central Scene is pressed, i.e. if you tap the top of a Zooz ZEN26 scene switch twice, the scene002 sensor's state will be updated to 3. If you press the top of the switch twice again, the state is already 3, so it is not a state change.
Instead you have to use an event trigger.
Fix is in the below PR

robertsLando:
[...]

PR #212, opened 2021-01-12
fix: central scene notification argument list changed in Zwave-JS 6.0

jcam:
Resolves https://github.com/zwave-js/zwavejs2mqtt/issues/210

The correct fix is to change to a stateless event notification, rather than tracking a stateful value, but this should work as a workaround.


robertsLando (approved):
LGTM. I still have to set retain flag to false when publishing notifications, that is what is missing right now

PR #215, opened 2021-01-12
fix: set retain to false for stateless valueIds

robertsLando:


robertsLando:
@billiaz Test this when you have time please :)

jcam:
@robertsLando is there a benefit to using a sensor and setting a timer to reset the valueID, rather than using a device trigger (https://www.home-assistant.io/integrations/device_trigger.mqtt/)?

robertsLando:
> @robertsLando is there a benefit to using a sensor and setting a timer to reset the valueID, rather than using a device trigger (https://www.home-assistant.io/integrations/device_trigger.mqtt/)?

@jcam I don't use hass but we could speak about this with @billiaz

billiaz:
the intent from @robertsLando was to make it work as OZW did! which is a good first step. Till now we used Sensors for Scenes.

Next is to also add support for Triggers on. As explained in Slack. This is a bit more complicated! as HASS gives no templating flexibility for triggers. And more time needs to be invested on this.

billiaz (approved):

PR #225, opened 2021-01-12
fix: error in filter options for null strings

ahochsteger:
If node values are `null` or `undefined` errors like the following show up in the browser console and the selection of string values for columns in the node table is not available:
```
vue.esm.js?a026:628 [Vue warn]: Error in render: "TypeError: Cannot read property 'header' of null"
found in
---> <VSelect>
       <VForm>
         <ColumnFilterString> at src/components/nodes-table/ColumnFilterString.vue
```
This PR fixes this problem by only considering non-null/non-undefined values for <v-select>.


robertsLando (approved):
LGTM! Thanks :)

PR #235, opened 2021-01-14
fix: better hass id and topics sanification

robertsLando:
Sanitize properties like `Door/Window` in topics and hass ids


robertsLando:
@billiaz Coudl you give this a try?

robertsLando:
merged in #247

Issue #232, opened 2021-01-13
[question] Aeon Labs Home Energy Meter G2 to much information


haavardNO:
One of the devices I have added recently is a Home energy meter.
I have used this device in my previous HA application (Indigo for mac) and zwave2mqtt. On both platforms the sensors made sense. But when added to zwavejs2mqtt there is a lot of sensors that I have never seen before. Like previous values and delta time etc... I sort of just want the energy readings. When created in Home Assistant it creates so long names that it is difficult to make out what is what.

Is it supose to be like this, or is there something faulty with the way zwavejs2mqtt is reading it?
<img width="2672" alt="Skjermbilde 2021-01-13 kl  18 11 53" src="https://user-images.githubusercontent.com/27728530/104486078-8e5b4400-55cb-11eb-9864-47293ec52f08.png">

robertsLando:
@haavardNO Could you show me the names you are getting? Have you tried to set a custom entity name? @billiaz

Swiftnesses:
I get this with my Aeon plug, the previous value is required to store the value when you reset values (the button in your screenshot), the delta entities can just been hidden / disabled (either in the zwavejs side, or the Home Assistant side).

haavardNO:
@robertsLando no I have set a name and location in the zwavejs2mqtt UI, and I have changed the name in HA afterwards. But it is pain to find the device that gives the correct value since there is so many. This metering device has one clamp. This generates 21 devices. The other metering device I use has 3 clamps, that device generates 36 devices. I guess I can tweak it to figure out what is what. 
<img width="1093" alt="Skjermbilde 2021-01-13 kl  20 35 22" src="https://user-images.githubusercontent.com/27728530/104500916-4fcf8480-55df-11eb-9cf5-f35084e12972.png">
<img width="1097" alt="Skjermbilde 2021-01-13 kl  20 40 12" src="https://user-images.githubusercontent.com/27728530/104501103-91602f80-55df-11eb-868e-b5ba685dea2a.png">
<img width="749" alt="Skjermbilde 2021-01-13 kl  20 41 03" src="https://user-images.githubusercontent.com/27728530/104501204-adfc6780-55df-11eb-8c1a-f817c60c31b5.png">

robertsLando:
I think I need to configure a way to skeep discovery for some values. @billiaz I think that in case of meter CC some values should be skipped, then we could also provide a way to ignore the discovery of some specific values using the gateway values table

[...]

PR #237, opened 2021-01-14
feat: reduce HASS device noise

billiaz:
add support for secondary/non essential devices

By default some devices will not be created if they are considered secondary.
- Add switch in settings
- Add list of devices (for Meters)
- Add documentation

improvement based on issue #232


robertsLando (changes requested):
@billiaz I would't discover them at all, I don't think we need a gateway option for that. What do you think? COuld them be needed?

aretakisv:
I like to have the option to enable these!
I like to compare trend between these and get an average over delta (for instance)
Disabled by default is good i think.

robertsLando:
Ok so I think we should adopt a different approach for this, by using gateway values table.

Issue #245, opened 2021-01-15
[feat] Expose refreshValues() through mqtt


blhoward2:
**Is your feature request related to a problem? Please describe.**
There should be a method to refresh only the dynamic values of a node, on demand. Basically manual polling. There is a function refreshValues() already used in places like HailCC, which should also be exposed through MQTT.

**Describe the solution you'd like**
expose refreshValues()

**Additional context**
Necessary for various workarounds where a node needs to be refreshed, like scene events with buggy devices, buggy switches, etc.

PR #246, opened 2021-01-15
feat: refreshValues api

blhoward2:
I'm attempting to expose the refreshValues() function to mqtt to permit a manual refresh of the dynamic values. To fix: #245 

I've already submitted a PR upstream making the function public. https://github.com/zwave-js/node-zwave-js/pull/1403


blhoward2:
@AlCalzone @robertsLando I'd appreciate some help here. I can't seem to make this work. I've made the function in Node.js public, and I've tried every possible way that I can think of to access that function but none work. this.nodes.refreshValues(), this.controller.node.refreshValues(), this.controller.nodes.refreshValues(), etc. What am I doing wrong? 

I get success:false when I call it in MQTT, with a message that function is either undefined or that the property cannot be read from undefined.

robertsLando:
@blhoward2 because `this.nodes` are not the driver nodes. In order to call that function you have to use `this.driver.controller.nodes`, also isn't that value node specific? If so even better to use `this.getNode(nodeId).refreshValues()`

AlCalzone:
We should make it clear that calling this often is not a good idea because it might cause lots of traffic.

Additionally, we should make it possible to refresh single values, so applications can work around device quirks, e.g. not updating some values correctly.

blhoward2:
> We should make it clear that calling this often is not a good idea because it might cause lots of traffic.
> 
> Additionally, we should make it possible to refresh single values, so applications can work around device quirks, e.g. not updating some values correctly.

I posted on the node pr too. I think having both is a good idea. I’m trying to duplicate ozw’s refreshDynamic service that updated all dynamic values, which I think this does. I agree it would be good to be able to get a single value too.

blhoward2:
> @blhoward2 because `this.nodes` are not the driver nodes. In order to call that function you have to use `this.driver.controller.nodes`, also isn't that value node specific? If so even better to use `this.getNode(nodeId).refreshValues()`

Ok, I tried this in the new commit but I get "Cannot read property 'refreshValues' of undefined" in mqtt. Thoughts?

robertsLando:
Are you passing the nodeid in mqtt payload args?

blhoward2:
> Are you passing the nodeid in mqtt payload args?

Hmm, I am not. I can do so, I suppose.

lib/ZwaveClient.js
```diff
@@ -1838,8 +1838,8 @@ ZwaveClient.prototype.getInfo = function () {
   return info
 }
 
-ZwaveClient.prototype.refreshValues = function () {
-  this.nodes.refreshValues()
+ZwaveClient.prototype.refreshValues = function (nodeId) {
+  this.getNode(nodeId).refreshValues()
```
AlCalzone, line 1842: That method returns a Promise which could (theoretically) be rejected - should probably add a .catch()

blhoward2:
> Are you passing the nodeid in mqtt payload args?

Same with the following:
```

{
  'nodeId': "11"
}
```

I tried it as both a string and an integer.

lib/ZwaveClient.js
blhoward2: Done in new commit

lib/ZwaveClient.js
```diff
@@ -1838,6 +1838,12 @@ ZwaveClient.prototype.getInfo = function () {
   return info
 }
 
+ZwaveClient.prototype.refreshValues = function (nodeId) {
+  this.getNode(nodeId).refreshValues()
```
robertsLando: Anyway I prefer something like: 

```js
ZwaveClient.prototype.refreshValues = async function (nodeId) {
  if (this.driver && !this.closed) {
    const zwaveNode = this.getNode(nodeId)

    return zwaveNode.refreshValues()
  }

  throw Error('Driver is closed')
}
```

robertsLando (changes requested):

lib/ZwaveClient.js
blhoward2: Should it be a throw or a catch per @AlCalzone comment?

robertsLando:
@blhoward2 

```
{
  "args": [nodeId]
}
```

Check https://zwave-js.github.io/zwavejs2mqtt/#/guide/mqtt?id=zwave-apis

lib/ZwaveClient.js
blhoward2: also should it still return true?

blhoward2:
> @blhoward2
> 
> ```
> {
>   "args": [nodeId]
> }
> ```
> 
> Check https://zwave-js.github.io/zwavejs2mqtt/#/guide/mqtt?id=zwave-apis

Is there noway to extra the nodeId from the mqtt command string? Like you do for getAssociations?

lib/ZwaveClient.js
robertsLando: This is how I have implemented all other methods if you check. It's correct to throw there

AlCalzone:
Upstream PR merged

robertsLando:
> Is there no way to capture the nodeId from the mqtt command string? Like you do for getAssociations? that would be easier to use since there are no other args

I don't understand what you mean here. What's the mqtt command string? All zwave apis and also others are called using the same way, I mean I simply apply the args coming from the mqtt to the function specified (if exists), making what you asked means create a big switch case where foreach api I fetch the required props from payload, this is just quicker

blhoward2:
> > Is there no way to capture the nodeId from the mqtt command string? Like you do for getAssociations? that would be easier to use since there are no other args
> 
> I don't understand what you mean here. What's the mqtt command string? All zwave apis and also others are called using the same way, I mean I simply apply the args coming from the mqtt to the function specified (if exists), making what you asked means create a big switch case where foreach api I fetch the required props from payload, this is just quicker

Maybe I've got this all wrong. If I publish homeassistant/_CLIENTS/ZWAVE_GATEWAY-11/api/getAssociations/set with no payload it returns Success:true. Is that not right? Where 11 is the node name?

robertsLando:
@blhoward2 YOu can even add it to UI, just add an entry here: 

https://github.com/blhoward2/zwavejs2mqtt/blob/add_refresh_values/src/components/ControlPanel.vue#L691

blhoward2:
I think it is working now with the committed changes. I'll verify in a bit once the network is full up. I get a success message back in mqtt, though.

blhoward2:
@zwave-js-bot fix lint

blhoward2:
@zwave-js-bot fix lint

blhoward2:
How come the bot isn't responding?

blhoward2:
Confirmed working on multiple devices.

docs/guide/mqtt.md
```diff
@@ -74,6 +74,7 @@ There are some custom apis that can be called that are not part of Zwave Client:
   - `lastUpdate`: Timestamp of latest event received from OZW
   - `status`: Client status. Could be: 'driverReady', 'connected', 'scanDone', 'driverFailed', 'closed'
   - `cntStatus`: Controller status received from ozw notifications controller command. If inclusion/exclusion is running it would be `Waiting`
+- `refreshValues`: Queries all actuator and sensor values of all node endpoints
```
robertsLando, line 77: This is not needed as it is a zwave api

robertsLando (changes requested):

docs/guide/mqtt.md
blhoward2: How do people know it exists, then? Someone on slack yesterday asked if it existed as they'd been looking through the docs.

robertsLando: because that api supports all zwave apis by default, in the list I only specify the ones that are not from zwave

robertsLando: We make a list of the supported ones too, If you want to

blhoward2: Might be a good idea. I think the mqtt documentation in general is a little hard to understand. May be worth having a second look at the whole thing with an eye towards an average user. I removed it from the PR for now.

blhoward2:
Whoops, that deletes the whole thing. How do I undo a commit?

docs/guide/mqtt.md
robertsLando: @blhoward2 I'm open to suggestion/help about how to improve it :)

robertsLando:
@blhoward2 `git revert 1e1ff05`

docs/guide/mqtt.md
blhoward2: mqtt is an odd beast. It is fairly advanced but every now and then a normal user needs to publish a message (and HA lets you do so within an automation as an action)

robertsLando (approved):
LGTM

robertsLando:
@blhoward2 Need a new beta or other release before merging this. @AlCalzone

blhoward2:
Understood

AlCalzone:
6.0.0 is going out tonight if there is no critical bug

PR #247, opened 2021-01-15
fix: topics and hass ids/names fixes

### LINKED ISSUE #239: [bug] propertyKeyName should be used instead of propertyKey if it exists
Found several locations where `propertyName` is used in place of `property`, but the same does not happen for `propertyKeyName`.

Also, this is completely wrong I think (`propertyName` instead of `propertyKeyName`):
https://github.com/zwave-js/zwavejs2mqtt/blob/6e2a0f726aa8f14bc06258860c002e6373697f05/lib/Gateway.js#L678

robertsLando:
Fixes #239 #232


robertsLando:
I need tests on this bdbogjoe

bdbogjoe:
still having 
`2021-01-19 21:27:37.163 INFO ZWAVE: Node 8: metadata updated: 50-1-value-65537`

robertsLando:
> still having
What's wrong with that? That will never change, it's the value id and it's unique, coming from zwave-js not decided by me

bdbogjoe:
ok i thought you wanted to rename this to have like /Power for some devices

lib/Gateway.js
```diff
@@ -1185,7 +1179,7 @@ Gateway.prototype.discoverDevice = function (node, hassDevice) {
         // Set device information using node info
         payload.device = deviceInfo.call(this, node, nodeName)
 
-        hassDevice.object_id = sanitizeId(hassDevice.object_id)
+        hassDevice.object_id = utils.sanitizeTopic(hassDevice.object_id, true)
```
robertsLando, line 1182:
```suggestion
        hassDevice.object_id = utils.sanitizeTopic(hassDevice.object_id, true).toLocaleLowerCase()
```

lib/Gateway.js
```diff
@@ -1740,15 +1727,16 @@ Gateway.prototype.discoverValue = function (node, vId) {
     payload.device = deviceInfo.call(this, node, nodeName)
 
     // multi instance devices would have same object_id
-    if (valueId.endpoint > 1) cfg.object_id += '_' + valueId.endpoint
+    if (valueId.endpoint) cfg.object_id += '_' + valueId.endpoint
+
+    // remove chars that are not allowed in object ids
+    cfg.object_id = utils.sanitizeTopic(cfg.object_id, true)
```
robertsLando, line 1733:
```suggestion
    cfg.object_id = utils.sanitizeTopic(cfg.object_id, true).toLocaleLowerCase()
```

PR #248, opened 2021-01-15
feat: zwave-js server support

robertsLando:


balloob (dismissed):

ahochsteger:
@robertsLando would this be the solution to restart the UI of zwavejs2mqtt without re-discovering all nodes?
I guess the zwave-js server would then run in a separate Docker container than zwavejs2mqtt.
If there's something I can do to make this happen (e.g. do some tests or help with Docker setup) just let me know.

balloob:
@ahochsteger that is not what this PR is about. This is about also exposing a websocket connection for other applications that want access to the Z-Wave JS driver.

ahochsteger:
@balloob ah, I see - thanks for clarifying!

PR #258, opened 2021-01-17
fix(ui): inconsistent, unsorted node selection

ahochsteger:
If you have a large number of nodes it is not easy to find the relevant ones in the selection list.
Therefore I suggest to sort them by display name (using `_name`).

Affected selections:
* groups: node and target selection
* scenes: node selection

In addition, the node selection in the scenes tab did not filter failed nodes, which has been changed for consistency too.

### DIFF
--- a/src/components/ControlPanel.vue
+++ b/src/components/ControlPanel.vue
@@ -296,7 +296,15 @@
                     <v-select
                       label="Node"
                       v-model="group.node"
-                      :items="nodes.filter(n => !n.failed)"
+                      :items="
+                        nodes
+                          .filter(n => !n.failed)
+                          .sort((n1, n2) =>
+                            n1._name.toLowerCase() < n2._name.toLowerCase()
+                              ? -1
+                              : 1
+                          )
+                      "
                       return-object
                       @change="resetGroup"
                       item-text="_name"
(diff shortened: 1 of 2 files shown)


src/components/ControlPanel.vue
```diff
@@ -296,7 +296,15 @@
                     <v-select
                       label="Node"
                       v-model="group.node"
-                      :items="nodes.filter(n => !n.failed)"
+                      :items="
```
robertsLando, line 299: Make this computed

src/components/ControlPanel.vue
```diff
@@ -355,7 +363,15 @@
                     <v-combobox
                       label="Target"
                       v-model="group.target"
-                      :items="nodes.filter(n => !n.failed && n != group.node)"
+                      :items="
```
robertsLando, line 366: dito

src/components/dialogs/DialogSceneValue.vue
```diff
@@ -18,7 +18,13 @@
                   item-text="_name"
                   :rules="[required]"
                   item-value="id"
-                  :items="nodes.filter(n => !!n)"
+                  :items="
```
robertsLando, line 21: dito

robertsLando (changes requested):

src/components/ControlPanel.vue
ahochsteger: Good point - done.

src/components/ControlPanel.vue
ahochsteger: done.

src/components/dialogs/DialogSceneValue.vue
ahochsteger: done.

robertsLando (approved):
LGTM

PR #261, opened 2021-01-18
fix: error building with Dockerfile.contrib

ahochsteger:
Fixes the following error when building a docker image using Dockerfile.contrib:

```bash
Step 10/39 : COPY --chown=node node-zwave-js /home/node/node-zwave-js
COPY failed: stat /var/lib/docker/tmp/docker-builder759222654/node-zwave-js: no such file or directory
```

### DIFF
--- a/docker/Dockerfile.contrib
+++ b/docker/Dockerfile.contrib
@@ -11,8 +11,8 @@ RUN git clone -b ${Z2M_BRANCH} --depth 1 https://github.com/zwave-js/zwavejs2mqt
 
 ##### LOCAL SOURCE #####
 FROM node:erbium-buster AS local-copy-src
-COPY --chown=node node-zwave-js /home/node/node-zwave-js
-COPY --chown=node zwavejs2mqtt /home/node/zwavejs2mqtt
+COPY --from=git-clone-src --chown=node /home/node/node-zwave-js /home/node/node-zwave-js
+COPY --from=git-clone-src --chown=node /home/node/zwavejs2mqtt /home/node/zwavejs2mqtt
 
 ##### BUILD #####
 FROM ${SRC} AS build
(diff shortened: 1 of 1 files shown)


robertsLando:
@larstobi Give this a check when you have time :)

robertsLando (approved):
LGTM

larstobi:
Trying

larstobi:
Please don't merge this yet. I have some comments.

docker/Dockerfile.contrib
```diff
@@ -11,8 +11,8 @@ RUN git clone -b ${Z2M_BRANCH} --depth 1 https://github.com/zwave-js/zwavejs2mqt
 
 ##### LOCAL SOURCE #####
 FROM node:erbium-buster AS local-copy-src
-COPY --chown=node node-zwave-js /home/node/node-zwave-js
-COPY --chown=node zwavejs2mqtt /home/node/zwavejs2mqtt
+COPY --from=git-clone-src --chown=node /home/node/node-zwave-js /home/node/node-zwave-js
```
larstobi, line 14: This will not work when using local-copy-src since it will take the cloned source

docker/Dockerfile.contrib
```diff
@@ -26,23 +26,23 @@ RUN yarn install --network-timeout=${YARN_NETWORK_TIMEOUT}
 RUN yarn run build:full
 RUN yarn install --production --frozen-lockfile
 RUN for i in config core serial shared; do \
-    cd packages/$i && \
-    yarn version --no-git-tag-version --new-version $(yarn versions --json| \
-        jq -r '[.data."@zwave-js/'${i}'"]'[0])-$(git rev-parse --short HEAD) && \
```
larstobi, line 31: The reason this is indented is that it's part of the previous command. I think that's a good thing.

ahochsteger: > The reason this is indented is that it's part of the previous command. I think that's a good thing.

I think, this was by mistake due to auto-save which I seem to have missed.

docker/Dockerfile.contrib
```diff
@@ -26,23 +26,23 @@ RUN yarn install --network-timeout=${YARN_NETWORK_TIMEOUT}
 RUN yarn run build:full
 RUN yarn install --production --frozen-lockfile
 RUN for i in config core serial shared; do \
-    cd packages/$i && \
```
larstobi, line 29: I would prefer not reducing indentation spaces from 4 to 2.  It is also consistent with the other Dockerfile in this directory.

ahochsteger: > I would prefer not reducing indentation spaces from 4 to 2. It is also consistent with the other Dockerfile in this directory.

You're right - this was slipped through as well due to auto-format on save.

larstobi:
@ahochsteger Thanks for taking the time to raise this PR! :-)

I've investigated your changes, and unfortunately I cannot say that I agree.

The reason you're getting that error is that you're not prefixing your docker command with `DOCKER_BUILDKIT=1`

The changes you've made breaks the Dockerfile when using `SRC=local-copy-src` so I think this should not be merged.

Now, I understand that you would like to use this dockerfile without enabling docker BuildKit, and I've investigated with some ways by making another build stage, called "start", but I still haven't been able to make docker skip unused stages without BuildKit.

If you can find a way that works for both SRC arguments (local and clone), without using BuildKit, then I'm all for it!

ahochsteger:
@larstobi thangs for taking the time to review this PR.
It seems, that I've not read the README which documents how to use it properly ;-).
I'm fine to cancel this PR and remove the branch, since it's working for me when using it the right way.

larstobi:
Great to hear that it works for you now! :-D

PR #262, opened 2021-01-18
chore(docker): docker labels opencontainers

robertsLando:
Fixes #257

Issue #257, opened 2021-01-16
[feat] Add labels to docker image linking to git hash


LordMike:
**Is your feature request related to a problem? Please describe.**
When downloading a docker image, it could be useful to know exactly which commit this image was from. This goes for both `zwave-js` and `zwavejs2mqtt`.

**Describe the solution you'd like**
Add a label on builds, with the git hash of the two projects.

I'm fairly certain the change has to happen [around here](https://github.com/zwave-js/zwavejs2mqtt/blob/8371752b9e9f20f596606aec6d9bfedf3d9b0b5d/.github/workflows/dockerbuild.yml#L44-L51), but I'm not sure what to do.

**Describe alternatives you've considered**
N/A

**Additional context**
Pro-tip: Use well-defined labels, [like these](http://label-schema.org/rc1/). I'd add `org.label-schema.vcs-ref`, `org.label-schema.build-date` and `org.label-schema.schema-version`.

robertsLando:
@LordMike I think you mean Docker labels right? Like this: 
https://github.com/zwave-js/zwavejs2mqtt/blob/master/docker/Dockerfile#L38

But with some more infos like: https://github.com/hassio-addons/addon-zwave2mqtt/blob/master/zwave2mqtt/Dockerfile#L83-L100

LordMike:
Exactly. :)

Damn. That's a large set of labels for the Hassio setup :O

robertsLando:
I think this could be enought: 


```Dockerfile
# Labels.
LABEL org.label-schema.schema-version="1.0"
LABEL org.label-schema.build-date=$BUILD_DATE
LABEL org.label-schema.name="zwave-js/zwavejs2mqtt"
LABEL org.label-schema.description="Zwave to Mqtt Gateway and control panel"
LABEL org.label-schema.url="https://zwave-js.github.io/zwavejs2mqtt/#/"
LABEL org.label-schema.vcs-url="https://github.com/zwave-js/zwavejs2mqtt"
LABEL org.label-schema.vcs-ref=$VCS_REF
LABEL org.label-schema.vendor="zwave-js"
LABEL org.label-schema.version=$BUILD_VERSION
LABEL org.label-schema.docker.cmd="docker run --rm -it -p 8091:8091 --device=/dev/ttyACM0 -v $(pwd)/store:/usr/src/app/store zwavejs/zwavejs2mqtt:latest"
```

robertsLando:
[...]

PR #266, opened 2021-01-18
chore(docker): docker labels opencontainers v2

chrisns:
alternative for #257 to #262


robertsLando:
@chrisns What about adding the git hash, build date and other labels? Should we calculate them in a previous step and put them in outputs?

chrisns:
The plug-in does those automatically

robertsLando:
Also with that I could remove the label mantainer inside the dockerfile right?

chrisns:
See the build log output

chrisns:
> Also with that I could remove the label mantainer inside the dockerfile right?

Yes, sorry forgot to tidy that up

robertsLando:
@chrisns Check my latest changes

PR #272, opened 2021-01-19
fix: hexColor parsing

### LINKED ISSUE #269: [bug] Sending '000000' to new hexColor yields error
**Version**

Build/Run method
- [x] Docker
- [ ] PKG

zwavejs2mqtt version: `dev`, with new hexColor valueId

**Describe the bug**
The new hass templates indicated to me the format of the RGB value was all hex, ie `112233`, and not as with Z2M: `#112233`. So I went with this.

Sending f.ex. `ffffff` turns a light all white, but conversely `000000` yields an error. It's interpreted as a number. 

**To Reproduce**
1. Send `000000` to `zwavejs2mqtt/NAME/51/1/hexColor/set`
2. Observe error in logs and no color change

**Expected behavior**
Value is interpreted as hexadecimal value, and lights turn off.

**Additional context**
Log of error:
image

Log of success:
image

robertsLando:
Fixes #269 

@LordMike could you try this?

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -559,8 +559,8 @@ function copy (obj) {
  * @param {string} id object id of the hass discovery payload
  * @returns true if the discovery payload object id is a rgb_dimmer
  */
-function isRgbDimmer (id) {
-  return id.startsWith('rgb_dimmer')
+function isLight (id) {
+  return id.startsWith('light_')
 }
 
 /**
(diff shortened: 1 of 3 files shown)


LordMike:
Would love to test it. The docker build failed, could it be rerun?

robertsLando:
@LordMike Re running, didn't know why it failed btw

LordMike:
Seems both `pr-272` and `sha-e2c0b10` tags are missing from docker hub.. Other sha-* tags are present, but I can't find any pre-*. .. Odd that the sha-* tags are updated like, 32 minutes ago .. so maybe they disappear over time?

robertsLando:
@LordMike THem are not pushed to dockerhub. Only things on master are builded

LordMike:
Oh - could pr's be pushed as well?

Perhaps remove then with an action, when the PR is closed/merged.. ?

robertsLando:
> Perhaps remove then with an action, when the PR is closed/merged.. ?

Need to think about this. @chrisns What do you think?

chrisns:
We don't pay for storage. So I wouldn't bother tidying up. You'll only be able to push PRs on this repo, forks won't work because secrets won't work fir good reason

chrisns:
Not sure if it would fit in the artefacts GitHub let you keep on a job run, you might be able to dump it to a tar and leave it there if someone wants it they could download the tar then docker load

robertsLando:
> Not sure if it would fit in the artefacts GitHub

chrisns:
It's not an issue of the user. What you're implying is in effect publishing the docker hub creds. Do not do this

chrisns:
If it fits in GitHub artefacts it's actually pretty simple. On phone can't see notes on limit or what our image size would be as a zip/tbz

chrisns:
If you just want to push on branches and ignore forks you can do that easily mind

PR #289, opened 2021-01-20
fix: remove markNodeAsFailed api

robertsLando:
Ref 

https://github.com/zwave-js/node-zwave-js/pull/1456
https://github.com/zwave-js/node-zwave-js/pull/1444#issuecomment-763699022


robertsLando:
cc @LordMike

PR #295, opened 2021-01-20
fix: scale and value templates for rgb and dimmer discovery

jcam:


robertsLando (approved):
LGTM thanks @jcam ! :)

PR #307, opened 2021-01-21
fix(hass): change currentValue to targetValue in the fan template docs

towerhand:
Change currentVaue to targetValue in the fan template like changed in:
https://github.com/zwave-js/zwavejs2mqtt/commit/0d0fbd678070dc2e979a809363815628b5a8a5d8


robertsLando (approved):
Thanks

PR #319, opened 2021-01-22
feat: add valueId, node and logger to parse functions

ahochsteger:
This PR makes the following objects available to the send and receive parse functions:
* `logger`: The `logger` object of the gateway
* `valueId`: The `valueId` object that contains more information about the affected value
* `node`: The `node` object that contains all information about the affected node

It allows something like this which was not possible before:
```javascript
// Some debugging:
logger.debug('Got the following arguments:')
logger.debug('- value: ' + value)
logger.debug('- valueId: ' + JSON.stringify(valueId))
logger.debug('- node: ' + JSON.stringify(node))

// Check for invalid values:
if (value < -1) {
  // Log a warning if invalid value reports have been fixed:
  logger.warn('Invalid value ' + value + ' detected - setting to -1.')
  value = -1
}
return value
```


robertsLando:
@ahochsteger I would also add the node object, before the logger :) ALso remember to update docs

ahochsteger:
> @ahochsteger I would also add the node object, before the logger :) ALso remember to update docs

Good idea, will add it and have a look at the docs ...

robertsLando (changes requested):
dito

ahochsteger:
> @ahochsteger I would also add the node object, before the logger :)

How would I get the node object in evalFunction() as it just has access to `code`, `valueId` and `value`?
Or did you mean the `valueId` object?

ahochsteger:
Ok, I can add it to `evalFunction` as 4th argument.
Does it make sense to provide `valueId` as well as it may contain more interesting information?

robertsLando:
@ahochsteger Yep I would add in order `(value, valueId, node, logger)`

ahochsteger:
> @ahochsteger Yep I would add in order `(value, valueId, node, logger)`

Done locally.
What about the receive parse function?
There doesn't seem to be a `node` object available in `parsePayload(payload, valueId, valueConf)`.
Should I pass an empty object (`{}`) or `null`?

ahochsteger:
I'm using an empty object for `node` for now at the receive part.

lib/Gateway.js
```diff
@@ -807,7 +812,8 @@ Gateway.prototype.parsePayload = function (payload, valueId, valueConf) {
         const parsedVal = evalFunction(
           valueConf.receiveFunction,
           valueId,
-          payload
+          payload,
+          {}
```
robertsLando, line 816: I suggest to fetch the node here someway :) maybe add the node to the parsePayload function signature too (remember to update all calls of this function then

robertsLando (changes requested):

lib/Gateway.js
ahochsteger: @robertsLando  I guess I'll need some help here on how to proceed:
Have I seen it correctly that the `nodeId` can (always?) be accessed using `valueId.nodeId` and the node then using `this.zwave.nodes[nodeID]`?
If so, we do not even need to pass `node` to `evalFunction` at all since we could fetch `node` this way within that anyway.
Or am I missing something?

PS: I'm looking forward to have TypeScript here, as it always shows you the types without looking at every line that might put properties into an object to see the full list of properties ;-)

ahochsteger: Hmmm ... it seems not to work in `evalFunction` but from within `parsePayload` I can get the node using `this.zwave.nodes`.
It's already pushed, I'm not sure, if it is a safe approach though.

lib/Gateway.js
```diff
@@ -803,11 +808,13 @@ Gateway.prototype.parsePayload = function (payload, valueId, valueConf) {
         payload = eval(payload + op)
       }
 
+      const node = this.zwave.nodes[valueId.nodeId]
```
robertsLando, line 811: Place this inside the if statement

src/components/dialogs/DialogGatewayValue.vue
```diff
@@ -172,7 +176,7 @@
           color="blue darken-1"
           text
           @click="$refs.form.validate() && $emit('save')"
-          >Save</v-btn
+          >Update</v-btn
```
robertsLando, line 179: We could make this to be `Add` or `Update` based on if it's a new or an existing value

robertsLando (changes requested):

lib/Gateway.js
robertsLando: > ave I seen it correctly that the nodeId can (always?) be accessed using valueId.nodeId and the node then using this.zwave.nodes[valueId.nodeId]?

lib/Gateway.js
ahochsteger: done.

src/components/dialogs/DialogGatewayValue.vue
ahochsteger: The only way I've seen to detect a new or existing value is to check, if editedValue is an empty object when the dialog is shown and store that in a data property (later it's not empty anymore).
Or do you know a better way to detect this (without adding a new component property)?

ahochsteger: done.

lib/Gateway.js
ahochsteger: Hmm, it has been working for me so far.
Are there any edge cases that you can imagine?
Can you suggest a better way?

src/components/dialogs/DialogGatewayValue.vue
```diff
@@ -209,6 +209,10 @@ export default {
     // eslint-disable-next-line no-unused-vars
     value (val) {
       this.$refs.form && this.$refs.form.resetValidation()
+      if (val) {
+        this.isNew = !this.editedValue.device
+        console.log('Set isNew to ' + this.isNew)
```
robertsLando, line 214: @ahochsteger Remove this

robertsLando (changes requested):

lib/Gateway.js
robertsLando: @ahochsteger the way you did it is ok :)

robertsLando: @ahochsteger Sorry maybe I explained wrong, it was ok to get the node in that way, instead of passing `{}`. Could you revert your last commit?

ahochsteger: Oh, sorry, it's reverted again.

src/components/dialogs/DialogGatewayValue.vue
ahochsteger: done.

Issue #254, opened 2021-01-15
[question] how to properly recreate from scratch an entity in homeassistant?


dvbit:
What would be the right sequence?
Remove entity from ha
Remove store in z2m
Stop ha
Delete entity from 
Mqtt 
Update node info in z2m
Heal node ?
Check log for the above to finish 
Restart ha
Rediscover node

Is that correct ?

robertsLando:
@billiaz @chilicheech I can speak about the zwavejs2mqtt side, could one of you provide a detailed answer about the complete steps? Could also be added to hass docs too

LordMike:
In HASS, it's quite simple. If you remove the retained discovery document for an entity, the entity disappears in HASS. If all entities from a device disappear, the device itself disappears.

An zwave2mqtt at least, did not push discovery documents unless it had to. So simply removing the mqtt topics, will remove it from HA "entirely".. .. in quotes, as the logger will still have data in it, so if you were to create entities with the same unique ids, they would "see" the historical data.

For z-wave, a refresh should be enough to clear all, right?.. Else, exclude (or remove failed & reset hardware device), then include again.

robertsLando:
@LordMike This could be good to add to docs, if you have time :)

dvbit:
Thanks! might be worth adding to the docs :-)

LordMike:
@robertsLando I think this file is appropriate?

https://github.com/zwave-js/zwavejs2mqtt/blob/91eef10f24c6568aefd51aad5c77f7e771d8eb53/docs/guide/homeassistant.md

robertsLando:
@LordMike Yepp!

PR #335, opened 2021-01-23
docs: removing entities from HASS

LordMike:
See #254

Needs:

* [ ] Check if steps for re-discovery (push discovery doc) is correct.


robertsLando:
@LordMike Let me know when this is ready, will be useful!

LordMike:
Actually. Since disabling the HASS integration on my end, I can't see that part of the UI anymore.. Can you confirm if the steps for rediscovery are correct?

LordMike:
Also. With github actions, it's possible to run some CI stuff only if certain paths are affected. So for these PR's, you could avoid rebuilding the application and running linters etc.. 

This could be done by adding path arguments to all jobs, where docs stuff only  runs when `docs/**` is affected, and the source stuff only runs when `src/**` (or whatever) is affected.

[Docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths).

robertsLando:
@LordMike Thanks for the suggestion about the ignore path, didn't know about that! 

Anyway the steps are a little different, you should press on `Rediscover Node` button and now node details are shown in the expanded row of the table

robertsLando:
@zwave-js-bot fix lint

LordMike:
I've changed it to `Rediscover Node`:

> Note that in order for a removed entity to appear again, it must be published by **zwavejs2mqtt** again. This happens automatically for new devices, if enabled, alternatively it can be done manually by selecting the node in **zwavejs2mqtt**, and then for each HASS device clicking `Rediscover Node`.

PR #343, opened 2021-01-25
feat: add pollValue api

### LINKED ISSUE #309: [feat] API to issue getValue for a single value
On topic with the polling we're discussing elsewhere, perhaps an API to `getValue` for a single thing? :)

There is `refreshValues` right now, but it's probably too much if just one value is desired.

Add a `getValue` API accepting the node, cc, endpoint and index (and possibly partial value). F.ex.:

> [12, 112, 0, 1, 8]

.. for Node 12, cc 112 (configs), endpoint 0 parameter 1, partial value 8

Should the `getValue` work on partial params?.. You _could_ ignore that, and just operate on the whole value - I expect the partials are updated when the full value is updated.

The MQTT topic could be in the `api` part, but it could also make sense to place it with the published values. Much like `targetValue` has a `/set`, you could make `currentValue` have a `/get` or `/refresh`.

I've used configuration as example. What I'm really interested in is a meter reading.. But I suppose the same solution applies to configurations, multilevels, switches, etc..

N/A

**Additional context**
N/A

robertsLando:
Fixes #309

Needs zwave-js 6.0.3


robertsLando:
@blhoward2

snagytx:
Would be useful if you could also provide a sample MQTT payload and topic. Something that you used during your development/testing.

robertsLando:
@snagytx https://zwave-js.github.io/zwavejs2mqtt/#/guide/mqtt?id=zwave-apis

snagytx:
The documentation didn't really provide an example of valueId object. Your update helps.

I'll keep an eye on the merge and once a docker image is available I will test it. I'm eager to use this feature.

LordMike:
Was this pushed as a `pr-nnn` tag?

```
> docker pull zwavejs/zwavejs2mqtt:pr-343
Error response from daemon: manifest for zwavejs/zwavejs2mqtt:pr-343 not found: manifest unknown: manifest unknown
```

snagytx:
Even if it was it’s still not working. It needs a new version of zwave-is,
the current on 6.0.2 doesn’t have the required code, the zwavejs2mqtt needs
to use that version for everything to work.

robertsLando:
> Was this pushed as a pr-nnn tag?

Nope

robertsLando:
> It needs a new version of zwave-js,
the current on 6.0.2 doesn’t have the required code, the zwavejs2mqtt needs
to use that version for everything to work.

You can test it by using `Dockerfile.contrib` : https://zwave-js.github.io/zwavejs2mqtt/#/troubleshooting/docker_custom_builds

robertsLando:
@billiaz would you like to bump zwavejs 6.1.0 here?

billiaz:
Will do and test

LordMike:
So, I managed to make my build server build images for me.. Should make it easier to test stuff now .. :)

And damn. Master looks neat :) - now we just need the "send to all nodes" part of issuing commands respect filters :D Anyhow.

I'm running z2m 423a7e7 and zwjs 6.1.0-5d49c45 - so should be this branch and master of zwavejs. I'm having trouble invoking this, because I'm not sure what property and property key are. I'm trying to update a meter, which publishes it values to the following topic:

> zwavejs2mqtt/meter_gas/50/0/value/131073

So I'm trying with this document (nodeId is 4):

```json
{
    "args": [{
            "nodeId": 4,
            "endpoint": 0,
            "commandClass": 50,
            "property": "value",
            "propertyKey": "131073"
        }
    ]
}
```

Various mixes of `property` and `propertyKey` (both "Value", both the number, swap them...) .. do not seem to work.

I'm getting `Meter: "131073" is not a supported property key for property "value"` back.

LordMike:
I fetched the node json for my node, and actually found the parameters (property, ..) in it for my value.. So I've tried with those - it still fails, but now I know what they're meant to be set to .. :)

Value export:

```json
{
    "id": "4-50-0-value-131073",
    "nodeId": 4,
    "commandClass": 50,
    "commandClassName": "Meter",
    "endpoint": 0,
    "property": "value",
    "propertyName": "value",
    "propertyKey": 131073,
    "propertyKeyName": "Gas_Cubic meters_Consumed",
    "type": "number",
    "readable": true,
    "writeable": false,
    "label": "Value (Gas, Consumed)",
    "stateless": false,
    "ccSpecific": {
        "meterType": 2,
        "rateType": 1,
        "scale": 0
    },
    "unit": "Cubic meters",
    "list": false,
    "value": 536062.066,
    "lastUpdate": 1611701226087,
    "newValue": 536062.066
}
```

My payload to `zwavejs2mqtt/_CLIENTS/ZWAVE_GATEWAY-HomeMQTT/api/pollValue/set`

```json
{
    "args": [{
            "nodeId": 4,
            "endpoint": 0,
            "commandClass": 50,
            "property": "value",
            "propertyKey": 131073
        }
    ]
}
```

Response:
```
{
    "success": false,
    "message": "Meter: \"131073\" is not a supported property key for property \"value\"",
    "args": [{
            "nodeId": 4,
            "endpoint": 0,
            "commandClass": 50,
            "property": "value",
            "propertyKey": "131073"
        }
    ]
}
```

LordMike:
Scratch that.. I can see my MQTT client is still showing me the last response. The response args still had the quotes on the propertyKey.. I think I'm not getting a response, because the message is queued and the promise only returns when the value is polled.

@robertsLando do you think this API should return when the poll is queued? - or as now, wait for the poll to succeed?

LordMike:
And all the polled values came back (AlCalzone hasn't deduplicated them yet :)).

So for me. This works. :)

Minor issue reg. the response of the API coming when the value was actually polled - but I can live with that.

robertsLando:
> I can see my MQTT client is still showing me the last response.

What? Is it retained? COuld you enable Debug level logs and show me them?

> I think I'm not getting a response, because the message is queued and the promise only returns when the value is polled.

I don't think so, the call is awaited

LordMike:
> 
> 
> > I can see my MQTT client is still showing me the last response.
> 
> What? Is it retained? COuld you enable Debug level logs and show me them?

No, all is fine - MQTT explorer just doesn't remove it while the app is running.. 

> 
> > I think I'm not getting a response, because the message is queued and the promise only returns when the value is polled.
> 
> I don't think so, the call is awaited

Exactly - so the promise first returns, when the value is actually polled - the docs don't say, but they do say the promise will contain the polled value, so it must wait. So this API won't reply back, until there actually is a response. I was expecting it to reply back immediately saying "successful" (as the message was queued) - I didn't expect it to reply first when the value was received by the controller.

It's fine, if this is intentional. It means though, that if the API was ever exposed via HTTP f.ex., you would respond to this call before the value was polled - that could be one very long HTTP request.. :)

robertsLando:
> So this API won't reply back, until there actually is a response. I was expecting it to reply back immediately saying "successful" (as the message was queued) - I didn't expect it to reply first when the value was received by the controller.

@AlCalzone I think he is right, I mean the api should return when the poll request is sent, then the value update will be catched by the value update event

LordMike:
I recall seeing other API methods within zwave-js do the same - the Promise returns when the response for the API call comes back.. f.ex. `refreshValues` says the following: 

> Although this method returns a Promise, it should generally not be awaited, since the update may take a long time.

As a library user, I think this is cool. It makes for easy coding, as we don't need to do a lot of work to identify the polled value - when it comes back.

LordMike:
I dunno.. Maybe a `bool: waitForReply` parameter ?

AlCalzone:
It depends on the use case imo. 

If you just want to tell the device to send it's updated value whenever it can, don't await.
If you need to work with the value, awaiting is fine as long as you know it might take a while.

Also keep in mind that there is an open pr that allows every GET command to return nothing if there was no response. The pollValue types already reflect that.

snagytx:
I suspecting most of the use cases from the MQTT perspective would be fire and forget. Since you will get the response on the normal topic. 

In my use case I need to get the current state of the Doorlock once a notification is received over MQTT since my device is not sending the updated status. So this would be a fire and forget use case.

I can confirm that it works with BE469 Door lock which requires a poll after a notification was sent.

robertsLando:
@LordMike @snagytx @AlCalzone With last changes now it's possible to enable the poll of values from the ui. The interval is handled by me

LordMike:
@snagytx 

If you need an image.

registry.mbwarez.dk/internal/zwavejs2mqtt:pr-343

snagytx:
@LordMike

Thank you, I was able to build my own and all seems to be working perfect. My use case with near realtime response is covered by the first changes. Periodic interval checking might be good for certain use cases but it can easily cause performance issues.

I would still like to have the option fire and forget option, but this is a "like" not a "need"

PR #355, opened 2021-01-26
feat(ui): show confirm improvments

robertsLando:


robertsLando:
@ahochsteger Look at this,will be useful in future ;)

PR #369, opened 2021-01-27
chore: define basic types

robertsLando:
@ahochsteger 

TODOs:

- [x] Add methods to Gateway Zwave/Mqtt clients (use `dtsmake -s Gateway.js`)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -7,7 +7,6 @@ import {
   InterviewStage,
   NodeStatus,
   ValueType,
-  ZWaveNode,
   ZWaveOptions
 } from 'zwave-js'
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,12 +1,16 @@
 import { ZwavejsServer } from '@zwave-js/server'
 import { EventEmitter } from 'events'
-import { MqttClient as Client } from 'mqtt'
+import { MqttClient as Client, IClientPublishOptions } from 'mqtt'
 import { Socket } from 'net'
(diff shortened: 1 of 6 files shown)


ahochsteger:
Whoa, TypeScript comes to life!

robertsLando:
@ahochsteger Now I have also added types on signatures, better then nothing :)

robertsLando:
@ahochsteger Not really typescript but at least now we have some better defined types

types/index.d.ts
```diff
@@ -0,0 +1,424 @@
...
+  CommandClass,
+  InterviewStage,
+  NodeStatus,
+  ValueID,
+  ValueType,
+  ZWaveNode,
+  ZWaveOptions
+} from 'zwave-js'
+
+export type Z2MValueIdState = {
```
AlCalzone, line 16: Normally, you'd use an `interface` for these kinds of types (also supports type inheritance with `extends`).

robertsLando: Couldn't use the `&` operator instead of the `extends` with types?

robertsLando: BTW this is just a temporary fix, I'm looking forward to switch to typescript completely. Also I think that using a namespace here (instead of prefix names with `Z2M` would be better`

AlCalzone: & is fine to combine types. It is just recommended to use interfaces for these simple object types without conditionals.

AlCalzone: So that also applies to TS later

PR #407, opened 2021-02-01
feat(ui): show/hide node table columns

ahochsteger:
This PR provides the possibility to show / hide individual table columns.

Button to show the selection menu:
image

Menu to choose the columns from:
image

The headers array allows to customize each column using the properties `active` (is currently visible or not) and `selectable` (is selectable in the menu or not).

Some not so important columns are now not shown by default (see screenshot above). 
Since this reflects my personal opinion I'm interested in feedback, what the defaults should be.

It lays the foundation to provide much more data in the nodes table (e.g. battery level, ...) without overloading the number of columns.


robertsLando (approved):
LGTM! Easy clean and useful. Thanks :)

robertsLando:
@ahochsteger Feel free to squash and merge once ready

ahochsteger:
@robertsLando thanks for your feedback!

Are you ok with the defaults I've chosen to not show certain columns by default?

robertsLando:
@ahochsteger I would keep all visible by default, and also I would store the shown in localstorage (like for the filters etc)

ahochsteger:
@robertsLando the list of columns is now stored in local storage and the full list of headers is shown by default.
As a consequence I've simplified the code a lot (eliminated the `selectable` and `active` header attributes since they are not really requred).

robertsLando:
@ahochsteger In order to fix the UI problems I think we could set a max number of columns based on the screen size. What do you think about this?

ahochsteger:
> @ahochsteger In order to fix the UI problems I think we could set a max number of columns based on the screen size. What do you think about this?

@robertsLando  I'm not sure, if the number of columns is a good solution here if you compare the width of the "ID" column with the "Product" column that shows for example something like "A real push button (switch) available in several colors." (which should be reduced in the device config anyway).

robertsLando:
@ahochsteger What other solution do you propose to fix the other UI problem?

ahochsteger:
> @ahochsteger What other solution do you propose to fix the other UI problem?

I'm not sure, which "other" UI problem you are referring to in this context?

robertsLando (approved):
LGTM

PR #413, opened 2021-02-01
fix(ui): labels and mobile view

### LINKED ISSUE #399: [bug] Can't scroll device horizontally without opening home assistant tab
Before submitting a bug please read: <https://zwave-js.github.io/zwavejs2mqtt/#/troubleshooting/bug_report>

### Version

Build/Run method

- [ ] Docker
- [ ] PKG
- [x] Manually built (git clone - npm install - npm run build )

zwavejs2mqtt version: 1.0.1
zwavejs version: 6.1.0

### Describe the bug

When using using mobile, the configure screen on multi-column devices always scrolls to the "Home Assistant" tab instead of allowing to scroll over to the columns to the right. 

### To Reproduce

Screenshot will be attached, I have a Multisensor 6 which has a lot of config options. I'm unable to get to the right 2 columns. 

### Expected behavior

I would like access to these config items when using safari on iPhone. 

### Additional context

Downloaded and built from GitHub about 24 hours ago. Running on a raspberry pi 4. Using safari on iPhone 11 Pro.

robertsLando:
- [x] Better labels
- [x] Fix expandable menu on mobile devices
- [ ] Table horizontal scroll on overflow

Fixes #399

cc @blhoward2


robertsLando:
@ahochsteger

robertsLando:
@ahochsteger For some reasons seems that the v-select isn't respecting display breakpoints, I fixed byt setting a max-width on that for now

robertsLando:
There is definetly some bugs in the grid system. Need to figure out what's going on there

robertsLando:
@ahochsteger Ok I have found the problem. It happens when the view is between 600 (mobile breakpoint for data table) and 1200px. In this case the table hasn't enought space to show all columns so it adds an horiontal scroll that results in wrong widths in the expanded item too. We should find a way to prevent this

robertsLando:
@ahochsteger This problems

robertsLando:
@ahochsteger Switched to the new grid system :tada:

ahochsteger:
@robertsLando great!
And does it solve some of the UI problems?

robertsLando:
> And does it solve some of the UI problems?

Nope, btw wasn't that the reason why I did it. To fix them we should check the available width and make the table not scrollable horizontally

robertsLando:
Merging this for now, will look at others ui problems in another PR

PR #419, opened 2021-02-02
feat(ui): improved add remove device UX

jarrettv:
I'm not happy with having to compare all these status strings. It feels like things could be cleaner if some of the logic happened upstream (perhaps `ZwaveClient.js`). However, it does seem to be working well.


src/components/ControlPanel.vue
```diff
@@ -74,15 +94,53 @@ export default {
...
+    ...mapGetters(['nodes', 'appInfo', 'zwave'])
+  },
+  watch: {
+    addRemoveEndDate: {
+      immediate: true,
+      handler (newVal) {
+        if (this.addRemoveTimer) {
+          clearInterval(this.addRemoveTimer)
+        }
+        this.timer = setInterval(() => {
```
robertsLando, line 110:
```suggestion
        this.addRemoveTimer = setInterval(() => {
```

Typo?

robertsLando (changes requested):

jarrettv:
nice catch @robertsLando. Also, should we update `nodes-management.md` with this PR?

src/components/ControlPanel.vue
```diff
@@ -74,15 +94,53 @@ export default {
...
+        }, 100)
+      }
+    },
+    appInfo: {
+      deep: true,
+      handler (newVal) {
+        if (newVal.controllerStatus.indexOf('clusion') > 0) {
+          this.onAddRemoveStateChange(newVal.controllerStatus)
+        }
+      }
```
robertsLando, line 129: A cleaner way to do this is to create a `computed` property let's call this `controllerStatus` that returns `this.appInfo.controllerStatus` (or even better call it inclusionStatus and make the checks needed there) and then watch only that :)

src/components/ControlPanel.vue
```diff
@@ -201,6 +243,58 @@ export default {
...
+      const action = data.mode === 'Exclusion' ? `${startStop}Exclusion` : `${startStop}Inclusion`
+      const args = []
+      if (data.mode !== 'Exclusion' && startStop === 'start') args.push(data.mode.indexOf('Secure') === 0)
+      this.apiRequest(action, args)
+    },
+
+    onAddRemoveStateChange (controllerStatus) {
+      const nodeCount = this.addRemoveMode === 'Exclusion'
+        ? this.nodes.filter(x => x.status === 'Removed').length
+        : this.nodes.filter(x => x.status !== 'Removed').length
```
robertsLando, line 261: I would move this on the lower if case as it's not used by all

robertsLando: What about wathing for node added and node removed socket events instead?

robertsLando (changes requested):

robertsLando:
> nice catch @robertsLando. Also, should we update nodes-management.md with this PR?

That would be good, yes :) Maybe with some screenshots even better

lib/ZwaveClient.js
```diff
@@ -245,6 +245,7 @@ function onNodeAdded (zwaveNode) {
   // the driver is ready so this node has been added on fly
   if (this.driverReady) {
     addNode.call(this, zwaveNode)
+    this.sendToSocket(socketEvents.onNodeAdded, zwaveNode)
```
robertsLando, line 248:
```suggestion
    this.sendToSocket(socketEvents.nodeAdded, zwaveNode)
```

src/components/ControlPanel.vue
```diff
@@ -74,15 +95,58 @@ export default {
...
+    timeoutMs () {
+      return (this.zwave.commandsTimeout * 1000) + 800 // add small buffer
+    },
+    controllerStatus () {
+      return this.appInfo.controllerStatus
+    }
+  },
+  watch: {
+    addRemoveEndDate: {
+      handler (newVal) {
```
robertsLando, line 112: You can make this cleaner:

`addRemoveEndDate(newVal) {}`

src/components/ControlPanel.vue
```diff
@@ -74,15 +95,58 @@ export default {
...
+            this.addRemoveWorking = `${this.addRemoveMode} started: ${s}s remaining`
+          }
+          if (this.now > newVal) {
+            this.now = newVal
+            clearInterval(this.addRemoveTimer)
+          }
+        }, 100)
+      }
+    },
+    controllerStatus: {
```
robertsLando, line 129: Same

src/components/ControlPanel.vue
```diff
@@ -74,15 +95,58 @@ export default {
...
+          this.onAddRemoveStateChange(newVal)
+        }
+      }
+    }
   },
-  watch: {},
   data () {
     return {
       settings: new Settings(localStorage),
+      now: new Date(),
```
robertsLando, line 140: Don't use this var, calcolate it when needed

src/components/ControlPanel.vue
```diff
@@ -201,6 +249,55 @@ export default {
...
+      this.addRemoveSucceeded = null
+      this.addRemoveFailed = null
+      this.addRemoveWorking = `${data.mode} ${data.status === 'start' ? 'starting…' : 'stopping…'}`
+      const action = data.mode === 'Exclusion' ? `${data.status}Exclusion` : `${data.status}Inclusion`
+      const args = []
+      if (data.isSecure) args.push(data.mode.indexOf('Secure') === 0)
+      this.apiRequest(action, args)
+    },
+
+    onAddRemoveStateChange (controllerStatus) {
```
robertsLando, line 265: I would move all the function content in the watch method, no need to pass the controllerStatus as it's a computed var

src/components/ControlPanel.vue
```diff
@@ -334,12 +431,25 @@ export default {
...
+    this.socket.on(socketEvents.nodeAdded, async node => {
+      console.debug(node)
+      this.addRemoveNode = node
+    })
   },
   beforeDestroy () {
     if (this.socket) {
       // unbind events
       this.socket.off(socketEvents.api)
+      this.socket.off(socketEvents.nodeRemoved)
```
robertsLando, line 449: We should check if this off also disables the nodeRemoved handler in App.vue

robertsLando (changes requested):

jarrettv:
Semantic PR title / commit messages ???

robertsLando (changes requested):
I have submitted a PR to your fork. Merge that and we are done here :)

robertsLando:
Thanks @jarrettv !

PR #422, opened 2021-02-02
fix: binary sensor notification changes break double taps and other events

blhoward2:
Closes #421 and https://github.com/zwave-js/node-zwave-js/issues/1555. Recent changes to how binary sensors are created for notification events (#396) broke double tap via basic set events because of the nesting of the else statement (valueId.list was not true, so the if block in which sensor_generic was setup was never entered). Fixing that caused other devices to have the undefined error where valueId.list was true, but the value.state.length was not 2. 

I changed the top-level if statement to only enter if valueId.list is true && valueIf.states.length === 2. In all other cases the else block should be entered to setup a sensor_generic. I have tested this and it works fine now. It is also much cleaner and gets rid of the nested if statement.

While reviewing this, I realized that we were defining a bunch of binary_sensor types in configurations.ts that are never used. We were only ever loading binary_sensor_contact, then overriding the object_id and device class depending on the case. We never used any of the others. I've now changed this so we setup a generic binary sensor, with device class None, and then we override the objectId and device class if appropriate. @billiaz Correct me if I'm wrong here, but I did a grep and the others aren't referenced anywhere else but in configurations.ts.

PR #433, opened 2021-02-03
fix: improve binary sensor code - add more sensors

aretakisv:
This PR tries to improve a bit more Pull #422 

- Convert repeated code into function.
- Add more devices


lib/Gateway.js
```diff
@@ -1680,65 +1702,76 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
+              discoveredObjectId = valueId.propertyName
               break
             default:
           }
+          // cfg not there?
+          cfg = cfg ? copy(cfg) : copy(hassCfg.binary_sensor)
+          // correct payload from true/false to numeric values
           cfg = setBinaryPayloadFromSensor(cfg, valueId, off)
+          // finally update object_id
+          cfg.object_id = discoveredObjectId
```
robertsLando, line 1774: @billiaz Shouldn't this be:

```suggestion
          cfg.object_id += discoveredObjectId
```

lib/Gateway.js
```diff
@@ -1680,65 +1702,76 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
+                  )
                   break
                 default:
               }
-              cfg.object_id = valueId.propertyName
-              break
-            case 'Water Alarm':
-              cfg.discovery_payload.device_class =
-                Constants.deviceClass.sensor_binary.MOISTURE
+              discoveredObjectId = valueId.propertyName
```
robertsLando, line 1765: Why? By default it is set to `property`

robertsLando (changes requested):

lib/Gateway.js
billiaz: we replace the object ID from the default. to fit the one we need to use

billiaz: i agree, let me make some changes, as this  'binary_sensor_' prefix misses everywhere

lib/Gateway.js
billiaz: yeap! right! :)

lib/Gateway.js
```diff
@@ -662,6 +662,23 @@ function setBinaryPayloadFromSensor (cfg, valueId, offStateValue = 0) {
...
   return cfg
 }
+
+/**
+ *
+ * @param {import('../types').HassDevice} cfg Hass discovery Configuration
+ * @param {string} devClass Choosen device class
+ * @param {boolean} reversePayload reverse payload order
+ */
+function setBinarySensorPayload (devClass, reversePayload = false) {
```
robertsLando, line 672:
```suggestion

/**
 * Create a binary sensor configuration with a specific device class
 *
 * @param {string} devClass Choosen device class
 * @param {boolean} reversePayload reverse payload order
 * @returns {import('../types').HassDevice} cfg Hass discovery Configuration for the binary sensor
 */
function getBinarySensorConfig (devClass, reversePayload = false) {
```

Change all the other references according to this

lib/Gateway.js
```diff
@@ -1680,65 +1702,76 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
-              break
-            case 'Water Alarm':
-              cfg.discovery_payload.device_class =
-                Constants.deviceClass.sensor_binary.MOISTURE
+              discoveredObjectId = valueId.propertyName
               break
             default:
           }
+          // cfg not there?
+          cfg = cfg ? copy(cfg) : copy(hassCfg.binary_sensor)
```
robertsLando, line 1770:
```suggestion
          cfg = cfg || copy(hassCfg.binary_sensor)
```

No need to copy it here, it's already copied in `getBinarySensorConfig` function

robertsLando (changes requested):

lib/Gateway.js
```diff
@@ -1680,65 +1703,76 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
-              cfg.discovery_payload.device_class =
-                Constants.deviceClass.sensor_binary.MOISTURE
+              discoveredObjectId = valueId.propertyName
               break
             default:
           }
+          // cfg not there?
+          cfg = cfg || copy(hassCfg.binary_sensor)
+          // correct payload from true/false to numeric values
           cfg = setBinaryPayloadFromSensor(cfg, valueId, off)
```
robertsLando, line 1773:
```suggestion
          setBinaryPayloadFromSensor(cfg, valueId, off)
```

No need to re-assign cfg here, you pass it as reference

robertsLando (changes requested):

robertsLando (approved):
LGTM

PR #455, opened 2021-02-04
docs: poll a value

robertsLando:
cc @AlCalzone @marcelveldt


marcelveldt:
Nice!

AlCalzone:
One thing I'd add:
Many values can only be polled together. For example `targetValue`, `currentValue` and `duration` for most of the switch-type CCs. Before enabling polling for all values of a node, users should check whether polling a single value already updates the desired other values too.

robertsLando:
Will do it in another one, I have merged settings in the other branch I'm working on

PR #456, opened 2021-02-04
fix(ui): settings refactor, fix mesh node names

robertsLando:
cc @marcelveldt


src/components/nodes-table/ExpandedNode.vue
```diff
@@ -53,6 +53,11 @@ export default {
     HomeAssistant,
     NodeDetails
   },
+  computed: {
+    showHass () {
+      return this.node.hassDevices && Object.keys(this.node.hassDevices).length > 0
```
marcelveldt, line 58: Isn't it true that you configure hassDevices with this tab ?
In other words: shouldn't you tie the visibility of the HA tab to the mqtt being enabled OR some toggle to enable HA MQTT discovery ?

robertsLando: You configure the ones I have discovered, you canno create new. If discovery is disabled no one will be created here

robertsLando: BTW settings have been completely refactored now. It's more clear and easy to use the gateway as only control panel

PR #458, opened 2021-02-04
feat(ui): improve mesh graph

### LINKED ISSUE #384: [feat] Improved Z-Wave Network Visualization
**Current Visualization**

The current visualization of the Z-Wave mesh network gives a good overview of the network but does not work well for bigger networks (shows big ball of wool ;-) ).
Additional information or improved visualization is missing that can help analyze problems of the network. In addition the network visualization could be used to directly interact with the nodes.

Screenshot:
image

**Proposed Features**

It has the potential of doing much more.
Here are some ideas, as a starting point for discussion:
* **Show information:**
  * link quality between nodes (bi-di vs. single-direction)
  * node status
  * battery or mains powered
  * battery level
  * number of hops to controller
  * node queue length
  * node errors
  * response times (support in zwave-js coming soon)
* **Grouping of nodes:**
  * no grouping
  * by location, device type, ... (similar to nodes table goruping)
* **Filtering of nodes:**
  * show all or only relevant nodes for routing
  * filter by different criteria (similar to nodes table filter)
* **Interaction with nodes:**
  * highlight neighbors on click
  * trigger actions (heal, refresh, replace, ...)
  * direct link to node in the control panel and vice versa (quickly switch between both views)
* **User experience**
  * better visualization of large networks (see filtering above)
  * fix canvas height problems in current implementation
  * don't triggering a redraw with zoom out/in every few seconds if devices are updated
* ... (more to be discussed)

**Inspiration of existing Visualizations**

This is a quick research about what kind of visualizations are already available for Z-Wave networks for inspiration:
* ZWaveGraphHA:
  * Source+Code: <https://github.com/AdamNaj/ZWaveGraphHA>
  * Screenshot:
    image
* Home Assistant Z-Wave Graph:
  * Source+Code: <https://github.com/OmenWild/home-assistant-z-wave-graph>
  * Screenshot:
    image
* Habmin Network Viewer
  * Source: <https://community.openhab.org/t/z-wave-routing-basics/92408/2>
  * Code: <https://github.com/openhab/org.openhab.ui.habmin/blob/master/src/web/app/tools/zwave/zwaveNetwork.js>
  * Screenshot:
    image
* Z-Wave PC Controller Topology Map:
  * Source: <https://community.openhab.org/t/z-wave-routing-basics/92408/3>
  * Screenshot:
    image
* ... (please add more you may know of)

**Request for Discussion**

Let's discuss the various requirements and use cases people have and based on that implement an improved (or additional) visualization(s) of the Z-Wave mesh network.
What are your thoughts on this?

**Possible Visualization Libraries**

* vue-d3-network (current solution):
  * Code: <https://github.com/emiliorizzo/vue-d3-network>
  * Demo: https://emiliorizzo.github.io/vue-d3-network/
  * Screenshot:
    image
* vue-org-chart:
  * Code: <https://github.com/Hoogkamer/vue-org-chart>
  * Demo: <https://hoogkamer.github.io/vue-org-chart/>
  * Website: <https://freeorgchart.netlify.app/>
  * Screenshot:
    image
* D3 Chord Dependency Graph
  * Demo + Source: <https://observablehq.com/@d3/chord-dependency-diagram>
  * Screenshot:
    image
* **ioBroker Network Map:** 
  * Code: <https://github.com/AlCalzone/ioBroker.zwave2/blob/master/admin/src/lib/networkMap.ts>

robertsLando:
Fixes #384

Preview: 

Schermata da 2021-02-05 16-27-45


AlCalzone:
Show us screenshots 🥸

robertsLando:
@AlCalzone nothing good to show yet, tomorrow you will see something 🙏🏻😎

robertsLando:
Found the reason: https://observablehq.com/@d3/d3v6-migration-guide

src/components/custom/ZwaveGraph.vue
```diff
@@ -0,0 +1,990 @@
...
+      render.shapes().house = this.renderHouse
+      render.shapes().battery = this.renderBattery
+
+      const { edges, nodes } = this.listNodes()
+      // Set the parents to define which nodes belong to which cluster
+
+      // add nodes  to graph
+      for (let i = 0; i < nodes.length; i++) {
+        const node = nodes[i]
+        g.setNode(node.id, node)
```
ahochsteger, line 454: @robertsLando according to <https://github.com/dagrejs/graphlib/wiki/API-Reference#setNode> the second parameter to `setNode()` is the label, so I think it should be `g.setNode(node.id, node.id)` instead.

robertsLando: nope, label can be an object so it's correct to use the node object there AFAIK. Did you try to use a string instead?

ahochsteger: You're right. Yes, I've tried and the graph vanished ;-).

ahochsteger: Was just a misleading documentation to name it "label" if it can be an object.

ahochsteger: So maybe <https://github.com/dagrejs/graphlib/wiki/API-Reference#setDefaultNodeLabel> can be used to set the default label for nodes. I'll try.

robertsLando:
Preview: 

Schermata da 2021-02-05 16-27-45

ahochsteger:
@robertsLando I'd like to keep you up-to-date about the approach I'll take on the filter possibilities for the mesh graph we talked about privately and get some feedback on it.

Since I'd like to reuse as much of the filtering code as possible from the nodes table on the control panel I'm going to refactor much of it into an UI-independent reusable component that provides functionality that can be used by the table as well as the mesh graph.
The UI-specific parts will be adjusted as well to be reusable on both sides.

I'm creating a `ManagedItems` component that takes an array of items (nodes), a property definition object and a string prefix for the local storage key names (e.g. `nodes_`) and encapsulates the following functionality:
* Visible columns
* Property filters
* Grouping
* Table page size
* Item selection
* Persistence of settings

It will have absolutely no hard-coded dependency on Z-Wave nodes, so it may be used as well for other item collections (e.g. hass devices, ...) if it makes sense.

This would also allow us to view the table and the mesh graph just as a different visual representation of the same things and keep the filters, selection, ... on both sides in sync (in case we'd want to have it that way) or keep them completely separated.

Of course everything will be quality checked using unit tests :-).

robertsLando:
I think that would be awesome! Nice idea @ahochsteger ! Btw I think we could merge this for now and keep that for another PR? At least we can get some feedbacks about the new implementation :)

ahochsteger:
> I think that would be awesome! Nice idea @ahochsteger ! Btw I think we could merge this for now and keep that for another PR? At least we can get some feedbacks about the new implementation :)

Yes, let's merge it and I'll start a new PR.

PR #462, opened 2021-02-04
feat: add node debug info tab

ahochsteger:
This PR provides a new node tab to display node debug infos - a JSON representation of the node object with all available details including the possibility to copy the code into the clipboard (e.g. for bug reports):
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,6 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.12.1",
-    "@mdi/font": "^5.9.55",
     "@zwave-js/server": "^1.0.0-beta.4",
     "ansi_up": "^5.0.0",
     "app-root-path": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.12.1",
+    "@mdi/font": "^5.9.55",
     "@zwave-js/server": "^1.0.0-beta.4",
     "ansi_up": "^5.0.0",
     "app-root-path": "^3.0.0",
--- a/src/components/nodes-table/ExpandedNode.vue
+++ b/src/components/nodes-table/ExpandedNode.vue
@@ -4,6 +4,8 @@
       <v-tab key="node">Node</v-tab>
       <v-tab v-if="showHass" key="homeassistant">Home Assistant</v-tab>
       <v-tab key="groups">Groups</v-tab>
+      <v-tab key="debug">Debug Info</v-tab>
+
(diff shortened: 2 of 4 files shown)


ahochsteger:
@robertsLando I've added the possibility to use MDI icons, since I've always had troubles finding good icons.
Now everybody can use <https://materialdesignicons.com/> to find good icons and use them within zwavejs2mqtt.

package.json
```diff
@@ -91,6 +91,7 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.12.1",
+    "@mdi/font": "^5.9.55",
```
robertsLando, line 94: @ahochsteger Why this? It makes the application much heavier, couldn't you use the https://material.io/resources/icons/?search=content&icon=content_copy&style=baseline ?

src/components/nodes-table/ExpandedNode.vue
```diff
@@ -31,8 +31,18 @@
...
           <association-groups :node="node" :nodes="nodes" :socket="socket" />
         </v-tab-item>
+
+        <!-- TAB DEBUG INFO -->
         <v-tab-item key="debug">
-          <pre>{{ nodeJson }}</pre>
+          <v-textarea
+            class="mx-2"
+            rows="10"
+            append-icon="mdi-content-copy"
```
robertsLando, line 40:
```suggestion
            append-icon="content_copy"
```

src/plugins/vuetify.js
```diff
@@ -4,6 +4,7 @@ import Vue from 'vue'
 import Vuetify from 'vuetify'
 import 'vuetify/dist/vuetify.min.css'
 import 'material-design-icons-iconfont/dist/material-design-icons.css' // Ensure you are using css-loader
+import '@mdi/font/css/materialdesignicons.css' // Ensure you are using css-loader
```
robertsLando, line 7: dito

robertsLando (changes requested):

package.json
ahochsteger: You are right, that it is not required for the copy icon - I just had troubles in the past to find good icons (e.g. for formatting columns) and found out how to use better icons while working at it.
I'll remove it, since it has nothing to do with this PR.

ahochsteger: done.

src/plugins/vuetify.js
ahochsteger: done.

ahochsteger:
@robertsLando PR is ready.

robertsLando (approved):
LGTM. Thanks! @ahochsteger

PR #479, opened 2021-02-06
docs: fix env-vars.md

jmgiaever:
Add missing env. vars and rename `host` and `port` variables.


docs/guide/env-vars.md
```diff
@@ -3,6 +3,9 @@
 This is the list of the actually supported env vars:
 
 - `NETWORK_KEY`: Zwave Network key
-- `PORT`: The port to listen to for incoming requests. Default is `8091`
```
robertsLando, line 6: That's not correct, this are used in the backend

robertsLando (changes requested):

docs/guide/env-vars.md
jmgiaever: Ah, sorry, I though I did a search for it in the code, but guess I missed that. Maybe you can explain what's the difference between SERVER_HOST/-PORT and just the HOST/PORT options in the text.

docs/guide/env-vars.md
```diff
@@ -3,6 +3,9 @@
...
-- `HOST`: The host address to bind to. Default is `0.0.0.0`
 - `STORE_DIR`: The absolute path to the directory where all files will be stored. Default is `<path to your z2m dir>/store`
+- `SERVER_PORT`: The port to listen to for incoming requests. Default is `8091`
+- `SERVER_SSL`: Set to true to use HTTPS / WSS scheme
+- `SERVER_HOST`: The host address to bind to. Default is `0.0.0.0
```
robertsLando, line 9:
```suggestion
- `SERVER_HOST`: The host address to bind to. Default is `0.0.0.0`
```

docs/guide/env-vars.md
robertsLando: `SERVER_*` env vars are only used when building using webpack, `PORT/HOST/STORE_DIR` are used from backend when running the application. Could you fix docs according to this please?

robertsLando (changes requested):

docs/guide/env-vars.md
jmgiaever: Hm, ok. Not sure if I follow. If they're only used for development, they probably shouldn't be here? Variables in this file should probably only list those that is modifiable on startup, and the others should be explained in the dev docs?

robertsLando: It's for the frontend, when releasing z2m webpack builds the frontend files in dist folders, those env vars are used in this process, I think them should be documented but not along with the others like it's now, you should split them in another list and specify that are only used in webpack build process

robertsLando:
I have documented them in my other pr #535

Issue #472, opened 2021-02-05
[feat]Option to turn off MQTT server


Chaotic:
**Is your feature request related to a problem? Please describe.**
A way to shut off the MQTT server in the docker while using the websocket

**Describe the solution you'd like**
I'd like a way to shut off the MQTT portion of the gateway while still being able to use the websocket gateway.  It seems there is a way to shut off the websocket gateway and just use the MQTT one but not a way to do the reverse.  I can just leave it pointing to localhost but that continues to attempt to connect and refuses filling the logs with unneeded errors.

**Describe alternatives you've considered**
None

**Additional context**
Using the new Homeassistant integration and using websockets and not MQTT the MQTT gateway is redundant in my setup

towerhand:
Try the disable gateway toggle.

Chaotic:
That also hides all the options for the websocket gateway so I assumed that shut that off as well.

I just quickly tested it and it does seem to shut off the websocket as well as MQTT.

FreelancerJay:
I've been running with "Disable Gateway" ticked, and my Home Assistant instance connecting to it via WebSocket:
![Screen Shot 2021-02-06 at 5 14 15 pm](https://user-images.githubusercontent.com/67540736/107110812-03eec300-689f-11eb-98a3-0cbd68b9d801.png)

robertsLando:
Just toggle the disable gateway flag, that's the way to do it. If you are using docker use the dev tag instead of the latest

PR #482, opened 2021-02-06
fix(ui): settings labels

towerhand:
Issue #472 

Temp "fix"

I'm thinking the settings panel needs to be rearranged a little.

This setting for example should be enable to enable instead of enable to disable. 

Also, I'm thinking the settings should be:
- General
- Zwave
- Enable WS server. (instead of inside of Zwave) (show the relevant options if toggle on)
- Enable MQTT Gateway. (show the relevant options if if toggle on) (opposite behavior from current)

Seems like an easy change so if you like the idea I'll figure out how to do it and submit another PR.

### DIFF
--- a/src/components/Settings.vue
+++ b/src/components/Settings.vue
@@ -195,9 +195,9 @@
 
             <v-container cols="12" sm="6" class="ml-1">
               <v-switch
-                hint="Enable this to use Z2M only as Control Panel"
+                hint="Enable this to use zwavejs2mqtt only as Control Panel"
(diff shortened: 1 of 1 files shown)


jarrettv (approved):
Agree we need to revamp settings but this is a good tweak for now

robertsLando:
@towerhand Could you show me your idea?

towerhand:
If you would like let's merge this, I have to figure out how to create a mock up to show you and make the changes I'm thinking about.

PR #495, opened 2021-02-07
refactor: extract items management into its own module

ahochsteger:
This PR extracts the complete management of items (nodes in most cases, but is not limited to just nodes) into its own module `ManagedItems`.
It lays the foundation to be reused in the new mesh graph to enable the same filter functionality.


.babelrc
```diff
@@ -3,7 +3,6 @@
     [
       "@babel/preset-env",
       {
-        "modules": false,
```
ahochsteger, line 6: @robertsLando I had to remove this line to make the (existing and new) tests in `src/modules` work using ES6 `import` instead of `require`.
Do you know, if this may have some unintended consequences?

package.json
```diff
@@ -53,7 +53,7 @@
     "lint-fix:markdownlint": "markdownlint --fix '**/*.md'",
     "lint:prettier": "prettier-standard --check",
     "lint-fix:prettier": "prettier-standard --format",
-    "test": "mocha",
+    "test": "mocha -r @babel/register test src/modules",
```
ahochsteger, line 56: @robertsLando I had to change this to run (existing and new) the tests in `src/modules` but I'm not super happy about this. Maybe we should talk about how want the tests to be written and where we want them to be located.
This should also take into account (future) testing of Vue components.

src/components/nodes-table/nodes-table.js
```diff
@@ -14,40 +13,42 @@ export default {
...
+          productLabel: { type: 'string', label: 'Product code' },
+          name: { type: 'string', label: 'Name' },
+          loc: { type: 'string', label: 'Location' },
+          isSecure: { type: 'boolean', label: 'Secure' },
+          isBeaming: { type: 'boolean', label: 'Beaming' },
+          failed: { type: 'boolean', label: 'Failed' },
+          status: { type: 'string', label: 'Status' },
+          interviewStage: { type: 'string', label: 'Interview stage' },
+          lastActive: { type: 'date', label: 'Last Active', groupable: false }
+        },
```
ahochsteger, line 34: @robertsLando this replaces the `headers` array and allows us to describe the properties using meta data that fits our needs without continuing to misuse the `headers` array, which is then created on-the-fly using `getTableHeaders()`.
In the future, when we are using TypeScript, we may put this information to decorators of the Z2MNode class and read it from there.

src/modules/NodeCollection.js
```diff
@@ -123,19 +123,6 @@ export class NodeCollection {
...
-      if (node[property]) {
-        const strVal = this._strValue(node[property])
-        uniqueMap[strVal] = uniqueMap[strVal] || node[property]
-      }
-    })
-    return Object.keys(uniqueMap)
-      .sort()
-      .map(key => uniqueMap[key])
-  }
 }
```
ahochsteger, line 126: This has been moved to `ManagedItems.getPropValues(propName)`.

ahochsteger:
@robertsLando from my PoV this PR is ready to be merged as it just covers the refactoring into a reusable component that is now used from the nodes table.
I'd continue the work to integrate this functionality into the mesh graph when both PRs (this and #458) have been merged.

Nevertheless there are some questions (see code notes) where I'd like to get some feedback.

In addition I wonder, if the managed nodes may be moved to the Vuex store if we not only want to re-use the functionality but also share the filter settings (or even grouping) between the nodes table and the new mesh graph. But I'd leave that for another PR.

.babelrc
robertsLando: Sincerly I don't know. I have checked it here but not 100% sure https://babeljs.io/docs/en/babel-preset-env#modules

package.json
robertsLando: @ahochsteger I would prefer to have all tests all in one folder, we could create two separete scripts to run them like `test:ui` and `test:server` and use `npm-run-all` to run them when running `npm run test`

src/components/nodes-table/index.vue
```diff
@@ -1,20 +1,16 @@
 <template>
   <v-data-table
-    :headers="activeHeaders"
-    :items="tableNodes"
+    :headers="managedNodes.getTableHeaders()"
+    :items="managedNodes.getFilteredItems()"
```
robertsLando, line 4: I'm not 100% sure about this. Is this reacting to changes? AFAIK it's not a good practise to use a function that returns a value in this places

robertsLando (changes requested):

src/components/nodes-table/index.vue
ahochsteger: Hmm, you may be right -  I'll have to think about it more deeply ...

robertsLando: LMK

ahochsteger: I assume, just using the properties without getter and setter functions would solve that problem.
But then I wouldn't easily recognize changes (which is done in the `set*` functions) to automatically update the local storage.
Can Vuex help to solve this problem?
AFAIK getters and mutatations are splitted there as well.

robertsLando: > I assume, just using the properties without getter and setter functions would solve that problem.
But then I wouldn't easily recognize changes (which is done in the set* functions) to automatically update the local storage.

I think that the correct way to do this is using the get/set methods in the class. You can use the set to detect changes

https://stackoverflow.com/questions/12827266/get-and-set-in-typescript

.babelrc
ahochsteger: Yes, that's what I thought too - removing it defaults to "auto" which seems to be the recommended way in combination with babel.

src/components/nodes-table/index.vue
ahochsteger: > I think that the correct way to do this is using the get/set methods in the class. You can use the set to detect changes
> 
> https://stackoverflow.com/questions/12827266/get-and-set-in-typescript

LGTM, thanks for the hint! :-)

package.json
ahochsteger: Yes, that makes sense, I'll move the tests from `src/modules` to `test/modules`.
So should we move the code from within `src/modules` to `libs` as well?
It's not really clear for me what goes into `lib` and what into `src/modules`.
Seem to serve the same purpose for me, or not?

robertsLando: I would rename `src/modules` to `src/lb` and place there all the `js/ts` files that can be reused in the code. 

For components in general I prefer to create a folder for each one (with the name of the component) and inside that insert all component related files (.vue, .js .css) like you did for nodes table

For tests I would keep them all inside `test` folder (from root dir) and split frontend and backend tests there in 2 folders

src/components/nodes-table/index.vue
robertsLando: Thanks to you :)

package.json
ahochsteger: I guess you meant `src/lib` instead of `src/lb`, right?
Does "place there all the js/ts files that can be reused in the code" also include the files within `lib` (from root)?
If not, what's the difference between `lib` and `src/lib`?

src/components/nodes-table/index.vue
ahochsteger: It's done.

package.json
robertsLando: > I guess you meant src/lib instead of src/lb, right?

Oh yep sorry. typo :laughing: 

> Does "place there all the js/ts files that can be reused in the code" also include the files within lib (from root)?

Noipe, just the files used in frontend :) I prefer to keep front and backend splitted

> If not, what's the difference between lib and src/lib?

Think about the root folder as everything that is backend related, everything in `src` is frontend related. Also all imports oof components use `/@/` that is used to point to the `src` folder

ahochsteger:
@robertsLando it's using now the get/set syntax everywhere possible and I've added a lot of tests to cover most of the functionality of `ManagedItems`.
I suggest to do the cleanup of the test locations and other testing-related stuff to a separate PR.
So it's ready to be reviewed again.

package.json
ahochsteger: FYI: your first comment (using commands `test:ui` and `test:server` and `npm-run-all` for `test`) is done.
As I said, I'd do the restructuring of the tests in a separate PR.

robertsLando (approved):
LGTM. GG @ahochsteger ! :)

PR #524, opened 2021-02-09
feat(ui): compact control panel menu

robertsLando:


robertsLando:
Wow, this is awesome! Please @ahochsteger Look at this

robertsLando:
I'm thinking about keep both control panel visualizations, let's call the second one `ControlPanelLite` and add a switch to see one or the other. I think that the lite version is perfect for mobile devices and could be anabled by default on lower sceen size screens

towerhand:
Really impressed, I like this new grid view and appreciate the "group by" option. Some people may prefer the current list view, would it be feasible to have an option to switch back and forth?

ahochsteger:
@jarrettv wow I'm totally impressed by this beautiful looking UI - very good work!

@robertsLando seems like I have to reconsider some of my plans, but I think the refactoring of `ManagedItems` that has been merged today may also be used to provide the node data with filtering, grouping and sorting functionality for this UI.

Since both UIs seem to have different strengths I see no problem in providing both UIs (until we may combine these into one if possible), ideally backed by `ManagedItems` to provide cross-UI functionality (filtering, grouping, sorting, ...), which may also be used by the new mesh graph.

What do you think?

robertsLando:
Yep I think the same, I think that some users could find the table more clear others could prefer @jarrettv, if we allow to switch from one to the other esily (and store the preference in settings like for the table settings etc) it would be awesome! 

I think Jarret version could be the default one on mobiles

ahochsteger:
> I think Jarret version could be the default one on mobiles

Yes, that sounds good.

We may see them as the **"simple UI"** with most every-day features and the **"advanced UI"** with all available features.
I can also imagine the simple UI by default for new users in the future - at least when all important features are there.

robertsLando:
> We may see them as the "simple UI" with most every-day features and the "advanced UI" with all available features.
I can also imagine the simple UI by default for new users in the future - at least when all important features are there.

I agree!

PR #547, opened 2021-02-10
fix: generate the action template when an action map is given

chilicheech:
fix: generate action template when an action map exists

This will make zjs2m automatically generate the `action_template` when an `action_map` is given. This is the same logic that is used for the `fan_mode_state_template` and for the `mode_state_template`.

This will make the action be reported correctly in hass.

References:
https://www.home-assistant.io/integrations/climate.mqtt/#action_template
https://www.home-assistant.io/integrations/climate.mqtt/#action_topic


robertsLando (commented):
@chilicheech Could you also create the logic to generate the action template on discoverClimate method?

chilicheech:
@robertsLando here's an idea. I tested both with my thermostat in the devices.js and without.. both seem to detect the action properly.

robertsLando (approved):
From a code side it seems good. Did you also tried to remove the device from `devices.js` file to see if the discoverClimate function detects it correctly?

chilicheech:
I tried both with the device in devices.js and without the device in
devices.js and they both worked correctly 😁

robertsLando:
@chilicheech Ok so if it works even without it in devices.js, could you remove that from there? I would like to completely remove all devices from that file :)

chilicheech:
so, @robertsLando, the action it is working properly, but there are other things that aren't working properly yet, like min temp, max temp.. temp_step..

i don't think we should remove the device from devices.js yet. i think we should go ahead and merge this PR as it gets us a step closer to a working auto discovery of climate devices.. but there's more work to be done.. i can try to look into the other issues.. but i think we should keep them as separate PRs, and keep this one as just the action_template stuff :)

robertsLando:
Ok perfect, thanks :)

PR #549, opened 2021-02-11
chore: jsonstore refactor and speed up docker builds

chrisns:
copied from convo in slack:

ok, so i've spent a few hours picking through this. the biggest time sink is on compiling serialport along with needing the dependencies for doing that
however serialport do helpfully provide some precompiled static binaries so we don't actually need to compile it

```
$ docker run --rm -ti --entrypoint=""  node:15.8.0-buster-slim bash
root@e088b102d582:/# mkdir /foo; cd foo
root@e088b102d582:/foo# time npm i --save serialport
added 76 packages, and audited 77 packages in 5s
5 packages are looking for funding
  run `npm fund` for details
found 0 vulnerabilities
npm notice
npm notice New patch version of npm available! 7.5.1 -> 7.5.3
npm notice Changelog: https://github.com/npm/cli/releases/tag/v7.5.3
npm notice Run npm install -g npm@7.5.3 to update!
npm notice
real    0m4.933s
user    0m4.157s
sys     0m1.194s
root@e088b102d582:/foo# du -hs node_modules/
4.0M    node_modules/
```

note the static binaries *don't work with alpine* but that's ok, the `slim` image is only a few mb bigger

we could also optimise the package.json so we only pull the things we need for a docker build, e.g. no need for `husky`, `nodemon` etc.. perhaps we could split out into `optionalDependencies`


ahochsteger:
@chrisns good approach to further optimize the build - I like it!

To summarize it looks like we have the following options to optimize the build time:
* Reduce build dependencies (esp. prevent build of `serialport` and use `--no-optional`)
* Move webpack build out of docker and reuse build results for every arch
* Cache results of `npm install` and `npm install --no-optional --only=prod` in reusable base images and only build them, if dependencies change

It seems that you removed the multi stage build from the Dockerfile, was that intended, since dev dependencies would still be part of the image size (in upper layers)?

Currently the build results in an error:
```
buildx call failed with: error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: no match for platform in manifest sha256:ac02538954112b4612052d88fbd2e8c5353bef2ba6848cd5b97d640f3ee052fd: not found
```

chrisns:
Yeah, aware I broke the build. It got to 1:30am and I ran out of steam

I'd suggest:

- multi step builds are fine if we cache all steps
- separate stuff not needed to build into optional deps
- copy the package.json and lock, npm install --no-optional
- device update
- then copy . .
- npm build
- npm prune --production
- next step
- copy things in

At the minute you need to recompile everything on any change, separating the copy will help with that

You could maybe feature gate copying in an externally built webpack, but I'd like to keep it happening for the local dev experience

If someone else cares they can take over, I won't get a chance to do anymore for a while

robertsLando:
> You could maybe feature gate copying in an externally built webpack, but I'd like to keep it happening for the local dev experience

Why this? If a dev needs webpack could just run an npm install inside docker

chrisns:
Being able to clone and docker build to get everything running is 'nice' without trying to understand our crazy

robertsLando:
OK we could still install dev deps too but making the build outside anyway, nope?

.github/workflows/dockerbuild.yml
```diff
@@ -45,6 +45,8 @@ jobs:
         with:
           platforms: linux/arm64,linux/amd64,linux/arm/v6,linux/arm/v7
           file: docker/Dockerfile
+          cache-from: type=registry,ref=zwavejs/zwavejs2mqtt:dev
+          cache-to: type=inline
```
robertsLando, line 49: Shouldn't this also have the  `mode=max`

docker/Dockerfile
```diff
@@ -1,51 +1,22 @@
-# ----------------
-# STEP 1:
-FROM node:erbium-alpine AS build-z2m
+FROM node:12.20.1-buster
+# removing the dependency on fibers would mean we could go for an image without the compilers+python, also allow us to get beyond node 14
```
robertsLando, line 2: It is needed in the webpack build

docker/Dockerfile
```diff
@@ -1,51 +1,22 @@
-# ----------------
-# STEP 1:
-FROM node:erbium-alpine AS build-z2m
+FROM node:12.20.1-buster
```
robertsLando, line 1:
```suggestion
FROM node:12.20.1-buster AS build-z2m
```

BTW it would prefer to stick to alpine if possibile

chrisns:
Right now most your time is sunk in npm install, specifically on compiling binaries and there’s no cache
I’d optimise that first before looking into external build which I see as additional not instead of

I have some thoughts on how to optimise that like how you're thinking but I'm not yet sure it's necessary and worth the extra complexity to shave seconds

robertsLando (changes requested):

.github/workflows/dockerbuild.yml
chrisns: Yes. If we keep multi step builds

docker/Dockerfile
chrisns: Yeah. It's an obsolete module and a bit ugly. It is possible apparently to remove though may incur performance hit

docker/Dockerfile
chrisns: Yeah. Though you won't get the static binaries. Not the end of the world if the steps are optimised and cached

ahochsteger:
@chrisns are the caching changes yet supposed to reduce the build time or is there some more work left to do?
I'm asking because the last build still took about 33 minutes.

For some reason the GitHub actions are stuck in status "Expected - Waiting for status to be reported" now thus my latest commit (fix lint errors) doesn't build.
I did some research and found this:
<https://github.community/t/expected-waiting-for-status-to-be-reported/18001/2>

Can it be that some steps are (intentionally) deactivated but still marked as required and thus expected to be run

FYI: @robertsLando

ahochsteger:
@chrisns, @robertsLando I'll try some experiments on the multi-staged docker builds with published intermediate base images and see if this works out. I suggest to create a separate PR so we don't step on each other.

Chris, you mentioned that you'd like to keep a simple build for people that just want to checkout and build the image. I agree this is what we should aim for and I have some ideas to support both: an optimized GitHub build while still keep it simple for first-time users.

Since there's already a `docker-compose.yml` file in the repo:
Would a docker-compose based solution for first-time-users be ok or do we want a provide pure docker build?
I always find it easier to use docker-compose since you can put the image name and build arguments into the file and don't have to tell the user how to correctly invoke `docker build`.

robertsLando:
I would firstly merge this if ready and then try other approachs.

robertsLando (dismissed):

Issue #381, opened 2021-01-29
[feat] RGBWWCW support


DanielWinks:
**Is your feature request related to a problem? Please describe.**

There doesn't seem to be a way to set (or read) the color temp for RGBWWCW bulbs when using Home Assistant. In the control panel, I see a place where I can set the 0-255 WW color and the 0-255 CW color. This lets me manually set it to any combination I want, but only from the admin panel. In Home Assistant, it expects that white value is a single MQTT topic. There's no way to configure an MQTT light in HA that uses separate MQTT topics for WW and CW.

In Zwave2MQTT it was set with a HEX string, with the first 3 numbers all 00, with a template like: `"{{ '#%02x%02x%02x%02x%02x' | format(0, 0, 0, (0.7349 * ((value|int) - 153)) | round(0), 255 - (0.7349 * ((value|int) - 153)) | round(0))}}"`

In the new zwavejs, there's a spot where I can submit a RRGGBB number, but nowhere I can submit an RRGGBBWWCW like Zwave2MQTT allowed.

This mean is I can't create an MQTT light in Home Assistant that allows me to control the white temps.

**Describe the solution you'd like**
Add a `hexWhite` topic to accompany the new `hexColor`. This would be better than what existed in zwave2MQTT, since the templates for MQTT in HASS would be simpler: 

`{{'%02x%02x%02x' | format(red, green, blue)}}` and `"{{ '%02x%02x' | format((0.7349 * ((value|int) - 153)) | round(0), 255 - (0.7349 * ((value|int) - 153)) | round(0))}}"` 

instead of what existed in zwave2mqtt (where you needed to pass in `#000000` in front of your WWCW value.

**Describe alternatives you've considered**
Add a full hex topic like zwave2mqtt has.

robertsLando:
@AlCalzone needs a new property along with the hexColor?

AlCalzone:
Imo this rather sounds like a workaround for a HA shortcoming which should be addressed there.
I've asked the HA devs, let's wait for their response.

marcelveldt:
I believe he's talking about MQTT to HA and not the new Zwave JS integration in HA.
In the integration we combine all values into RGB-CT/RGBWWCT

AlCalzone:
alternate solution as discussed on Discord:
https://github.com/zwave-js/node-zwave-js/issues/1526

marcelveldt:
Great, both the HA integration as others (like MQTT) can benefit from that.

[...]

PR #568, opened 2021-02-12
feat: correctly parse `currentColor` value and improve rgb discovery

### LINKED ISSUE #477: [bug] Zipato v2 rgb not showing/working colors
The bug was already described here: https://github.com/zwave-js/zwavejs2mqtt/issues/351, but the issue was closed.
The same bug was on old integration on zwave 1.4 in HA and in zwave2mqtt.
I have contacted few times Zipato support to get firmware update, but they dont respond.
### Version

Build/Run method

- [ x] Docker
- [ ] PKG

zwavejs2mqtt version: 1.0.5
zwavejs version: 6.1.3

Testing with Ha addon and on extra rpi in docker.

### Describe the bug

Dont show and dont work color and temperature picker.
Zaslonska slika 2021-01-31 10-26-55
Zaslonska slika 2021-01-31 10-28-12

### To Reproduce

Try change color or temperature on Zipato v2 rgb light.

### Expected behavior

To have working color and temperature picker.

### Additional context

Logs: 
[zwavejs2mqtt-store.6.zip](https://github.com/zwave-js/zwavejs2mqtt/files/5936450/zwavejs2mqtt-store.6.zip)

I have tried to add this config on another rpi, picker is showing but still not working:
```
Also, as a short-term workaround:

Shut down zwavejs2mqtt
open <your-home-id>.values.json, add these two lines at the end:
{"k":"{\"nodeId\":24,\"commandClass\":51,\"endpoint\":0,\"property\":\"supportedColorComponents\"}","v":[0,1,2,3,4]}
{"k":"{\"nodeId\":24,\"commandClass\":51,\"endpoint\":0,\"property\":\"supportsHexColor\"}","v":true}
Start zwavejs2mqtt.
Leave your hands off the refreshInfo button!
```

After this config:
```
2021-02-04 07:38:08.715 INFO ZWAVE: Controller status: Driver: Color Switch: property "targetColor" requires a property key, but none was given
2021-02-04 07:38:08.721 INFO ZWAVE: Controller status: Driver: Color Switch: property "targetColor" requires a property key, but none was given
2021-02-04 07:38:08.727 INFO ZWAVE: Controller status: Driver: Color Switch: property "targetColor" requires a property key, but none was given```

### PR DESCRIPTION
**Pull request by billiaz**
Z-wave gives posibility to manage RGB WW and CW on devices.
Each device has these channels separately.

Currently all devices has 5 channels or less (RGBWWCW, RGBW, RGB)

As devices (bulbs) manage RGB, WW and CW separately. there is no easy way to bind this with HASS
Hass light entity supports:
1. On/OFF
2. Dimmer
3. RGB
4. White Temperature
5. HS, WX and some other features

the split WW/CW do not allow us to us the same entity for RGB and for Whites.

This patch implements:

- RGB with brightness OR switch (depends on what exists)
- Dimmer for both Whites (Cold and Warm)

Some devices support change of Warm white or cold white through Configuration Param. I will review if this is feseable to also supported.

fixes #477 #381  #351 #595

### CODE REVIEW

**Comment by robertsLando** [comment]
@billiaz Works?

**Comment by billiaz** [comment]
Not fully tested, I am waiting for a Bulb to arrive! lets keep it as "draft" for now"

**Comment by billiaz** [comment]
This has to wait the Priority CC change, before it can be updated/used #625

File: lib/Gateway.js [hunk 1]
```diff
@@ -786,6 +807,77 @@ function addRgbDimmer (cfg, node, valueId) {
...
+  cfg.discovery_payload.rgb_state_topic = this.mqtt.getTopic(
+    this.valueTopic(node, valueId)
+  )
+  cfg.discovery_payload.rgb_command_topic =
+    cfg.discovery_payload.rgb_state_topic + '/set'
+
+  // The following part of code, checks if ML or Binary works. If one exists the other
+  let brightnessValue = false
+  let switchValue = false
+  if (node.values.hasOwnProperty(`38-${valueId.endpoint}-currentValue`)) {
```
robertsLando, line 823: don't use hasOwnProperty here

lib/Gateway.js
```diff
@@ -786,6 +807,77 @@ function addRgbDimmer (cfg, node, valueId) {
...
+    brightnessValue = `38-${valueId.endpoint}-currentValue`
+  } else if (
+    node.values.hasOwnProperty(`38-${valueId.endpoint + 1}-currentValue`)
```
robertsLando, line 826: same

lib/Gateway.js
```diff
@@ -786,6 +807,77 @@ function addRgbDimmer (cfg, node, valueId) {
...
+  ) {
+    brightnessValue = `38-${valueId.endpoint + 1}-currentValue`
+  } else if (
+    node.values.hasOwnProperty(`37-${valueId.endpoint}-currentValue`)
```
robertsLando, line 830: same

lib/Gateway.js
```diff
@@ -786,6 +807,77 @@ function addRgbDimmer (cfg, node, valueId) {
...
+    cfg.discovery_payload.brightness_command_topic = this.mqtt.getTopic(
+      this.valueTopic(node, node.values[brightnessValue], true).targetTopic
+    ) + '/set'
+    cfg.discovery_payload.state_topic = false
+  } else if (switchValue) {
+    cfg.discovery_payload.state_topic = this.mqtt.getTopic(
+      this.valueTopic(node, node.values[brightnessValue])
+    )
+    cfg.discovery_payload.command_topic = this.mqtt.getTopic(
+      this.valueTopic(node, node.values[brightnessValue], true).targetTopic
```
robertsLando, line 852: use a var here as you are using the same value in the previous step

lib/Gateway.js
```diff
@@ -786,6 +807,77 @@ function addRgbDimmer (cfg, node, valueId) {
   }
 
   return false
+} */
+
+function addRgbColorSwitch (cfg, node, valueId) {
+  cfg = copy(cfg)
```
robertsLando, line 813: you are coping a switch config, you need to change the discovery payload to match this to light entity instead

robertsLando (changes requested):

lib/Gateway.js
billiaz: after slack chat Thanks for help and explanation!

lib/Gateway.js
billiaz: refactored it to use a var

DanielWinks:
This does not fix #381. It introduces major regressions in generated HASS discovery, creating 3 unnecessary extra entities:

<details>
<summary>light_rgbw_bulb</summary>

````
{
  "type": "light",
  "object_id": "rgbw_bulb",
  "values": [
    "38-0-currentValue",
    "38-0-targetValue",
    "51-1-0"
  ],
  "discovery_payload": {
    "state_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/currentValue",
    "command_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/targetValue/set",
    "on_command_type": "brightness",
    "brightness_state_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/currentValue",
    "brightness_command_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/targetValue/set",
    "state_value_template": "{{ \"OFF\" if value == 0 else \"ON\" }}",
    "brightness_value_template": "{{ (value) | round(0) }}",
    "brightness_scale": "99",
    "color_temp_state_topic": "51-1-0",
    "color_temp_command_template": "{{ '#%02x%02x%02x%02x%02x' | format(0, 0, 0, (0.7349 * (value - 153)) | round(0), 255 - (0.7349 * (value - 153)) | round(0))}}",
    "color_temp_command_topic": "51-1-0",
    "color_temp_value_template": "{{ (((value[7:9] | int(0, 16)) / 0.7349 ) | round(0)) + 153 }}",
    "rgb_command_template": "{{'#%02x%02x%02x%02x%02x' | format(red, green, blue,0,0)}}",
    "rgb_command_topic": "51-1-0",
    "rgb_state_topic": "51-1-0",
    "rgb_value_template": "{{ value[1:3] | int(0, 16) }},{{ value[3:5] | int(0, 16) }},{{ value[5:7] | int(0, 16) }}",
    "device": {
      "identifiers": [
        "zwavejs2mqtt_0xebd8bb21_node20"
      ],
      "manufacturer": "Inovelli",
      "model": "Multi-Color Bulb (LZW42)",
      "name": "family_room_floor_lamp_rgbw1",
      "sw_version": "2.30"
    },
    "name": "family_room_floor_lamp_rgbw1_rgbw_bulb",
    "unique_id": "zwavejs2mqtt_0xebd8bb21_Node20_rgbw_bulb"
  },
  "discoveryTopic": "light/family_room_floor_lamp_rgbw1/rgbw_bulb/config",
  "persistent": false,
  "ignoreDiscovery": false,
  "id": "light_rgbw_bulb"
}
````

</details>

<details>
<summary>light_rgb_dimmer</summary>

````
{
  "type": "light",
  "object_id": "rgb_dimmer",
  "discovery_payload": {
    "state_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/currentValue",
    "command_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/targetValue/set",
    "brightness_state_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/currentValue",
    "brightness_command_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/targetValue/set",
    "on_command_type": "brightness",
    "state_value_template": "{{ \"OFF\" if value == 0 else \"ON\" }}",
    "brightness_value_template": "{{ value }}",
    "brightness_scale": 99,
    "rgb_command_template": "{{ \"#%02x%02x%02x\" | format(red, green, blue)}}",
    "rgb_command_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/hexColor/set",
    "rgb_state_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/hexColor",
    "rgb_value_template": "{{ value[0:2] | int(0, 16) }},{{ value[2:4] | int(0, 16) }},{{ value[4:6] | int(0, 16) }}",
    "device": {
      "identifiers": [
        "zwavejs2mqtt_0xebd8bb21_node20"
      ],
      "manufacturer": "Inovelli",
      "model": "Multi-Color Bulb (LZW42)",
      "name": "family_room_floor_lamp_rgbw1",
      "sw_version": "2.30"
    },
    "name": "family_room_floor_lamp_rgbw1_rgb_dimmer",
    "unique_id": "zwavejs2mqtt_0xebd8bb21_20-51-0-hexColor"
  },
  "discoveryTopic": "light/family_room_floor_lamp_rgbw1/rgb_dimmer/config",
  "values": [
    "51-0-hexColor"
  ],
  "persistent": false,
  "ignoreDiscovery": false,
  "id": "light_rgb_dimmer"
}
````

</details>

<details>
<summary>light_dimmer</summary>

````
{
  "type": "light",
  "object_id": "dimmer",
  "discovery_payload": {
    "command_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/targetColor/1/set",
    "state_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/currentColor/1",
    "state_value_template": "{{ \"OFF\" if value == 0 else \"ON\" }}",
    "brightness_command_topic": true,
    "brightness_scale": 255,
    "brightness_state_topic": true,
    "brightness_value_template": "{{ value }}",
    "on_command_type": "brightness",
    "device": {
      "identifiers": [
        "zwavejs2mqtt_0xebd8bb21_node20"
      ],
      "manufacturer": "Inovelli",
      "model": "Multi-Color Bulb (LZW42)",
      "name": "family_room_floor_lamp_rgbw1",
      "sw_version": "2.30"
    },
    "name": "family_room_floor_lamp_rgbw1_dimmer",
    "unique_id": "zwavejs2mqtt_0xebd8bb21_20-51-0-currentColor-1"
  },
  "discoveryTopic": "light/family_room_floor_lamp_rgbw1/dimmer/config",
  "values": [
    "51-0-currentColor-1",
    "51-0-targetColor-1"
  ],
  "persistent": false,
  "ignoreDiscovery": false,
  "id": "light_dimmer"
}
````

</details>

<details>
<summary>light_dimmer_0</summary>

````
{
  "type": "light",
  "object_id": "dimmer_0",
  "discovery_payload": {
    "command_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/targetColor/set",
    "state_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/currentColor",
    "state_value_template": "{{ \"OFF\" if value == 0 else \"ON\" }}",
    "brightness_command_topic": true,
    "brightness_scale": 255,
    "brightness_state_topic": true,
    "brightness_value_template": "{{ value }}",
    "on_command_type": "brightness",
    "device": {
      "identifiers": [
        "zwavejs2mqtt_0xebd8bb21_node20"
      ],
      "manufacturer": "Inovelli",
      "model": "Multi-Color Bulb (LZW42)",
      "name": "family_room_floor_lamp_rgbw1",
      "sw_version": "2.30"
    },
    "name": "family_room_floor_lamp_rgbw1_dimmer_0",
    "unique_id": "zwavejs2mqtt_0xebd8bb21_20-51-0-currentColor-0"
  },
  "discoveryTopic": "light/family_room_floor_lamp_rgbw1/dimmer_0/config",
  "values": [
    "51-0-currentColor-0",
    "51-0-targetColor-0"
  ],
  "persistent": false,
  "ignoreDiscovery": false,
  "id": "light_dimmer_0"
}
````

</details>

What is needed to fix #381 is adding a hex-compatible topic, such as `51-0-hexWhite`, where HASS can send WWCW hex payloads. The `light_rgb_dimmer` discovery, as well as `light_dimmer` and `light_dimmer_0` should not be created, they are not needed and just clutter HA. With a hexWhite topic available, the HASS discovery should just be something like this:

<details>
<summary>light_dimmer_0</summary>

````
{
  "type": "light",
  "object_id": "rgbw_bulb",
  "values": [
    "38-0-currentValue",
    "38-0-targetValue",
    "51-1-0"
  ],
  "discovery_payload": {
    "state_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/currentValue",
    "command_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/targetValue/set",
    "on_command_type": "brightness",
    "brightness_state_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/currentValue",
    "brightness_command_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/targetValue/set",
    "state_value_template": "{{ \"OFF\" if value == 0 else \"ON\" }}",
    "brightness_value_template": "{{ (value) | round(0) }}",
    "brightness_scale": "99",
    "color_temp_state_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/hexWhite",
    "color_temp_command_template": "{{ '%02x%02x' | format((0.7349 * (value - 153)) | round(0), 255 - (0.7349 * (value - 153)) | round(0))}}",
    "color_temp_command_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/hexWhite/set",
    "color_temp_value_template": "{{ (((value[7:9] | int(0, 16)) / 0.7349 ) | round(0)) + 153 }}",
    "rgb_command_template": "{{'%02x%02x%02x' | format(red, green, blue)}}",
    "rgb_command_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/hexColor/set",
    "rgb_state_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/hexColor",
    "rgb_value_template": "{{ value[1:3] | int(0, 16) }},{{ value[3:5] | int(0, 16) }},{{ value[5:7] |
  int(0, 16) }}",
    "device": {
      "identifiers": [
        "zwavejs2mqtt_0xebd8bb21_node20"
      ],
      "manufacturer": "Inovelli",
      "model": "Multi-Color Bulb (LZW42)",
      "name": "family_room_floor_lamp_rgbw1",
      "sw_version": "2.30"
    },
    "name": "family_room_floor_lamp_rgbw1_rgbw_bulb",
    "unique_id": "zwavejs2mqtt_0xebd8bb21_Node20_rgbw_bulb"
  },
  "discoveryTopic": "light/family_room_floor_lamp_rgbw1/rgbw_bulb/config",
  "persistent": false,
  "ignoreDiscovery": false,
  "id": "light_rgbw_bulb"
}
````

</details>

**Comment by scyto** [comment]
This PR has fixed:

- Fibaro RGBWv1 discovery via MQTT - and I now get a color wheel in HA.
- Switch 6 binary switch issue introduced in https://github.com/zwave-js/zwavejs2mqtt/pull/290/files 

This is the fibaro
image

I only have an RGB strip so can't test the warm W channel on this controller (hmm let me see if I can overnight myself a new LED strip)

**Comment by scyto** [comment]
i note that `light_rgb_dimmer` seems to be needed for the fibaro issue

**Comment by billiaz** [comment]
@scyto That is a strange one! I need to test it but I might not be able to, can you send me the node data from this?
It would be nice to test if you get similar mixups from master. to rule out causes

@DanielWinks 
> This does not fix #381. It introduces major regressions in generated HASS discovery, creating 3 unnecessary extra entities:

I would like to understand in the point of view of home assistant how the entity would suppor WW and CW

Currently hass has one White switch which controls, from my understanding white color temperature.
Z-wave is not working as Hue bulbs (at least the majority of bulbs I've seen). Important: I have not seen any Color Switch V3 yet.
This brings the two dimmers. As they are three different dimming channels to control (RGB,WW and CW)

What I see in your discovery is wierd, I would like to see the node export (Will try to find it in topic)

We could have these 3 dimmers under:
WHITE and HS values to include them all together in one payload.

I see no clutter in HASS except the extra hass entry you report. I would be interested to help debug this issue too.

**Comment by billiaz** [comment]
@DanielWinks 
I looked source of Hass. I cannot find hex the way you asking, supported by the hass itself. I want more info. And please point me on source to understand how that works
HEX to RGB: https://github.com/home-assistant/core/blob/788134cbc408cec45e1ff2014d112a8cdaf1683c/homeassistant/components/mqtt/light/schema_basic.py#L338 AND the hex decoder: https://github.com/home-assistant/core/blob/788134cbc408cec45e1ff2014d112a8cdaf1683c/homeassistant/util/color.py#L347

Now. The way you ask us to implement it is not possible, from what I see (even the json is not covering how Z-wave approaches this. Please refer to specification for more details! 

Most devices use WW and CW as separate channels, (channel 4 and 5) this makes it not easy with the current HASS integration, and to avoid confusion... to implement it.
As said, happy to point me on source showing what you suggest really works! Until then. I believe the way is implemented is the least confusing.
Way is:
- RGB entity
- WW dimmer
- CW dimmer

**Comment by DanielWinks** [comment]
> Now. The way you ask us to implement it is not possible, from what I see (even the json is not covering how Z-wave approaches this. Please refer to specification for more details!

It has to be possible, since this is how Zwave2MQTT handled it. These bulbs, under Zwave2MQTT had a single hex topic for both color and whites. The topic in Z2M was `#RRGGBBWWCW`. If you wanted to change the light to 240 R 120 G 0 B, you'd send an MQTT message like `#F078190000` (leaving the WW and CW as 00 and 00). If you wanted some shade of white, say full warm white, you'd send `#000000FF00` (leaving RRGGBB as 000000).

You can see the discovery that worked this way in Z2M in my [post](https://github.com/zwave-js/zwavejs2mqtt/pull/568#issuecomment-782458316) above, for the `light_rgbw_bulb`, as this payload is the old Z2M style.

You don't need to look at the source for HASS, just look at the documentation for setting up MQTT lights. Look at the settings for Color Temp [command](https://www.home-assistant.io/integrations/light.mqtt/#color_temp_command_topic) and [state](https://www.home-assistant.io/integrations/light.mqtt/#color_temp_state_topic). These do NOT have multiple channels/topics. The expect a unified topic for "color temp", not "warm white" and "cold white" separately. Z2M had such a unified topic, as I described above.

> I believe the way is implemented is the least confusing.
Way is:

> RGB entity
> WW dimmer
> CW dimmer

No, that's much worse than how this worked in Z2M, and not the "least confusing" way by any means. My Inovelli Ilumin bulbs are a single bulb, not 3. With Z2M, I get a single color picker and a slider that I can choose color temp with, on a single entity. This is least confusing. If I want a color, I just click somewhere on the color wheel and I get a color. If I want a white temp instead, I just move the slider where I want it and it's now in WWCW mode.

Having what you're describing here is a massive regression in functionality from Z2M and, quite frankly, borderline useless. The bulbs I have will NOT display an RGB color if either WW or CW are anything other than 0. So doing what you're describing here, if I want a color, I need to open the WW entity, set it to 0, close it, open the CW entity, set it to 0, close it, go to the RGB picker and pick my color. Then if I want, say, 3500K, I could not just use the `light.turn_on` service in HA and pass it `kelvin: 3300`, like I could when using Z2M. Instead, I'd need to manually go set the WW slider to like 255 and the CW slider to, I dunno, 50 or something. The `light.turn_on` service would not be able to set a color temp because it wouldn't have a color_temp topic to send a command on. It cannot work with separate entities like this.

All that needs done is creating a topic in ZwaveJS2MQTT for RGBWWCW bulbs that can set both WW and CW with a single payload. It doesn't matter if it's hex or decimal, it could be WWCW hex, or WWWCCC decimal, or JSON, it really doesn't matter. All that matters is that you can single a single payload from HASS to ZwaveJS2MQTT that contains both WW and CW.

This is how Zwave2MQTT worked, so it's absolutely possible in ZwaveJS2MQTT too. Zwave itself doesn't need a combined WWCW, I understand that Zwave itself commands each channel separate. What needs done here is that ZwaveJS2MQTT needs to add a topic, where it can get a payload sent with both WW and CW, then split that and send each part to the WW and CW Zwave commands. And another topic needs created which takes WW and CW from Zwave and puts it on a single topic for getting state.

Just go spin up Zwave2MQTT with an RGBW bulb, and look at the topics it uses. It has a unified topic for RRGGBBWWCW. We just need the same here, or one for RRGGBB like we currently have for color and a separate WWCW one for white.

**Comment by DanielWinks** [comment]
[#1782](https://github.com/zwave-js/node-zwave-js/pull/1782) fixes the issues described here in a much better way. #674 needs merged as well, to resolve the issue described in that PR, which prevents the changes in [#1782](https://github.com/zwave-js/node-zwave-js/pull/1782) from working properly.

With the above 2 PRs merged, I am able to fully control my RGBW bulbs, and due to the nature of how [#1782](https://github.com/zwave-js/node-zwave-js/pull/1782) affects all other multicolor lights, it should also resolve issues with RGBW, as well as any bulbs that have amber, cyan, or purple channels.

Creating needless extra dimmers for WW and CW channels is not needed and this PR should not be merged. This PR will not function properly with Home Assistant MQTT light entities, which expect to be able to control RGB on a single command topic, get RGB state on a single state topic, and get white color temp state on a single topic, and set white color temp on a single topic.

**Comment by scyto** [comment]
There is still a need for some of the items in this PR to fix issues with Fibaro v1 and mis-classification of a binary switch as a color switch.

**Comment by billiaz** [comment]
Abandoned,  fir issues rgbw has please reder to @DanielWinks

**Comment by DanielWinks** [comment]
> Fibaro v1 and mis-classification of a binary switch as a color switch.

This seems like something that should be handled by the device config, rather than at the gateway level.

**Comment by billiaz** [comment]
> > Fibaro v1 and mis-classification of a binary switch as a color switch.
> 
> This seems like something that should be handled by the device config, rather than at the gateway level.

I disagree, as the gateway config is confusing, this pr WAS focusing on also splitting the color switch by other devices.
The previous way of discovering an RGB switch was through multilevel or binary switches, where devices confused there.
Additionally, not all devices provide `ccSpecific` information to help us understand what these switches are.

**Comment by robertsLando** [comment]
What I don't undesrtand is if there is a PR from @DanielWinks somewhere ?

**Comment by DanielWinks** [comment]
@robertsLando This PR was made to address some issues better resolved by [1782](https://github.com/zwave-js/node-zwave-js/pull/1782) in ZwaveJS (which also needs #674 to actually work). I don't have any PRs for these issues myself. I'm just testing [1782](https://github.com/zwave-js/node-zwave-js/pull/1782) which solves the issue this PR was created to try to also solve. However, this PR created some other issues, such as extra entities, and incompatibility with HASS MQTT light (so it wouldn't actually fix things in HASS). [1782](https://github.com/zwave-js/node-zwave-js/pull/1782) does indeed allow for a properly functioning RGBW light in HASS, and it also allows for the discovery to be changed for any RGB lights as well to cut z-wave traffic a bit when changing color, as an added benefit.

However, adding separate entities, as this PR does, to allow some control over WW and CW (as separate dimmers in HASS) isn't how HASS light entities should work. The solution is allowing multiple white values or colors to be set at once, as in [1782](https://github.com/zwave-js/node-zwave-js/pull/1782), which works with HASS properly, without added extra entities, and also lowers z-wave network traffic, as it can set multiple channels at once.

With this PR, I ended up with a still-broken light in HASS, and 4 separate entities: a broken RGBW bulb entity, which cannot be fixed by this PR, and needs the changes in 1782, a semi working RGB entity which would allow the control of color only, and 2 dimmers, one for the WW LED and one for CW LED. And in order for the semi-working RGB entity to work, I'd have to go in to each of those dimmers and set both of them to 0, otherwise the bulb is in "white mode" and won't display RGB color.

Additionally, by attempting to control the WW and CW channels with separate dimmer entities, rather than as part of the light itself, none of the HASS light service calls would work. If the light had anything other than 0 on WW or CW, and you attempted to call `light.turn_on` and specify `color_name: red`, it wouldn't work, because the light service would send an MQTT payload on the rgb_command_topic, but wouldn't be able to actually set the white channels.

With [1782](https://github.com/zwave-js/node-zwave-js/pull/1782) I was able to submit this Discovery and have a fully working Inovelli LZW42 Multi-Color Bulb:

<details><summary>Working discovery for RGBW bulb</summary>

````
{
  "device": {
    "identifiers": [
      "z2m_inovelli_lzw42_multi_color_bulb_family_room_floor_lamp_rgbw1"
    ],
    "manufacturer": "Inovelli",
    "model": "LZW42 Multi-Color Bulb",
    "name": "Family Room Floor Lamp RGBW1",
    "sw_version": "2.30"
  },
  "availability": [
    {
      "payload_available": "true",
      "payload_not_available": "false",
      "topic": "z2m/family_room_floor_lamp_rgbw1/status"
    }
  ],
  "brightness_state_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/currentValue",
  "brightness_command_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/targetValue/set",
  "brightness_value_template": "{{ (value|int) | round(0) }}",
  "brightness_scale": 99,
  "state_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/currentValue",
  "state_value_template": "{{ \"on\" if value|int > 0 else \"0\" }}",
  "command_topic": "z2m/family_room_floor_lamp_rgbw1/38/0/targetValue/set",
  "on_command_type": "brightness",
  "color_temp_state_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/currentColor",
  "color_temp_command_template": "{{ {'warmWhite': ((0.7349 * (value - 153))|round(0)), 'coldWhite': (255 - (0.7349 * (value - 153))|round(0)), 'red': 255, 'green': 255, 'blue': 255}|to_json }}",
  "color_temp_command_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/targetColor/set",
  "color_temp_value_template": "{{ '%03d%03d' | format((value_json.warmWhite|int), (value_json.coldWhite|int)) }}",
  "min_mireds": 153,
  "max_mireds": 500,
  "rgb_command_template": "{{ {'warmWhite': 0, 'coldWhite': 0, 'red': red, 'green': green, 'blue': blue}|to_json }}",
  "rgb_command_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/targetColor/set",
  "rgb_state_topic": "z2m/family_room_floor_lamp_rgbw1/51/0/targetColor",
  "rgb_value_template": "{{ value_json.red }},{{ value_json.green }},{{ value_json.blue }}",
  "name": "Family Room Floor Lamp RGBW1",
  "unique_id": "family_room_floor_lamp_rgbw1",
  "payload_on": "on",
  "payload_off": "0",
  "platform": "mqtt"
}
````
</detail>

If the Fibaro v1 has an issue with a binary switch that ends up in Zjs2MQTT as a color switch, wouldn't that be better handled by the device config for that one device? I haven't dug into the device configs in ZwaveJS much, but I was under the impression that there's a compat flag that can be used to work around buggy devices.

**Comment by robertsLando** [comment]
> I haven't dug into the device configs in ZwaveJS much, but I was under the impression that there's a compat flag that can be used to work around buggy devices.

**Comment by robertsLando** [comment]
@DanielWinks I like your discovery template, I think it would fix all problems with rgb lights. ATM the rgb lights discovery is handled in a function [addRgbDimmer](https://github.com/zwave-js/zwavejs2mqtt/blob/master/lib/Gateway.js#L776) that checks the available values and creates the discovery payload needed. Are you able to submit a fix for that?

**Comment by DanielWinks** [comment]
My particular device seems to have a hand-written [discovery](https://github.com/zwave-js/zwavejs2mqtt/blob/27245ec79a7c7ac6134c6403b82482d4deb82c25/hass/devices.js#L222) for it, which I will put out a PR for here today... just need to swap over from "value only" payloads to "time:value" first, since the discovery I posted would only work for "value only", I think... will need to slightly alter the `value_json` stuff to something probably a bit like `value_json.value.red` instead of just `value_json.red` like I'm currently using.

And on the note of me using "value only" (so I can use MQTT Device Triggers in HASS), I reached out to check on a different PR to see if the templates it added would allow templating the MQTT payload in the MQTT Device Triggers in HASS, but it didn't. Before I could even see that message, another [PR](https://github.com/home-assistant/core/pull/46891) was kindly put in that DOES allow for it, so hopefully quite shortly, HASS MQTT Device Triggers will allow templates in the payload, and thus be compatible with "time:value" and "entire object" gateway settings. So we'll be able to get all the various Z-wave buttons and scene controllers that currently create sensors in HASS to create actual "buttons" instead. I've got Discovery payloads for quite a few devices for that already (Inovelli Red dimmers, GE dimmers, Hank 4 button scene controller), so I'll be monitoring that and working to get some PRs in here so we can have device triggers.

**Comment by billiaz** [comment]
@DanielWinks 
Color Switch V2 supports separate brightness using a multilevel switch. V3 will also support brightness through the rgb
when you have a colorswitch device, Primary command class is the CS and not the multilevel or binary. Discovering the Color switch first makes sense, no?

Having the discovery on binary/multilevel first... can lead to issues like some devices without RGB face!
also some devices with binary and multilevel (plugs with RGB light) confuse the whole setup.

Also many devices (aeotec multi-color 6, fibaro rgb controller) as no ccSpecific to identify them. therefore it complicates it even more.
should we make hand written discoveries for all devices do not fit one setup? Or make a logic fitting most?

I understand the use of Color_value instead of white_value do not support template. but still can confuse people!
I strongly believe HASS needs to be patched also to support white_value_command_template. in order to keep the entity user friendly.

**Comment by DanielWinks** [comment]
> I strongly believe HASS needs to be patched also to support white_value_command_template. in order to keep the entity user friendly.

Probably, but it would be a significant patch. Presently, HASS has 2 command topics for (default schema) lights: RGB and White/color_temp.

Whenever a user selects a color in the [color picker in HASS](https://i.imgur.com/asw8Yra), it sends a single payload to the rgb_color command topic. Whenever they select a white shade on the "Color temperature" slider, it sends a single payload to the color_temp command topic (poorly named, as it's really "white temp"). RGB gets `red`, `green`, and `blue` as the inputs to the template for rgb_commands. When you select a white temp, HA only gives you `value` to use in the template, which is in mireds, and it expects that the template you write will be delivered on a single topic.

If they had some sort of command_topic_template where you could specify separate topics for WW and CW, that would be pretty hard to deal with, since HA just uses mireds as the value. If you look at the template I'm using presently, you'll see how convoluted it is to split the mireds HASS sends as "value" into the template into separate values for WW and CW: `"{{ {'warmWhite': ((0.7349 * (value - 153))|round(0)), 'coldWhite': (255 - (0.7349 * (value - 153))|round(0)), 'red': 255, 'green': 255, 'blue': 255}|to_json }}"` (this also sets RGB to 255 each, to center the color picker back on white, so you don't end up with an entity in HASS that looks like its in RGB mode when it's really in white mode).

We could have HASS allow for multiple command topics for white, but that has significant drawbacks. First, it amplifies zwave traffic, as it'd send 2 messages to set each of the 2 white channels, instead of 1 by using the json payloads allowed in [1782](https://github.com/zwave-js/node-zwave-js/pull/1782), and then you'd also need to have HASS patched to send a simultaneous command on rgb_command whenever you set a white value, otherwise you'd end up with a situation where the bulb reports something like 255, 173, 15 for RGB, which causes HASS to set the light entity to a color instead of the default "white" (off white, but it's the "this bulb is white" icon). This would result in a lot of extra z-wave traffic.

Basically, with how HASS works right now, if the bulb is on and in RGB mode, HA will only send a single payload to whatever is set for "color_temp" command topic when you change the white temp slider. This single payload must do the following: set RGB to 255,255,255, and set both WW and CW to the desired values, in a single payload. If your bulb is in white-mode (ie, WW or CW are anything other than 0), then when selecting a color HA will send a single message on rgb_command topic which must set RGB to the desired values, and set WW and CW both to 0, in a single payload.

Changing this in HASS would also probably break every MQTT light that anyone is using with Zwave2MQTT, or WiFi based MQTT lights using the default schema, or any other MQTT based light that's not ZwaveJS2MQTT. Overall, it feels like a much better solution to merge [1782](https://github.com/zwave-js/node-zwave-js/pull/1782) and #674, and update our Discovery payloads to work with the json payloads available in 1782. This will both fix most RGB/RGBW lights that presently aren't working with Zwavejs2mqtt master branch, and not result in needing a bunch of PRs in HASS that would also cause significant breaking changes to many people.

> Color Switch V2 supports separate brightness using a multilevel switch. V3 will also support brightness through the rgb
when you have a colorswitch device, Primary command class is the CS and not the multilevel or binary. Discovering the Color switch first makes sense, no?

Is doing that first what causes the LZW42 bulbs I have to end up with 4 entities? If so, then I'm not sure if doing that first is any better than how it currently works. While I wouldn't particularly care if I ended up with an RGBW entity (that worked, after 1782 is merged), and an unneeded RGB entity, and a dimmer for WW and CW, which I could just ignore... Many HASS users would be quite confused. With just 1782, I'm able to control all 5 channels on my bulbs, with a single entity. Whatever solution we pursue for other issues this PR is intended to correct for, we need to ensure that it doesn't result in devices ending up with 2 or 3 or 4 entities in HASS when 1 is sufficient. 

> Having the discovery on binary/multilevel first... can lead to issues like some devices without RGB face!
also some devices with binary and multilevel (plugs with RGB light) confuse the whole setup.

> Also many devices (aeotec multi-color 6, fibaro rgb controller) as no ccSpecific to identify them. therefore it complicates it even more.
should we make hand written discoveries for all devices do not fit one setup? Or make a logic fitting most?

The issue here is which devices are the "correct" ones, and which are the "incorrect" ones? From what I can see, my LZW42 are done the "correct" way. They have 5 LED channels, RGB WW CW. All 5 channels are controlled with COMMAND_CLASS_SWITCH_COLOR_V1 and they also use COMMAND_CLASS_SWITCH_MULTILEVEL_V2, as per the Z-wave specs. I'm not sure how taking what is a 5 channel COMMAND_CLASS_SWITCH_COLOR_V1 on the device and breaking it out to RGB and separate controls for WW and CW makes sense. From what I see, the LZW42 is about as spec-compliant of a COMMAND_CLASS_SWITCH_COLOR_V1 device as you can have.

It seems like if the changes in this PR break LZW42, in order to accomodate stuff like the Fibaro or Aeotec, which from what I can see, are not z-wave spec compliant, and feels like it is the wrong way to go about things. I don't think we should make something like the LZW42 worse just to accommodate a RGB ring on the Aeotec plug (which doesn't have any switch color cc attached). So if we're going to have to hand-write things, shouldn't we be doing that for the non-spec stuff, like the Aeotec, and not stuff that implements the proper CC's like the LZW42?

**Comment by scyto** [comment]
To be clear the fibaro doesn’t have a binary switch that is detected as color, it has a color switch that is not detected by hass mqtt discovery or hass integration, despite the color switch having a hex value.

The aeotec smart switch 6 was working fine until a couple of weeks ago when some change decided that if a device had a binary switch and a color switch then that binary switch must also be a color switch.

Maybe the fixes for rgbwwcw will fix or change that behavior, I will wait until the two changes for rgbwcw go through and then we can revisit the fibaro yet again.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

File: lib/Constants.js [hunk 6]
```diff
@@ -57,6 +57,32 @@ module.exports = {
       SHADE: 'shade',
       SHUTTER: 'shutter',
       WINDOW: 'window'
+    },
+    rgbWhiteTemp: {
```
robertsLando, line 61: @billiaz What is this for?

lib/Gateway.js
```diff
@@ -1674,12 +1686,12 @@ Gateway.prototype.discoverValue = function (node, vId) {
...
             // brightness level
-            cfg = addRgbDimmer.call(
+            /* cfg = addRgbDimmer.call(
               this,
               hassCfg.light_rgb_dimmer,
               node,
-              valueId
-            )
+              // valueId
+            ) */
```
robertsLando, line 1694: ?

package.json
```diff
@@ -125,36 +125,65 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.12.1",
+    "@release-it/conventional-changelog": "^2.0.0",
+    "@types/node": "^14.14.30",
+    "@varet/zj2m-prom-exporter": "^0.1.6",
```
robertsLando, line 130: Remove this

robertsLando (changes requested):

lib/Constants.js
billiaz: useless

lib/Gateway.js
billiaz: have to cleanup :)

lib/Gateway.js
```diff
@@ -771,43 +771,65 @@ function setDiscoveryValue (payload, prop, node) {
...
+    cfg.discovery_payload.state_topic = discoveredStateTopic
+    cfg.discovery_payload.command_topic = discoveredCommandTopic
+
+    cfg.discovery_payload.state_value_template = '{{ value_template.json }}'
+    cfg.discovery_payload.on_command_type = 'last'
   }
 
-  return false
+  // Use color_value_* from HASS discovery as white switch
+  if (node.values['51-0-currentcolor']) {
```
robertsLando, line 824:
```suggestion
  if (node.values[`51-${valueId.endpoint}-currentColor`]) {
```

lib/Gateway.js
```diff
@@ -771,43 +771,65 @@ function setDiscoveryValue (payload, prop, node) {
...
+  )
+  cfg.discovery_payload.rgb_command_topic =
+    cfg.discovery_payload.rgb_state_topic + '/set'
+
+  // The following part of code, checks if ML or Binary works. If one exists the other
+  let brightnessValue = false
+  let switchValue = false
+  if (node.values[`38-${valueId.endpoint}-currentValue`]) {
+    brightnessValue = `38-${valueId.endpoint}-currentValue`
+  } else if (node.values[`38-${valueId.endpoint + 1}-currentValue`]) {
```
robertsLando, line 789: Is this really needed? Why `+1`?

robertsLando (changes requested):

lib/Gateway.js
billiaz: I changed it a bit
fibaro uses endpoint 0 for hex but endpoint 1 for the 38 CC
new change is handling it better + comment

Remember still working on it! :) thanks for pointing out!

lib/Gateway.js
billiaz: more stuff need change here! i added some changed + `-0`in end

scyto:
i built and tested the branch.

For the Fibaro the color wheel now shows up in HA UI via MQTT but when used it doesn't seem to change the color values.  The log at info shows:

```
zwavejs2mqtt    | 2021-02-25 16:25:47.290 INFO MQTT: Message received on zwavejs2mqtt/WetBar/RGBW/51/0/hexColor/set, '{"warmWhite": 0, "coldWhite": 0, "red": 127, "green": 0, "blue": 255}'
zwavejs2mqtt    | 2021-02-25 16:25:47.293 INFO ZWAVE: Writing { warmWhite: 0, coldWhite: 0, red: 127, green: 0, blue: 255 } to 15-51-0-hexColor
zwavejs2mqtt    | 2021-02-25 16:25:47.297 INFO ZWAVE: Controller status: Driver: Color Switch: "hexColor" must be of type "string", received "object"
zwavejs2mqtt    | 2021-02-25 16:25:47.300 ERROR ZWAVE: Unable to write { warmWhite: 0, coldWhite: 0, red: 127, green: 0, blue: 255 } on 15-51-0-hexColor
``` 
this might be because my device only has the warm white i think?

scyto:
for the switch 6 issue - yay the binary switch is now a switch again
the dimmer is now treated as a dimmer by HA even though it should be treated as an rgbw_dimmer (via MQTT)

Lemme know if you want the nodes.json or the ones you have on file are good enough?

billiaz:
@scyto what is the driver version? 6.4.0 or 6.5.0?

scyto:
6.5

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

thundergreen:
Do we have to re-interview the RGB devices? At the momont I am not able to test as lastest HASS beta version Home Assistant 2021.3.0b1 breaks Z-Wave JS to MQTT (b897934) as stated here: https://github.com/hassio-addons/addon-zwavejs2mqtt/issues/58

Hope there will be a fix soon and that we can finally properly use our RGB lights with color wheel as expected :) Thanks a lot for this efforts here! TOP

robertsLando:
> Do we have to re-interview the RGB devices

thundergreen:
Just updated to version 2.0.0 via Hassio ADDOn but my rgb controller still has one switch for each color. What steps are neccessairy to get this rgb running? i tried re-nterview but without success. :( Any hint. ?

robertsLando:
@thundergreen If you are usiung zwavejs integration this will have no effect as here we fixed mqtt discovery

PR #577, opened 2021-02-13
Update docker-compose.yml to add timezone and time sync volumes

scyto:
Added options to sync timezone and time as someone asked about TZ env var on slack.

I am sure we will get more ask on discord over time so getting ahead of that.
This is tested and working on my own machine. 

I prefer this solution as it should work for most people with only need to remove the comment char.
If we set TZ var in this file it would be right for only one TZ in the world.


scyto:
i am unhappy with the PR, pulling until i can figure out why it changed the vscode settings

PR #608, opened 2021-02-17
chore: improved release workflow and changelog

robertsLando:
@chrisns I'm thinking about switch to a new release flow. I need feedbacks about it


chrisns (approved):

chrisns (changes requested):
I'm really really uncomfortable with logic gates that control use/access to privileged tokens not even having unit tests around them.

robertsLando:
@chrisns Could you exmplain better? What should I add to make this secure?

.github/workflows/release-it.yml
```diff
@@ -0,0 +1,36 @@
...
+  workflow_dispatch:
+    inputs:
+      type:
+        description: "Type. Can be `patch` `minor` or `major`"
+        required: true
+        default: "patch"
+
+jobs:
+  # #########################################################################
+  # Create a new release when an authorized person posts "@zwave-js-bot release <patch|minor|major>"
```
AlCalzone, line 13: The comment is no longer correct

.github/workflows/release-it.yml
```diff
@@ -0,0 +1,36 @@
+name: "Release-it: Create a new release on demand"
+
+on:
+  workflow_dispatch:
+    inputs:
+      type:
+        description: "Type. Can be `patch` `minor` or `major`"
```
AlCalzone: I don't see where this is used?

AlCalzone: And do you need to verify the input somehow?

AlCalzone (changes requested):

.github/workflows/release-it.yml
robertsLando: Fixed :)

.github/workflows/release-it.yml
robertsLando: That's automatically done by release-it. Also it automatically do tests, lint and all the other stuff needed

.github/workflows/release-it.yml
```diff
@@ -0,0 +1,36 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      type:
+        description: 'Type. Can be `patch` `minor` or `major`'
+        required: true
+        default: 'patch'
+jobs:
+  release-it:
```
AlCalzone, line 14: You should limit this action to whoever you're comfortable with to do releases.

> To trigger a workflow in a repository, the user should be a collaborator with Write permission in the repository. Normally the external users can’t trigger workflows in the repository.

AlCalzone (changes requested):

chrisns (approved):

.github/workflows/release-it.yml
robertsLando: Fixed! Thanks :)

.github/workflows/release-it.yml
```diff
@@ -0,0 +1,37 @@
...
+    if: contains('["robertsLando","chrisns"]', github.actor)
```
AlCalzone, line 15: I'm not sure if you even need that strange array syntax - it is a string after all ;)

.github/workflows/release-it.yml
```diff
@@ -0,0 +1,37 @@
+# #########################################################################
+# Creates a new release using `workflow_dispatch` event trigger with `type`
+# as input to describe the type of release to create
+name: 'Release-it: Create a new release on demand'
+
+on:
+  workflow_dispatch:
+    inputs:
+      type:
+        description: 'Type. Can be `patch` `minor` or `major`'
```
AlCalzone, line 10: Do you need verification that the type was entered correctly?

robertsLando: Nope that's handled by release-it :)

.github/workflows/release-it.yml
robertsLando: Dunno how I missed that but good catch! :) Fixed

.github/workflows/release-it.yml
```diff
@@ -12,7 +12,7 @@ on:
         default: 'patch'
 jobs:
   release-it:
-    if: contains('["robertsLando","chrisns"]', github.actor)
+    if: contains(["robertsLando","chrisns"], github.actor)
```
AlCalzone: And now I'm not sure if that works. Either
```
if: contains("robertsLando,chrisns", github.actor)
```
which is probably unsafe if someone decides to call himself "robert", or with `fromJSON`:
https://github.community/t/passing-an-array-literal-to-contains-function-causes-syntax-error/17213/3

robertsLando: Used the safer way: `    if: github.actor == 'robertsLando' || github.actor === 'chrisns'`

AlCalzone (approved):

PR #609, opened 2021-02-17
feat: bump zwavejs server beta 7 and show version on ui

robertsLando:
cc @frenck


frenck:
Thanks, @robertsLando 👍

PR #618, opened 2021-02-18
fix(chore): dockerfile.contrib use slim for runtime, also clean devices index

jcam:
- The devices index.json should be generated at run time, any index in the tree is likely out of date.
- do not need full erbium for runtime


robertsLando (approved):
LGTM

Issue #595, opened 2021-02-16
[bug]  Aeotec Switch 6 - binary switch doesn't work any more as it has wrong (light) domain.


scyto:
Before submitting a bug please read: <https://zwave-js.github.io/zwavejs2mqtt/#/troubleshooting/bug_report>

### Version
1.1.1+94ff7ed.94ff7ed

Build/Run method

- [X] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build )

zwavejs2mqtt version: 1.1.1+94ff7ed.94ff7ed
zwavejs version: 6.3.0-8f22052

### Describe the bug

The Aeotec Switch 6 - binary switch doesn't work any more.
This breaks the switch when using MQTT and HASS discovery. 
This seems to be a regression and was fine in earlier versions.

### To Reproduce

1. Install latest version.
2. Look at the device value and see  binary switch 
 ![image](https://user-images.githubusercontent.com/11369180/108007609-4311c800-6fb3-11eb-9f6c-d5e8d13b8704.png)

5. Look at discovery code and note it an rgb switch.

7. In previous version this was discovered as switch.masterbedroom_outlet_rgb_switc.   In this version it is discovered as light.masterbedroom_outlet_rgb_switch

10. This binary switch is not a color light - it only controls load for the outlet on this device.
11. Yes this device has am rgb color device - but that is purely for reporting status, energy use etc.
12. the binary switch doesn't control the rgbw device in an way.

### Expected behavior

I expect this to work as it did in earlier versions and be surfaced as a binary switch.

### Additional context

@billiaz not sure if this is related to any of the changes to making color lights works?

scyto:
[node_11.json.txt](https://github.com/zwave-js/zwavejs2mqtt/files/5985276/node_11.json.txt)

scyto:
I note the whole device is now in the light discovery section, on previous versions it was in the switch discover section (which would be correct, this is a switch not a light).  

![image](https://user-images.githubusercontent.com/11369180/108015189-7dd02c00-6fc4-11eb-83f6-e81d57504334.png)

scyto:
With help from RobC pretty certain it is these changes that caused this. One can't assume a device with a switch and rgbw is an rgbw dimmer / light https://github.com/zwave-js/zwavejs2mqtt/pull/290/files.

The switch in this device just controls the 120v outlet on it.

billiaz:
[...]

PR #625, opened 2021-02-18
feat: split prioritized CCs and discover first 

billiaz:
This PR tries to help dependent CCs to be discovered first.

- Add Constant to allow easy addition of priority CCs
- add a small function to create split list
- iterate through the list


lib/Gateway.js
```diff
@@ -567,6 +570,26 @@ function getNodeName (node, ignoreLoc) {
...
+ * @param {node.valueId} values valueId keys
+ * @returns An nested array, first value is priority CC values and second the remaining
+ */
+
+function getPriorityCCFirst (values) {
+  const priorityList = []
+  const remainingList = []
+  for (const k in values) {
+    if (Constants.priorityCommandClass.includes(values[k].commandClass)) {
+      priorityList.push(k)
```
robertsLando, line 585: You can use a single array and use the unshift command to push in the front of the array

lib/Gateway.js
```diff
@@ -567,6 +570,26 @@ function getNodeName (node, ignoreLoc) {
...
+ *  Return re-arranged based on critical CCs
+ *
+ * @param {node.valueId} values valueId keys
+ * @returns An nested array, first value is priority CC values and second the remaining
+ */
+
+function getPriorityCCFirst (values) {
+  const priorityList = []
+  const remainingList = []
+  for (const k in values) {
```
robertsLando, line 583: I would call K id

lib/Gateway.js
```diff
@@ -1080,8 +1103,11 @@ Gateway.prototype.rediscoverNode = function (nodeID) {
     nodeDevices.forEach(device => this.discoverDevice(node, device))
 
     // discover node values (that are not part of a device)
-    for (const id in node.values) {
-      this.discoverValue(node, id)
+    // iterate prioritized first, then the remaining
+    for (const groupedValues of getPriorityCCFirst(node.values)) {
```
robertsLando, line 1107: Use vId here instead of groupedValues

lib/Gateway.js
```diff
@@ -1080,8 +1103,11 @@ Gateway.prototype.rediscoverNode = function (nodeID) {
...
+      for (const id of groupedValues) {
```
robertsLando, line 1108: I don't understand this, you should just use 

node.values[vId] here, you don't need another nested for

robertsLando (changes requested):

lib/Gateway.js
billiaz: good one, done that!

lib/Gateway.js
billiaz: it is simplified once we used the single array

billiaz:
This is part of multiple PRs to solve the RGBW

this first PR is to increase HASS flexibility and tries to partially solve : #595 #477 and #351

After that another pr will introduce an improved `Color Switch`

lib/Gateway.js
```diff
@@ -567,6 +568,25 @@ function getNodeName (node, ignoreLoc) {
   )
 }
 
+/**
+ *  Return re-arranged based on critical CCs
+ *
+ * @param {node.valueId} values valueId keys
```
robertsLando, line 574:
```suggestion
 * @param {Map<string, import('../types').Z2MValueId>} node values map
```

lib/Gateway.js
```diff
@@ -567,6 +568,25 @@ function getNodeName (node, ignoreLoc) {
...
+ * @returns An nested array, first value is priority CC values and second the remaining
```
robertsLando, line 575:
```suggestion
 * @returns {string[]} Array of values Ids sorteb by CC discovery priority
```

robertsLando (changes requested):

lib/Constants.js
```diff
@@ -59,6 +59,9 @@ module.exports = {
       WINDOW: 'window'
     }
   },
+  priorityCommandClass: [
+    51 // Color Switch
+  ],
```
robertsLando, line 64: I suggest to move this const inside the function `getPriorityCCFirst` on gateway side for now, and instead of hard coding the 51 use the zwavejs COmmandClasses struct (like we do in the switch case in discover value

robertsLando (changes requested):

robertsLando (approved):
LGTM

PR #628, opened 2021-02-19
chore(docker): updated docker-compose.yml to add TZ variable

scyto:
added TZ env var to docker-compose file as hint to new folks asking about timestamp in logs


BluGeni (commented):
Line 9 there already is an environment declaration. Might want to move your TZ comment to below that.

scyto:
> Line 9 there already is an environment declaration. Might want to move your TZ comment to below that.

done, when did that new secrets env var sneak in :-)

robertsLando:
@scyto Could you add that also to docker readme? it contains a copy of that copose file

scyto:
yes, will do it today or tomorrow

scyto:
@robertsLando amended readme, sorry took so long.

robertsLando (approved):

PR #656, opened 2021-02-19
feat(ui): add location to graph label and tooltip

ahochsteger:
This PR adds the node location (if set) to the node label and node tooltip:
image


src/components/custom/ZwaveGraph.vue
```diff
@@ -769,10 +771,11 @@ export default {
           failed: node.failed,
           title:
             '<b>' +
-            nodeName +
+            node._name +
```
robertsLando, line 774: @ahochsteger Shouldn't this be `nodeName` instead? BTW I was planning to remove the tooltip and use the popup I was using in the old mesh version. The code is still there, we should detect the node click (it's already detected but nothing happens atm) and when clicked use the code in Mesh.vue to display the draggable info panel. Thoughts?

robertsLando (changes requested):

src/components/custom/ZwaveGraph.vue
```diff
@@ -752,7 +752,9 @@ export default {
 
         batlev = batlev ? batlev.value : undefined
 
-        const nodeName = node.name || 'NodeID ' + node.id
+        const nodeName =
+          (node.name || 'NodeID ' + node.id) +
+          (node.loc ? '\n(' + node.loc + ')' : '')
```
robertsLando, line 757:
```suggestion
          (node.loc ? ' (' + node.loc + ')' : '')
```

robertsLando (changes requested):

src/components/custom/ZwaveGraph.vue
ahochsteger: @robertsLando you're right - this was a left-over from a previous variant that used `_name` since it already contained the location. It's done.
Yes, it would be good to get the view back that was shown in the old graph. This would allow us to put more functionality there that is helpful in the graph view.

robertsLando: Would you add that in this pr?

src/components/custom/ZwaveGraph.vue
ahochsteger: @robertsLando I used a line break here to use two lines and prevent the node symbols to get too wide, since the graph is already very wide.
I did not manage to show both lines centered though.

ahochsteger: This is what mine looks currently - without showing the location, which would make it much wider:
image

robertsLando: Ok ignore this, btw if we will use the popup window this part will be removed, I prefer the info window to the popup

src/components/custom/ZwaveGraph.vue
ahochsteger: @robertsLando  I can try.

While we are at it, I've got a question concerning which functionality of `ManagedItems` we'd like to sync between the nodes table and the mesh graph.

What I'd like to see synced is:
* Node filter
* Node grouping

I'm unsure about:
* Node selection
* Visible node properties (or table columns)

Also I'm unsure, if we always want to sync this functionality or if the user should be able to decide if they are linked together.
In case we always sync, I'm thinking about making the managed nodes stuff available using Vuex to access it from both the nodes table and the mesh graph.
What do you think?

robertsLando: I wouldn't sync anything bewtween the two tables, I would simply use the same underlying logic for filtering

PR #666, opened 2021-02-20
bug: fixed dockerfile and dockerfile.contrib to use node:erbium-buster-slim

scyto:
dependabot changed all references to  to `node:erbium-buster-slim` to `node:15.9.0-alpine`.
This PR reverts those changes.,

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 # STEP: 1
-FROM node:15.9.0-alpine AS build-z2m
+FROM node:erbium-buster-slim AS build-z2m
 
 WORKDIR /usr/src/app
 
@@ -38,7 +38,7 @@ RUN npm run build && \
         stylesheets
 
 # STEP: 2 (runtime)
-FROM node:15.9.0-alpine
+FROM node:erbium-buster-slim
 
 RUN apk add --no-cache \	
     libstdc++  \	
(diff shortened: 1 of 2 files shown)


scyto:
@jcam  erbium-buster-slim is failing on the build check 

the github workflow expects `linux/arm64,linux/amd64,linux/arm/v6,linux/arm/v7`
node:erbium-buster-slim
has 
image

options:

1. we pick a different base buster and install node manually
2. we change the github workflow to remove arm/v6

jcam:
What workflow is running the dockerfile.contrib?
I picked erbium-buster-slim because that was one we used to use here before dependabot broke it (the first time, back in january):
https://github.com/zwave-js/zwavejs2mqtt/commit/9767d76df0e1760504ee74b8b32ba0e18e85b304#diff-4ddc2e10398d5c169795b7937f893ec72d2619b395259c592156a5dd0b476c9f

jcam:
the regular Dockerfile should be using node alpine

scyto:
> the regular Dockerfile should be using node alpine

ok will make the change, why the difference?

jcam:
contrib does a lot more :D

jcam:
if you use slim up top, you'll need to run an apt-get install a few things... build-essential, git, jq... i think that's all

scyto:
> if you use slim up top, you'll need to run an apt-get install a few things... build-essential, git, jq... i think that's all

I put it back to alpine for the dockerfile but to erbirum-buster-slim (to mirror your change that got wiped away), 
i had (wrongly) assumed that both edits by the dependabot was bad.

robertsLando (approved):

PR #674, opened 2021-02-21
fix: propertyKey may be zero

AlCalzone:
This fixes the missing color channel 0 for Color Switch CC and avoids value clashes between all values and color 0 when https://github.com/zwave-js/node-zwave-js/pull/1782 is merged.

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -1043,7 +1043,7 @@ Gateway.prototype.valueTopic = function (node, valueId, returnObject = false) {
         topic.push('endpoint_' + (valueId.endpoint || 0))
 
         topic.push(utils.removeSlash(valueId.propertyName))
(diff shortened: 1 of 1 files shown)


DanielWinks:
I can confirm this does indeed fix the issue described above. Also, when in combination with [1782](https://github.com/zwave-js/node-zwave-js/pull/1782) I can now fully and properly control my Inovelli Ilumin bulbs, which have RGB and WW and CW channels.

billiaz (approved):

billiaz:
I found the same issue, @AlCalzone you where faster submitting. ;)

robertsLando:
Oh damn, thanks @AlCalzone !

billiaz:
@robertsLando I am closing PR #655 as it fixes the same thing

PR #688, opened 2021-02-22
feat: plugins improved support

robertsLando:
With this PR users will have more power when developing custom plugins

BREAKING CHANGE: `plugins` are now stored in an array of `strings` on settings `gateway` prop instead of `zwave`


chrisns (changes requested):

docs/guide/plugins.md
```diff
@@ -0,0 +1,52 @@
+# Plugins
+
+Plugins are nodejs packages that can be integrated in zwavejsmqtt in order to add new awesome features. They have access to all the clients (zwave and mqtt) and express instance.
+
+## Usage
+
+A plugin is imported in zwavejs1mqtt using `require(pluginName)(contex)` where context provides access to these elements:
```
robertsLando, line 7:
```suggestion
A plugin is imported in zwavejs1mqtt using `require(pluginName)(context)` where context provides access to these elements:
```

PR #710, opened 2021-02-23
refactor: make nodes a map instead of an array

robertsLando:
ZwaveClient `nodes` is now a Map nodeId -> node

BREAKING CHANGE: `getNodes` function of ZwaveClient returns an array of available nodes but the index doesn't match the nodeId
`refreshNeighbours` returns a map nodeId -> neighbours[]
node removed socket event now returns the node removed
renamed `setNodeName` and `setNodeLocation` apis (removed the leading `_`) as now them also store the value to controller


lib/Gateway.js
```diff
@@ -1091,7 +1091,7 @@ Gateway.prototype.valueTopic = function (node, valueId, returnObject = false) {
  * @param {number} nodeID
  */
 Gateway.prototype.rediscoverNode = function (nodeID) {
-  const node = this.zwave.nodes[nodeID]
+  const node = this.zwave.nodes.get(nodeID)
   if (node) {
```
AlCalzone, line 1095: If you check first with `this.zwave.nodes.has(nodeID)`, you shouldn't need to check if `node` is truthy.

lib/Gateway.js
```diff
@@ -1219,15 +1219,15 @@ Gateway.prototype.rediscoverAll = function () {
   if (!this.config.hassDiscovery) return
 
   const nodes = this.zwave ? this.zwave.nodes : []
-  for (let i = 0; i < nodes.length; i++) {
-    const devices = nodes[i] && nodes[i].hassDevices ? nodes[i].hassDevices : {}
+  for (const [nodeId, node] of nodes) {
+    const devices = node.hassDevices ? node.hassDevices : {}
```
AlCalzone, line 1223:
```suggestion
    const devices = node.hassDevices || {}
```

lib/ZwaveClient.js
```diff
@@ -107,8 +107,26 @@ function ZwaveClient (config, socket) {
...
+        this.storeNodes[n.id] = n
+      }
+    }
+
+    this.storeNodes = storeNodes
+
+    jsonStore.put(store.nodes, storeNodes).catch(err => {
+      logger.error('Error while updating store nodes', err)
+    })
+  }
```
AlCalzone, line 128: I don't really see the point of this.

robertsLando: It would work in both ways, was just for consistency

lib/Gateway.js
robertsLando: yeah did mostly find/replace in the code :)

lib/ZwaveClient.js
robertsLando: Nope I really need this

PR #728, opened 2021-02-24
feat: zwavejs@6.5.0 and nodeFilter setting support

robertsLando:


mgriffin13:
I tried the `:dev` image, but it doesn't look like it supports multiple nodes. Is that intentional? like, i can set the `Log Nodes` field to `5`, but I cannot set it to `5, 21`. I also tried `[5, 21]` and `5,21` (no space)

robertsLando:
you have to insert the node number and press enter

robertsLando:
Schermata da 2021-02-24 16-29-10

mgriffin13:
aha, perfect. thank you for the clarification

PR #732, opened 2021-02-24
hass(improvement): stop specific values from populating

billiaz:
Some values cause too much chatter in Hass, we remove these from discovery populating

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -577,21 +577,9 @@ function getNodeName (node, ignoreLoc) {
 
 function getPriorityCCFirst (values) {
   const priorityCC = [CommandClasses['Color Switch']]
-  const ignoreDiscovery = {
-    Meter: ['previousValue', 'deltaTime']
-  }
   const prioritizedValueIds = []
 
   for (const id in values) {
-    if (ignoreDiscovery[values[id].commandClassName]) {
-      if (
-        ignoreDiscovery[values[id].commandClassName].includes(
-          values[id].Property
-        )
-      ) {
-        continue
-      }
-    }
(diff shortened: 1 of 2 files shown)


robertsLando (changes requested):
I prefer to move the check in `discoverValue` function

billiaz:
wouldn't that iterate all values and just return, instead of stoping earlier?

robertsLando:
@billiaz discoverValue could also run 'on fly' and in that case this values will be discovered because you rely on the filter done by the prioritized cc function.

billiaz:
@zwave-js-bot fix lint

lib/Gateway.js
```diff
@@ -1608,6 +1609,14 @@ Gateway.prototype.discoverClimates = function (node) {
...
   const valueId = node.values[vId]
 
+  // List of values
+  const ignoreDiscovery = {
+    Meter: ['previousValue', 'deltaTime']
+  }
+  if (ignoreDiscovery[valueId.commandClassName]) {
+    if (ignoreDiscovery[valueId.commandClassName].includes(valueId.property))
+      return
+  }
```
robertsLando, line 1619: I like this but for now there are much other cc we are ignoring, IMO this check should be done inside the CC switch-case, like `if(property === x || property === y) return`

robertsLando (changes requested):

lib/Gateway.js
aretakisv: thought of. looking if we have second class, otherwise i will move it under. the idea was to cover all

robertsLando:
With zwavejs 7 deltaTime and previous value will not be available anymore so this can be closed

PR #769, opened 2021-02-28
fix: lookup associated node by id not by index

### LINKED ISSUE #766: [BUG] Display bug in associations
Description:
Using the ZwaveJS2MQTT HomeAssistant addon (V0.5.0), the control panel shows lifeline associations to node 2, not to node 1.  The associations themselves appear to be made to node 1, so this seems to be purely a display bug.

Describe the bug
Lifeline group associations show association to node 2, instead of the expected association to the controller (node 1).  Removing the association and adding an association to node 1 again displays as an association to node 2.

I also verified that other associations, at least in group 1, also exhibit this off-by-1 behavior.  (E.G., association to node 5 displays as an association to node 6.)

Expected behavior
Name of associated node displayed should reflect the associated node, rather than the name of nodenumber+1.

osenft:
Fixes #766.

### DIFF
--- a/src/components/nodes-table/AssociationGroups.vue
+++ b/src/components/nodes-table/AssociationGroups.vue
@@ -20,7 +20,8 @@
                 <v-list-item-title
                   >Node:
                   <strong>{{
-                    nodes[ass.nodeId]._name || ass.nodeId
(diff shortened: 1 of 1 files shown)


robertsLando:
@osenft Working on a better solution here: #780

PR #797, opened 2021-03-02
fix: read/parse buffers as hex strings

### LINKED ISSUE #782: [question] Enrollment codes for BeNext/Zipato Keypad
Hello

Sorry in advance if this question has already been posted/answered.
I have a BeNext/Zipat mini-keypad that I used to use with Domoticz/OZW. I am now using ZWaveJS2mqtt with HomeAssistant, and I'd like to use my tags. I can see in the UI that I can manually add user codes (most likely including the tags codes), but I can't figure out how to enroll my tags. I tried to press home/away on my tag reader using debug, but the code won't appear here.
I currently don't have a compatible NFC reader, but would buy one if needed.

Can you please help me?

Thanks in advance

robertsLando:
Fixes #782

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -1092,7 +1092,19 @@ function updateValue (zwaveNode, args) {
     const valueId = node.values[getValueID(args)]
 
     if (valueId) {
-      valueId.value = args.newValue
+      let newValue = args.newValue
+      if (Buffer.isBuffer(newValue)) {
+        // encode Buffers as HEX strings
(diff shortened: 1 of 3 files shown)


AlCalzone (changes requested):
You should also make sure to print buffers in that form, instead of `[object ArrayBuffer]`

AlCalzone (approved):
This doesn't affect the zwave-js-server, right?

robertsLando:
> This doesn't affect the zwave-js-server, right?

Nope, they have their own event listeners I only pass the driver instance to them :)

PR #816, opened 2021-03-04
fix: checks on `sendCommand` api

robertsLando:


docs/guide/mqtt.md
```diff
@@ -80,7 +80,10 @@ This are the available apis:
 - `beginFirmwareUpdate(nodeId, fileName, data, target)`: Starts a firmware update of a node. The `fileName` is used to check the extension (used to detect the firmware file type) and data is a `Buffer`
 - `abortFirmwareUpdate(nodeId)`: Aborts a firmware update
 - `writeValue(valueId, value)`: Write a specific value to a [valueId](https://zwave-js.github.io/node-zwave-js/#/api/valueid?id=valueid)
-- `sendCommand(valueId, command, args)`: Send a custom command
+- `sendCommand(ctx, command, args)`: Send a custom command.
+  - `ctx`:context to get the instance to send the command (`{ nodeId: number, endpoint: number, commandClass: number }`)
+  - `command`: the command string name
```
AlCalzone, line 85:
```suggestion
  - `command`: the command name
```
or `method name` or `function name`. Should have a reference to https://zwave-js.github.io/node-zwave-js/#/api/CCs/index

lib/ZwaveClient.js
```diff
@@ -2376,45 +2376,51 @@ ZwaveClient.prototype.hardReset = async function () {
...
       )
     }
-    const api = endpoint.commandClasses[valueId.commandClass]
 
+    // get the command class instance to send the command
+    const api = endpoint.commandClasses[ctx.commandClass]
     if (!api || !api.isSupported()) {
       throw Error(
-        `Node ${valueId.nodeId} (Endpoint ${valueId.endpoint}) does not support CC ${valueId.commandClass}`
+        `Node ${ctx.nodeId} (Endpoint ${ctx.endpoint}) does not support CC ${ctx.commandClass}`
```
AlCalzone, line 2415: The message is not 100% correct. If `!api` then the CC API is not implemented. That does **not** mean that the endpoint does not support hte CC.

AlCalzone (changes requested):

lib/ZwaveClient.js
robertsLando: I think that error is the same I copied from your io.broker moudule. BTW I have make it more clear now

PR #854, opened 2021-03-09
fix(ui): redirects to main when auth is disabled

### LINKED ISSUE #675: [bug] Page deep-links and browser reloads always switch to control panel
### Version

Build/Run method: Docker
zwavejs2mqtt version: 1.2.2
zwavejs version: 6.4.0

### Describe the bug

When I reload a page other than the control panel it always switches back to the control panel.

### To Reproduce

Steps to reproduce the behavior for browser reload:

1. Go to any page other than the control panel (e.g. the settings page)
2. Click on reload in the browser
3. The browser changes the URL to the control panel

Steps to reproduce the behavior for deep-linking:

1. Open a direct link to any page other than the control panel (e.g. .../settings)
2. The browser changes the URL to the control panel

### Expected behavior

The browser should keep the current URL on reload or stay at the given URL when using a direct link.

### Additional context

It seems that this behavior change has been introduced with #591 and #640.

robertsLando:
Fixes #675

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -497,7 +497,7 @@ export default {
           }
 
           if (!newAuth && this.$route.path === Routes.login) {
-            this.$router.push(Routes.main)
+            this.$router.push(this.$route.params.nextUrl || Routes.main)
           }
           this.startSocket()
         }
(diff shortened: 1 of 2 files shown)


robertsLando:
Unfortunaly this breaks the refrsh when auth is enabled :cry:

robertsLando:
Tricky solutions but didn't find anything better. Now this works for both auth and no-auth environments :tada:

PR #864, opened 2021-03-10
feat: move to zwavejs 7

robertsLando:
Updated zwavejs to version 7

BREAKING CHANGE: Most changes are already documented [here](https://zwave-js.github.io/node-zwave-js/#/getting-started/migrating-to-v7):
- Corrected parsing of Node Information Frames (NIF), reworked node properties
- No automatic query of all node values when restarting from cache, `interview completed` event is no longer emitted on startup
- Reworked "notification" event, node notifications are mapped to mqtt using a different topic/payload
- This version is incompatible with HA versions before 2021.4.x


jcam:
There is work required for the HA integration as well, this version will most likely not work. Might be useful to include a note in the breaking change list saying it is incompatible with HA versions before N, where N is likely 2021.4.x
(and since it doesn't have the updated server in it yet, because it hasn't been written yet, it probably is incompatible regardless)
Reference: https://github.com/home-assistant-libs/zwave-js-server-python/issues/162

robertsLando:
@jcam Updated breaking changes list

raman325:
https://github.com/zwave-js/zwave-js-server/pull/165 was merged and https://github.com/zwave-js/zwave-js-server/pull/167 is pending. Once that is merged, `zwave-js-server` *should* be backwards compatible although I am still testing that

lib/ZwaveClient.js
```diff
@@ -899,13 +922,14 @@ function initNode (zwaveNode) {
   node.productType = zwaveNode.productType
   node.manufacturer = deviceConfig.manufacturer
   node.firmwareVersion = zwaveNode.firmwareVersion
-  node.zwaveVersion = zwaveNode.version
+  node.protocolVersion = zwaveNode.protocolVersion
   node.zwavePlusVersion = zwaveNode.zwavePlusVersion
-  node.nodeType = zwaveNode.nodeType
-  node.roleType = zwaveNode.roleType
+  node.nodeType = zwaveNode.zwavePlusNodeType
+  node.roleType = zwaveNode.zwavePlusRoleType
```
AlCalzone, line 928: Are you not interested in the new `nodeType` (controller / slave) property?

AlCalzone: And since you're dumping all of this info, don't you want to know or display the (max) data rate(s)?

lib/ZwaveClient.js
```diff
@@ -899,13 +922,14 @@ function initNode (zwaveNode) {
...
   node.endpointsCount = zwaveNode.getEndpointCount()
   node.isSecure = zwaveNode.isSecure
-  node.isBeaming = zwaveNode.isBeaming
+  node.supportsSecurity = zwaveNode.supportsSecurity
+  node.supportsBeaming = zwaveNode.supportsBeaming
   node.isControllerNode = zwaveNode.isControllerNode()
   node.isListening = zwaveNode.isListening
   node.isFrequentListening = zwaveNode.isFrequentListening
```
AlCalzone, line 935: This is not just a boolean now, in case you're expecting that somewhere

robertsLando: Nope, I'm using `isListening` in zwave graph

lib/ZwaveClient.js
robertsLando: Forgot to dump that, adding it now

raman325:
FYI `zwave-js-server 1.3.0` was just released and it has `zwave-js 7.0.0` support

raman325:
@robertsLando I noticed that HA is still listed as a breaking change. See my comment here: https://github.com/zwave-js/zwavejs2mqtt/pull/864#issuecomment-805978952 (edited after the fact)

robertsLando:
Sorry will remove that from release notes thanks

PR #899, opened 2021-03-16
chore(deps): bump @zwave-js/server@1.2.0

robertsLando:


raman325:
@robertsLando we should wait until the patch release gets released for zwave-js-server and bump here before the next release. There's a pretty bad bug in 1.2.0 that will get fixed in 1.2.1


https://github.com/zwave-js/zwave-js-server/pull/166

PR #920, opened 2021-03-21
fix: use prefixed node.id in mqtt discovery topic

tmaroschik:
If spaces are used within the node name, the mqtt discovery topic contains spaces as well.
This is not supported by a lot of MQTT clients, like homeassistant.

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -676,8 +676,10 @@ function deviceInfo (node, nodeName) {
  * @param {string} nodeName Node name from getNodeName function
  * @returns The topic string for this device discovery
  */
-function getDiscoveryTopic (hassDevice, nodeName) {
-  return `${hassDevice.type}/${nodeName}/${hassDevice.object_id}/config`
+function getDiscoveryTopic (hassDevice, nodeId) {
+  return `${hassDevice.type}/${NODE_PREFIX + nodeId}/${
+    hassDevice.object_id
+  }/config`
 }
 
 /**
@@ -1456,7 +1458,7 @@ Gateway.prototype.discoverDevice = function (node, hassDevice) {
           '_' +
           hassDevice.object_id
 
-        const discoveryTopic = getDiscoveryTopic(hassDevice, nodeName)
+        const discoveryTopic = getDiscoveryTopic(hassDevice, node.id)
         hassDevice.discoveryTopic = discoveryTopic
 
         // This configuration is not stored in nodes.json
@@ -2157,7 +2159,7 @@ Gateway.prototype.discoverValue = function (node, vId) {
       '_' +
       utils.sanitizeTopic(valueId.id, true)
 
-    const discoveryTopic = getDiscoveryTopic(cfg, nodeName)
+    const discoveryTopic = getDiscoveryTopic(cfg, node.id)
 
     cfg.discoveryTopic = discoveryTopic
     cfg.values = cfg.values || []


robertsLando:
@tmaroschik This could break existing instances. I would prefer to sanitize node name instead using the sinitize function you can find in utils

tmaroschik:
I don't think so, because the topic in HASS discovery doesn't have a meaning. At least during testing it didn't make a difference upon changing.

robertsLando:
So you can confirm that if someone had a node previously discovered using the node name instead the id this would not break his instances? Where does that node id is used?

tmaroschik:
Yes I can confirm this by reading the manual: https://www.home-assistant.io/docs/mqtt/discovery/#discovery_prefix

>>> <node_id> (Optional): ID of the node providing the topic, **this is not used by Home Assistant but may be used to structure the MQTT topic**.

robertsLando (approved):

robertsLando:
Thanks @tmaroschik

embcla:
This was a breaking change for me as I run multiple instances of zwavejs2mqtt, now there are unavoidable collisions whenever there are the same node IDs in different instances.
Anyone with multiple instances will needs something different than just node ID.

blhoward2:
@robertsLando I think this may need to be reverted or reformulated. See my note on Slack. Is this also breaking node naming generally?

blhoward2:
> Yes I can confirm this by reading the manual: https://www.home-assistant.io/docs/mqtt/discovery/#discovery_prefix
> 
> 
> 
> > <node_id> (Optional): ID of the node providing the topic, **this is not used by Home Assistant but may be used to structure the MQTT topic**.

I think this is a misreading of how this is used. It's not used by HA to define the node ID but this PR removes the naming which is used to setup the name.

blhoward2:
Maybe the docs are inaccurate. I wonder if in testing this PR mqtt didn't really rediscover the device?

robertsLando:
Totally forgot about this, this is the reason that other user was expecting issue with discovery for sure. Should I revert this and release a new patch for now then we think about how to make things work in both cases?

tmaroschik:
If its breaking the multi-instance scenario, then maybe make the prefix configurable, or prefix the node with the home id as well? Otherwise the hame has to be sanitized so it doesn't contain any spaces or other special charcters.

tmaroschik:
@blhoward2 The node name itself is derived in hass from the json payload, not the mqtt topic. You could even drop the whole topic segment and things would still work.

tmaroschik:
@robertsLando In Hass, the node_id is used to form the discovery_id, which becomes the discovery_hash together with the component prefix. That tells the discovery that it has seen this object already and updates it. https://github.com/home-assistant/core/blob/f86beed7b047563c58cdddae71d55abde298ac85/homeassistant/components/mqtt/discovery.py#L136

It doesn't form any other identifier or name, visible in the UI. Throughout the flow, it is clear, that only the payload gets converted finally to the config, like here https://github.com/home-assistant/core/blob/99f9f8dec031d6e8d5f3f5443950d7980fceb739/homeassistant/components/mqtt/mixins.py#L154
In the end, you land back in the discovery module and the config entry gets added to Hass Data.

I'm no Python expert, so I might oversee some stuff. Do you know somebody, that could clarify this finally?

robertsLando:
Some guys from hass maybe @balloob @marcelveldt @MartinHjelmare ?

MartinHjelmare:
What is it that you want clarified?

robertsLando:
@MartinHjelmare how does the `node_id` part in the discovery topic is used by hass?

MartinHjelmare:
The whole discovery topic needs to be unique for Home Assistant to be able to distinguish between devices and its entities.

https://www.home-assistant.io/docs/mqtt/discovery/

It is recommended to make the `object_id` part of the topic a unique id.

tmaroschik:
So the current implementation is incomplete, as I can freely define names that are either, using not allowed chars or duplicates. 

Solutions options:
* use the location and name of the device to form a maybe unique identifier and pay attention to sanitisation
* use the home id and node id in conjunction

embcla:
Is the homeID unique and does it match the limitations on acceptable characters?

Personally I'd go for location-deviceName sanitised

blhoward2:
The homeId is long and sort of messy. It partially defeats the benefit of using the node name as the topic. I'd just sanitize it.

robertsLando:
so what? We could eventually use object_id as suggested by @MartinHjelmare , or the sanitized location + node name, just tell me what you prefer

blhoward2:
I personally think users like the location and name because it keeps it organized and human readable.

PR #981, opened 2021-03-31
fix: better sanitize topics

robertsLando:
Related to #920 

cc @blhoward2 @tmaroschik let me know your feedback on this please

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -666,14 +666,12 @@ function deviceInfo (node, nodeName) {
  *
  * @param {import('../types').HassDevice} hassDevice The Hass device object configuration
  * @param {string} nodeName Node name from getNodeName function
- * @param {string} location Node location
  * @returns The topic string for this device discovery
  */
-function getDiscoveryTopic (hassDevice, nodeName, location) {
-  return `${hassDevice.type}/${utils.sanitizeTopic(
-    nodeName + location ? '_' + location : '',
-    true
-  )}/${hassDevice.object_id}/config`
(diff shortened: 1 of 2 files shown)


tmaroschik (approved):
This change looks fine.

One thing that might cause problems down the road is, that `utils.sanitizeTopic()` doesn't remove special chars like umlauts or emojis. So instead of just removing "known" invalid chars, it would be better to remove all "unknown" invalid chars with a regex like `str.replace(/[^A-Za-z0-9-_]/g, "")` or even transliterate in advance to that if you want to go the extra mile.

The topic char constraints apparently aren't defined by the protocol spec itself (they basically allow every UTF-8 char, except slash/star/etc.), but rather by the clients and their capabilities (UTF-8 on an ESP consumes more memory than plain ASCII).

robertsLando:
> So instead of just removing "known" invalid chars, it would be better to remove all "unknown" invalid chars with a regex like str.replace(/[^A-Za-z0-9-_]/g, "") or even transliterate in advance to that if you want to go the extra mile.

You right, I think a regex like `/[^A-Za-z0-9-_]/g` should do the job!

robertsLando:
@tmaroschik I have also added the `/` char to allowed chars in the regex as it should not be removed by default.

tmaroschik (approved):
True. Looks good.

robertsLando:
Wait wait... by default nodeName alreadt contains location-name value, if `ignoreLocation` setting is false. What we should do so is to just sanitize nodename

PR #1055, opened 2021-04-13
fix: notification event not publiished to mqtt

### LINKED ISSUE #1044: [bug] Basic set events not sent to mqtt anymore?
### Version

**Checklist:**

- [X ] I am **not** using HomeAssistant. **Or:** a developer has told me to come here.
- [X ] I have checked the troubleshooting section and my problem is **not** described there.
- [ X] I have read the changelog problem was **not** mentioned there.

Build/Run method

- [x ] Docker
- [ ] PKG
- [ ] Snap package

zwavejs2mqtt: 3.2.0
zwave-js: 7.1.1

### Describe the bug

Until some unknown period in the past, basic set events from devices with a compat flag to undo the mapping and treat them as an event appeared under command class 32 under 32/0/event. After a complete rebuild of my network I no longer see 32 under any devices regardless of whether I use:

1. ValueID Topics
2. Named Topics (for this one basic isn't listed)

Double tapping my GE switches does send through the event properly (I can see it in HA) and I see it in the zwavejs1mqtt log. But I don't see any messages anywhere in mqtt.

With named topics:
Screen Shot 2021-04-12 at 12 18 13 AM

With ValueID topics:

Screen Shot 2021-04-12 at 12 42 51 AM

### Expected behavior

Double tap up/down should generate an event notification with a value of 255/0, and this should appear in mqtt.

### Additional context

zwave-js:
```
2021-04-12T04:38:31.385Z CNTRLR   [Node 002] treating BasicCC::Set as a value event
2021-04-12T04:38:31.385Z CNTRLR   [Node 002] [!] [Basic] event: 255                                 [Endpoint 0]
```

zwavejs2mqtt:

[node2.log](https://github.com/zwave-js/zwavejs2mqtt/files/6293936/node2.log)

robertsLando:
Fixes #1044 (maybe)

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -1186,6 +1186,8 @@ function updateValue (zwaveNode, args) {
           self.emit('valueChanged', valueId, node, false)
         }, 1000)
       }
+    } else {
+      addValue.call(this, zwaveNode, args)
     }
     node.lastActive = Date.now()
   }
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -619,6 +619,7 @@ function onNodeValueNotification (zwaveNode, args) {
   args.newValue = args.value
   // specify that this is stateless
   args.stateless = true
+
   onNodeValueUpdated.call(this, zwaveNode, args)
 }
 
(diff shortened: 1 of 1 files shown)


lib/ZwaveClient.js
```diff
@@ -1186,6 +1186,8 @@ function updateValue (zwaveNode, args) {
           self.emit('valueChanged', valueId, node, false)
         }, 1000)
       }
+    } else {
+      addValue.call(this, zwaveNode, args)
```
AlCalzone, line 1190: Wouldn't you want to apply the timeout logic here too?

robertsLando: What? This just add the value to values so it will also be displayed on UI

AlCalzone: Yeah, but in valueChanged, you reset the value to undefined after one second. I don't see this happening here.

robertsLando: right, fixed

PR #1060, opened 2021-04-15
fix: prevent 404 errors in store when using nginx

robertsLando:
Fixes https://github.com/hassio-addons/addon-zwavejs2mqtt/issues/4#issuecomment-819371878

/cc @frenck


frenck:
will set up a local test with this branch 👍

frenck:
Yep, works, instantly 🎉  👍 ❤️

PR #1095, opened 2021-04-22
feat: support managing associations on endpoints

robertsLando:
```ts
// Deprecated overload:
isAssociationAllowed(nodeId: number, group: number, destination: AssociationAddress): boolean;
// Use this one instead
isAssociationAllowed(source: AssociationAddress, group: number, destination: AssociationAddress): boolean;

// Deprecated overload:
addAssociations(nodeId: number, group: number, destinations: AssociationAddress[]): Promise<void>;
// Use this one instead
addAssociations(source: AssociationAddress, group: number, destinations: AssociationAddress[]): Promise<void>;

// Deprecated overload:
removeAssociations(nodeId: number, group: number, destinations: AssociationAddress[]): Promise<void>;
// Use this one instead
removeAssociations(source: AssociationAddress, group: number, destinations: AssociationAddress[]): Promise<void>;

// Deprecated overload:
getAssociationGroups(nodeId: number): ReadonlyMap<number, AssociationGroup>;
// Use this one instead
getAssociationGroups(source: AssociationAddress): ReadonlyMap<number, AssociationGroup>;

// Deprecated overload:
getAssociations(nodeId: number): ReadonlyMap<number, readonly AssociationAddress[]>;
// Use this one instead
getAssociations(source: AssociationAddress): ReadonlyMap<number, readonly AssociationAddress[]>;

// Deprecated type
interface Association {
	nodeId: number;
	endpoint?: number;
}
// Use this identical one instead
interface AssociationAddress{
	nodeId: number;
	endpoint?: number;
}
```

Ref: 
https://github.com/zwave-js/node-zwave-js/pull/2287
https://github.com/zwave-js/node-zwave-js/pull/2460

Schermata da 2021-04-27 17-25-12

Schermata da 2021-04-27 17-25-46

BREAKING CHANGE: Signature of methods `getAssociations`, `addAssociations` and `removeAssociations` have changed. This will have no effect on normal users but for the ones that are using those apis via MQTT remember to check the changes in signature as now you also have to specify source endpoint instead of just the nodeId


lib/ZwaveClient.js
```diff
@@ -1571,22 +1571,27 @@ ZwaveClient.prototype.getGroups = async function (
...
-        text: group.label,
-        value: groupIndex,
-        maxNodes: group.maxNodes,
-        isLifeline: group.isLifeline,
-        multiChannel: group.multiChannel
-      })
+    for (const [endpoint, groups] of endpointGroups) {
+      for (const [groupIndex, group] of groups) {
+        // https://zwave-js.github.io/node-zwave-js/#/api/controller?id=associationgroup-interface
+        node.groups.push({
```
robertsLando, line 1586: @AlCalzone I think this could have some dups when there are multiple endpoints, right?

src/components/nodes-table/AssociationGroups.vue
```diff
@@ -197,7 +211,7 @@ export default {
     getSourceAddress () {
       return {
         nodeId: this.group.node.id,
-        endpoint: this.group.nodeEndpoint || 0
+        endpoint: this.group.group.endpoint || 0
```
AlCalzone: I know we talked about zwave-js not caring about missing endpoint vs endpoint 0. However, that was for value IDs. Associations are different, because endpoint 0 gets treated as a Multi Channel Association, while missing endpoint gets treated as a "classic" node Association.

AlCalzone (changes requested):

src/components/nodes-table/AssociationGroups.vue
robertsLando: Yeah my question was about this 😆 No problem BTW now I fix it :)

lib/ZwaveClient.js
```diff
@@ -1585,7 +1585,7 @@ ZwaveClient.prototype.getGroups = async function (
         // https://zwave-js.github.io/node-zwave-js/#/api/controller?id=associationgroup-interface
         node.groups.push({
           text: group.label,
-          endpoint: endpoint || 0,
+          endpoint: endpoint, // when wndpoint is undefined Classic association is used instead of multi instance association
```
AlCalzone, line 1588: what's a `wndpoint`? :D

AlCalzone: Btw, this `endpoint` variable shouldn't ever be `undefined`.

AlCalzone:
FYI, I've updated the PR text explaining which signatures changed. And they are all still supported, just deprecated and should be phased out from usage before 8.0.0

robertsLando:
> FYI, I've updated the PR text explaining which signatures changed

It's ok for me btw I will release a 4.0.0 version anyway when this is done

PR #1134, opened 2021-05-04
fix(hass): sending `true` to MultilevelSwitchCC doesn't restore old level

### LINKED ISSUE #1132: [bug] Sending ON to Multilevel Switch sets value to max instead of Previous State
### Version

**Checklist:**

- [ X ] I am **not** using HomeAssistant. **Or:** a developer has told me to come here.
- [ X ] I have checked the troubleshooting section and my problem is **not** described there.
- [ X ] I have read the changelog problem was **not** mentioned there.

Build/Run method

- [ ] Docker
- [ ] PKG
- [ X ] Snap package
- [ X ] Manually built (git clone - npm install - npm run build )

zwavejs2mqtt: 3.5.0.42d15cb
zwave-js: 7.2.4

### Describe the bug

Sending an ON event to a multilevel switch (Inovelli LZW31-SN) is setting the switch to 99 instead of the last state. 

### To Reproduce

zwave/dimmer_1/38/0/targetValue/set 50
zwave/dimmer_1/38/0/targetValue/set off
zwave/dimmer_1/38/0/targetValue/set on

### Expected behavior

I expect that the device would return to 50% when turned on, but instead a value of 99 is sent to the device.

### Additional context

I believe the issue is:
```javascript
    //Gateway.js
    // on/off becomes 100%/0%
    if (typeof payload === 'boolean' && valueId.type === 'number') {
      payload = payload ? valueId.max : valueId.min
    }
```
If I manually build and change that file to:

```javascript
    //Gateway.js
    // on/off becomes 100%/0%
    if (typeof payload === 'boolean' && valueId.type === 'number') {
      payload = payload ? 0xff : valueId.min
    }
```
It works as I would expect. I'm not familiar enough with other zwave switches to know if inovelli is alone in sending a 255 to the Multilevel Switch channel turns the switch back to the previous state or if it is a standard method. If it is an edge case a way to manually define the "on" event would be helpful.

sovapatr:
Fixes #1132

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -969,7 +969,7 @@ Gateway.prototype.parsePayload = function (payload, valueId, valueConf) {
 
     // on/off becomes 100%/0%
     if (typeof payload === 'boolean' && valueId.type === 'number') {
-      payload = payload ? valueId.max : valueId.min
+      payload = payload ? 0xff : valueId.min
     }
 
     if (valueId.commandClass === CommandClasses['Binary Toggle Switch']) {


robertsLando (approved):
LGTM

PR #1136, opened 2021-05-04
feat(docker): push image to ghcr

onedr0p:
👋🏼 Hi @robertsLando here's a PR to publish the Docker image to Github container registry and keeps dockerhub.

The reason for this is Dockerhub has pretty serious rate limits (100 per 6 hours). Having and additional Docker registry would be awesome and it's so simple with GHCR.

This change requires a PAT being created and secrets in teh repo settings for:

- `GHCR_USERNAME`
- `GHCR_TOKEN`

After the image is pushed you will need to make the image public in the Github package.

Thanks for the consideration, hoping you find this useful.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -21,10 +21,16 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Login to GHCR
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ secrets.GHCR_USERNAME }}
+          password: ${{ secrets.GHCR_TOKEN }}
       - id: docker_meta
         uses: crazy-max/ghaction-docker-meta@v1
         with:
-          images: zwavejs/zwavejs2mqtt
+          images: zwavejs/zwavejs2mqtt,ghcr.io/zwavejs/zwavejs2mqtt
           tag-sha: true
           tag-latest: true
           tag-semver: |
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-test.yml
```diff
@@ -23,11 +23,18 @@ jobs:
...
+          username: ${{ secrets.GHCR_USERNAME }}
+          password: ${{ secrets.GHCR_TOKEN }}
+
       - name: Create Docker Meta
         id: docker_meta
         uses: crazy-max/ghaction-docker-meta@v1
         with:
-          images: zwavejs/zwavejs2mqtt
+          images: zwavejs/zwavejs2mqtt,ghcr.io/zwavejs/zwavejs2mqtt
           tag-sha: false
```
robertsLando, line 38: @onedr0p Not needed here, test images are not pushed so it is useless. Revert changes to this file

.github/workflows/docker-release.yml
```diff
@@ -21,10 +21,16 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Login to GHCR
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ secrets.GHCR_USERNAME }}
+          password: ${{ secrets.GHCR_TOKEN }}
```
robertsLando, line 29: For security reasons I prefer to avoid setting my secrets here. A better approach would be to use: 
```suggestion
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
```

robertsLando: This workflow is always triggered by the bot that already has read/write access to this repo

robertsLando (changes requested):

.github/workflows/docker-test.yml
onedr0p: Updated

robertsLando (commented):
LGTM

PR #1139, opened 2021-05-05
chore: switch to yarn

robertsLando:
/CC @AlCalzone


.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         uses: actions/cache@v2.1.4
         with:
           path: node_modules
-          key: ${{ matrix['node-version'] }}-yarn-${{ hashFiles('package-lock.json') }}
```
AlCalzone: You might want to do this instead to cache and install deps:
https://github.com/zwave-js/node-zwave-js/blob/efe8d4a8c843b1376db1e8ac2873c66476e14a43/.github/workflows/test-and-release.yml#L34-L48

docker/Dockerfile.contrib
```diff
@@ -24,21 +24,21 @@ USER root
 RUN apt-get update && apt-get install -y jq
 USER node
 WORKDIR /home/node/node-zwave-js
-RUN rm -f package-lock.json
+RUN rm -f yarn.lock
```
AlCalzone, line 27: Do you really need to remove the lockfile? This will cause pain when someone on the contrib image wants to update configs.

AlCalzone (changes requested):

docker/Dockerfile.contrib
robertsLando: They was removing package-lock.json, dunno the reason. I just did that to keep it unchanged

.github/workflows/ci.yml
robertsLando: essentially I should change the cache key to that one and use `yarn install --prefer-offline --frozen-lockfile`?

AlCalzone: Getting the cache key and installing the node modules should be changed to the lines I linked. That's what I've found to be best practise for caching yarn installs on CI

AlCalzone: So all of this:
grafik

And maybe even edit the cache key for your cached build?

robertsLando: done

PR #1142, opened 2021-05-06
chore(docker): push to github packages

chrisns:


robertsLando (approved):
Seems good to me, just dunno how the push will use the correct tags based on the different images

PR #1146, opened 2021-05-06
fix: use `getNeighbors` method

robertsLando:


AlCalzone (changes requested):
I don't think this makes sense - the entire reason why I'm going to remove the neighbors property is because it is almost impossible to keep up to date. Moving nodes, including/excluding nodes, pulling plugs, etc. - all these can change the neighbors a node sees. And it all happens on the Z-Wave protocol level, so we don't get notified.

AFAIK you only use it when generating the map, so before doing that, you should request all neighbors for all nodes and use this fresh information for the map.

robertsLando:
> AFAIK you only use it when generating the map, so before doing that, you should request all neighbors for all nodes and use this fresh information for the map.

AlCalzone:
> keeping them on nodes is not so wrong

It may lead someone to believe that the node property is reliable though - maybe even get cached somewhere. If there's only a method to retrieve the current list, it is much clearer that it has to be retrieved fresh.

> I could show a map and then once the refresh ends I update it?

I don't see where it is used in the frontend. However I don't think it is bad UX to request the neighbors and show the map with up2date neighbors 200ms later.

robertsLando:
> I don't see where it is used in the frontend. However I don't think it is bad UX to request the neighbors and show the map with update neighbors 200ms later.

Ok so I will remove `neighbors` proeprty

PR #1195, opened 2021-05-12
fix: battery devices not ready

robertsLando:
Fixes https://github.com/zwave-js/node-zwave-js/issues/2600


AlCalzone:
Looks better, but the dead node's info is still not shown, although it is known:
grafik

bushvin:
Thank you fr fixing it. Just tried it and I have all my info back!

PR #1211, opened 2021-05-13
chore(docker): add HEALTHCHECK

### LINKED ISSUE #1210: [feat] Built in healthcheck?
I saw that in the guide you could set up a health check using curl and upon doing more research on how to do that with a docker run command I saw that it could be done inside the container with a simple script.

Add `HEALTHCHECK` to the docker file with a script to update the status.

**Additional context**
Example with .js file: https://blog.sixeyed.com/docker-healthchecks-why-not-to-use-curl-or-iwr/

robertsLando:
Fixes #1210

Reference for docker compose: https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck

### DIFF
--- a/healthcheck.js
+++ b/healthcheck.js
@@ -1,0 +1,39 @@
+const http = require('http')
+
+const availableClients = {
+  all: '',
+  zwave: '/zwave',
+  mqtt: '/mqtt'
+}
+
+let clientCheck = availableClients.all
+
+if (process.env.HEALTH_CLIENT) {
+  for (const c in availableClients) {
+    if (c === process.env.HEALTH_CLIENT) {
+      clientCheck = availableClients[c]
+    }
+  }
+}
+
+const options = {
+  port: require('./config/app.js').port,
+  timeout: 2000,
+  path: '/health' + clientCheck
+}
+
+const request = http.request(options, res => {
+  console.log(`STATUS: ${res.statusCode}`)
+  if (res.statusCode === 200) {
+    process.exit(0)
+  } else {
+    process.exit(1)
+  }
+})
+
+request.on('error', function (err) {
+  console.log('ERROR: ' + err.message)
+  process.exit(1)
+})
+
+request.end()
--- a/healthcheck.js
+++ b/healthcheck.js
@@ -12,6 +12,7 @@ if (process.env.HEALTH_CLIENT) {
   for (const c in availableClients) {
     if (c === process.env.HEALTH_CLIENT) {
       clientCheck = availableClients[c]
+      break
     }
   }
 }
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -64,7 +64,7 @@ ENV HEALTH_CLIENT=zwave
 WORKDIR /usr/src/app
 
 HEALTHCHECK --interval=12s --timeout=12s --start-period=30s \  
-    CMD node /healthcheck.js
+    CMD node healthcheck.js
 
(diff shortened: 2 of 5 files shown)


healthcheck.js
```diff
@@ -0,0 +1,40 @@
...
+      clientCheck = availableClients[c]
+      break
+    }
+  }
+}
+
+const options = {
+  port: require('./config/app.js').port,
+  timeout: 2000,
+  path: '/health' + clientCheck
```
chrisns, line 23:
```suggestion
  path: `/health${clientCheck}`
```

healthcheck.js
```diff
@@ -0,0 +1,40 @@
...
+    if (c === process.env.HEALTH_CLIENT) {
+      clientCheck = availableClients[c]
+      break
+    }
+  }
+}
+
+const options = {
+  port: require('./config/app.js').port,
+  timeout: 2000,
```
chrisns, line 22: you've a 2 sec timeout here but a 12sec timeout in the healthcheck itself?

https://github.com/zwave-js/zwavejs2mqtt/pull/1211/files#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcR66

chrisns (changes requested):
the only thing to really consider with this is if you're setting up for the first time, and using the gui to point at your zwave device, its possible the healthcheck will mark your container unhealthy before you finish setting it up.
I suspect you want to check first somehow that it is expected to be able to work
you may also have a 'none' client that doesn't check any of them perhaps

robertsLando:
@chrisns You right, didn't think about that... What's your suggestion? Maybe remove the hardcoded healtcheck and just document it in docker compose?

healthcheck.js
robertsLando: In case the healthcheck stucks for any other reason I dind't think about 😆 But I could set the two timeouts to be equal, what do you think it's best?

chrisns:
yeah, I guess so :( or add a config option for it maybe ¯\_(ツ)_/¯

ZetaPhoenix:
Is the healthcheck during setup an issue? Other than the status being set as unhealthy (which makes sense as it is not running as expected) I do not see any negative side affects by default but I will admit I have limited experience here.

chrisns:
@ZetaPhoenix many setups like docker-compose would restart it, you may also find that it wouldn't be included in load balancer configurations where an ingress controller is present

robertsLando:
>  I do not see any negative side affects by default but I will admit I have limited experience here.

The server that keeps restarting it's an issue...

robertsLando:
@chrisns Do you think the actual solution is good? Essentially everything is disabled by default, users can choose to enable it when needed

chrisns:
¯\_(ツ)_/¯
i'm not really sure what we're trying to achieve + why :)

robertsLando:
@chrisns A user asked me to add healthcheck to dockerfile in #1210, I just proposed this as a solution

chrisns:
probably doesn't help things I really strongly disagree with the article they cite :) https://blog.sixeyed.com/docker-healthchecks-why-not-to-use-curl-or-iwr/
yeah this works as a solution, though it's a bit ugly, the reasons in the article don't really stack up IMO and curl remains easier
maintaining curl is installed seems wayyy easier than maintaining a bit of custom code :)

robertsLando:
¯_(ツ)_/¯

No idea what to do so

PR #1212, opened 2021-05-13
chore: support dotenv files

Glenf:
When developing UI against remote backend the common variables can be saved to .env file. This change adds support for that.


robertsLando (approved):

.env.example
```diff
@@ -0,0 +1,8 @@
+# The port the server is running. Default is `8091`
+SERVER_PORT=8091
+# Set to 'true' if server is using HTTPS / WSS scheme
+SERVER_SSL=true
+# The host address the server is binded to. Default is `0.0.0.0`
+SERVER_HOST=0.0.0.0
+# Complete URL to server. If not set a combination of `SERVER_SSL`, `SERVER_HOST` and `SERVER_PORT` will be used.
+SERVER_URL="http://localhost:8091"
```
robertsLando, line 8: Could you reneme this file to `.env.webpack` and also create a `.env.app` with application env vars? https://zwave-js.github.io/zwavejs2mqtt/#/guide/env-vars

robertsLando (changes requested):

.env.example
Glenf: So you'd want to have separate env file for frontend and backend application? To what file should the backend configuration be required?

robertsLando: > So you'd want to have separate env file for frontend and backend application

Yes. Backend file should be required on `bin/www` file, it's the main entry point

config/app.js
```diff
@@ -1,4 +1,6 @@
 const { joinPath } = require('../lib/utils')
+require('dotenv').config({path: './.env.app'})
```
Glenf: Is this correct place to require dotenv? The variables below seem to change correctly when placed in .env.app file.

robertsLando (approved):
LGTM

PR #1230, opened 2021-05-17
fix: allow `sendCommand` to accept string command classes

### LINKED ISSUE #1219: [feat] Modify sendCommand to accept command class names in addition to CC IDs
Request to modify sendCommand so it'll accept the command class human readable names ("Multilevel Switch") in addition/in lieu of their command class numbers.

See discussion in #1196 .  

(copy pasta from there)

@robertsLando I think it would be good if the mqtt API not only accepted the CC IDs (numbers) but also their names as defined in node-zwave-js. So for a startLevelChange the command could look like this:

{ "args": [
  {
    "nodeId": 23,
    "commandClass": "Multilevel Switch",
    "endpoint": 0,
  },
  "startLevelChange",
]}

robertsLando:
Fixes #1219

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -1425,7 +1425,7 @@ class ZwaveClient extends EventEmitter {
   /**
    * Send a command
    *
-   * @param {{nodeId: number, endpoint: number, commandClass: number}} ctx context to get the instance to send the command
+   * @param {{nodeId: number, endpoint: number, commandClass: number | string}} ctx context to get the instance to send the command
    * @param {string} command command name
    * @param {any[]} args args to pass to `command`
    * @returns {Promise<any>}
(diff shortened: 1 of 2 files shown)


lib/ZwaveClient.js
```diff
@@ -1425,7 +1425,7 @@ class ZwaveClient extends EventEmitter {
   /**
    * Send a command
    *
-   * @param {{nodeId: number, endpoint: number, commandClass: number}} ctx context to get the instance to send the command
+   * @param {{nodeId: number, endpoint: number, commandClass: number | string}} ctx context to get the instance to send the command
```
AlCalzone, line 1428: The correct type for `commandClass` would be: `CommandClasses | keyof typeof CommandClasses`, where `CommandClasses` is the enum from `zwave-js`.

lib/ZwaveClient.js
```diff
@@ -1454,8 +1454,12 @@ class ZwaveClient extends EventEmitter {
         )
       }
 
+      const commandClass = isNaN(ctx.commandClass)
+        ? CommandClasses[ctx.commandClass]
+        : ctx.commandClass
```
AlCalzone, line 1459: I'd turn this around:
```js
      const commandClass = typeof ctx.commandClass === "number"
        ? ctx.commandClass
        : CommandClasses[ctx.commandClass]
```

robertsLando:
@AlCalzone Implemented in a new PR, thanks

PR #1259, opened 2021-05-21
fix: detect driver failed and restart

### LINKED ISSUE #1246: [bug] zwavejs2mqtt does not reconnect if the serial port (TCP socket) is closed
### Version

**Checklist:**

- [x] I am **not** using HomeAssistant. **Or:** a developer has told me to come here.
- [x] I have read the changelog problem was **not** mentioned there.

Build/Run method

- [x] Docker
- [ ] PKG
- [ ] Snap package

zwavejs2mqtt version: 4.3.0
zwave-js version: 7.5.0

### Describe the bug

In #1167, a problem with detecting a dropped TCP socket connection to the serial device was reported and fixed in https://github.com/zwave-js/node-zwave-js/pull/2601. However, the application needs to handle the lost connection. For now, the state is "Driver: The socket closed unexpectedly". IMHO, zwavejs2mqtt should try to auto-reconnect in this case, as the connection might become available (this could also be a configuration option, like auto-reconnect interval).

It finally reconnects if I go to Settings and click on "Save".

### To Reproduce

1. Connect the Z-Wave dongle via ser2net
2. Stop ser2net
3. Wait for the lost connection in zwavejs2mqtt
4. Restart ser2net

### Expected behavior

After 4., the connection should be auto-restored.

robertsLando:
Fixes #1246
Fixes https://github.com/zwave-js/node-zwave-js/issues/2609

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -1760,9 +1760,10 @@ class ZwaveClient extends EventEmitter {
     this.error = 'Driver: ' + error.message
     this.status = ZWAVE_STATUS.driverFailed
     this._updateControllerStatus(this.error)
+    this.emit('event', eventEmitter.driver, 'driver error', error)
 
+    // // this cannot be recovered by zwave-js, requires a manual restart
     if (error instanceof ZWaveErrorCodes.Driver_Failed) {
-      // this cannot be recovered by zwave-js, requires a manual restart
       try {
         await this.close(true)
         this.closed = false
@@ -1771,8 +1772,6 @@ class ZwaveClient extends EventEmitter {
         logger.error(`Error while restarting driver: ${error.message}`)
       }
     }
-
-    this.emit('event', eventEmitter.driver, 'driver error', error)
   }
 
   _onScanComplete () {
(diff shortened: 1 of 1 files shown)


lib/ZwaveClient.js
```diff
@@ -1750,11 +1756,22 @@ class ZwaveClient extends EventEmitter {
     logger.info(`Scanning network with homeid: ${homeHex}`)
   }
 
-  _onDriverError (error) {
+  async _onDriverError (error) {
     this.error = 'Driver: ' + error.message
     this.status = ZWAVE_STATUS.driverFailed
     this._updateControllerStatus(this.error)
 
+    if (error instanceof ZWaveErrorCodes.Driver_Failed) {
```
AlCalzone, line 1764:
```suggestion
    if (error.code === ZWaveErrorCodes.Driver_Failed) {
```

AlCalzone: Error is `instanceof ZWaveError`

AlCalzone (changes requested):

lib/ZwaveClient.js
robertsLando: Right...

PR #1279, opened 2021-05-26
fix: don't query neighbors on node ready event

robertsLando:
CC @AlCalzone. Fixes https://github.com/zwave-js/node-zwave-js/issues/2493

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -1962,7 +1962,6 @@ class ZwaveClient extends EventEmitter {
     node.values = {}
 
     this._dumpNode(zwaveNode)
-    node.neighbors = await this.getNodeNeighbors(zwaveNode.id, true)
 
     const values = zwaveNode.getDefinedValueIDs()
 
--- a/src/components/Mesh.vue
+++ b/src/components/Mesh.vue
@@ -150,15 +150,9 @@ export default {
         switch (data.api) {
           case 'refreshNeighbors': {
             this.showSnackbar('Nodes Neighbors updated')
-            const neighbors = data.result
-            for (const nodeId in neighbors) {
-              this.setNeighbors({
-                nodeId: nodeId,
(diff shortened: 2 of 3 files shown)


AlCalzone:
You still refresh this when opening the map, right?

robertsLando:
@AlCalzone Yep

ualex73:
Just a question :-) There is a corruption of the cache file, caused by the neighhors query - the "fix" seems to change the query moment only. Does this really fix any possible cache corruption?

robertsLando:
@ualex73 Based on what @AlCalzone said to me seems the problem is that I actually query neighbors when nodes are ready. In large networks this could cause problems as 

> the routing table is in the so-called NVM. Accessing that requires to stop the stick from sending/receiving while you do it (@AlCalzone)

In poor words requesting neighbors could cause congestion on the controller expecially if it is doing interviews on startup or other things, with this fix neighbors are only refreshed when network graph is shown on the ui (or the api `refreshNeighgbors` is called).

Another fix in future will be to turn off the radio before query neighbors and turn on it once done. This needs https://github.com/zwave-js/node-zwave-js/pull/2636

ualex73:
Ok, I understand it would lock up the stick if such query happens at start-up. Only if such situation happens (e.g. no response from stick during start-up), then it should never corrupt the cache, which still could happen? Just with this fix, the change will go towards to <1%.

robertsLando:
I think the problem is not the corrupted cache but I  let speak @AlCalzone about this. It's not a complete fix but could save some corruptions

AlCalzone:
Normally during the startup, listening nodes are pinged to figure out if they are alive or dead, then they are marked "ready". zwavejs2mqtt uses this event to display the information we already know, but that doesn't happen if the stick gets stuck before that. This looks like a corrupted cache, but the info is just not shown.

robertsLando:
Second part in #1283

PR #1283, opened 2021-05-27
fix: stop RF when requesting neighbors

robertsLando:
Reference: https://github.com/zwave-js/node-zwave-js/issues/2493
Introduced by: https://github.com/zwave-js/node-zwave-js/pull/2636

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -645,9 +645,7 @@ class ZwaveClient extends EventEmitter {
    */
   async refreshNeighbors () {
     if (this._lockNeighborsRefresh === true) {
-      throw Error(
-        'Cannot refresh neighbors now, you can do this once every 60 seconds'
-      )
+      throw Error('you can refresh neighbors only once every 60 seconds')
     }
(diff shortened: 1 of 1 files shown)


AlCalzone:
Do you have some kind of safeguard in place so this method isn't invoked too often? Especially overlapping calls should be avoided, but I think you should take care not to turn the radio off too often.

robertsLando:
@AlCalzone Actually nope but I can add it, What could be a safe time between calls of this method?

10 seconds?

AlCalzone:
I mean the neighbors shouldn't change that often and you're caching them anyways. I dunno - maybe max. once per minute and reset the lock on the call when a node was added or removed?

ualex73:
Question: if you turn off the radio and you restart ZwaveJS, does the stick automatically start receiving again? Just curious if a "problem" (app dies?) happen when turning the radio off and it should automatically recover with a restart.

AlCalzone:
I'm not sure tbh. Guess we have to try 😅

PR #1284, opened 2021-05-27
docs: improve bug report

robertsLando:


AlCalzone:
Need to check if my bug report template needs an update, but LGTM

PR #1296, opened 2021-05-31
docs: fix example device config loaded log

robertsLando:


AlCalzone (commented):
Wouldn't it be better to have examples for both, so users understand there are two possible outcomes?

robertsLando:
Humm not sure, I think it's useless. I mean if a user see this it means it's from his priority dir, if not it's the embedded one

PR #1297, opened 2021-05-31
refactor: switch to typescript

robertsLando:


app.ts
```diff
@@ -0,0 +1,1160 @@
...
+
+/**
+ * Binds socketManager to `server`
+ *
+ * @param {HttpServer} server
+ */
+function setupSocket(server: HttpServer) {
+	socketManager.bindServer(server)
+
+	socketManager.on(inboundEvents.init, function (socket: Socket) {
```
AlCalzone, line 443: try to avoid inline function type declarations - let TS infer them from the method signature that has the callback. (Example see Slack)

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+				error: gw.zwave.error,
+				cntStatus: gw.zwave.cntStatus,
+			})
+		}
+	})
+
+	socketManager.on(
+		inboundEvents.zwave,
+		// eslint-disable-next-line @typescript-eslint/no-misused-promises
+		async function (socket: Socket, data) {
```
AlCalzone, line 457: try to avoid inline function type declarations

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+		}
+
+		socket.emit(socketEvents.api, result)
+	})
+
+	// eslint-disable-next-line @typescript-eslint/no-misused-promises
+	socketManager.on(inboundEvents.hass, async function (socket, data) {
+		logger.info(`Hass api call: ${data.apiName}`)
+
+		let res: any, err: any
```
AlCalzone, line 506: ewww, `any` :)

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+		throw Error('Invalid token header')
+	}
+	const decoded = await verifyJWT(token, sessionSecret)
+
+	// Successfully authenticated, token is valid and the user _id of its content
+	// is the same of the current session
+	const users = jsonStore.get(store.users)
+
+	const user = users.find(
+		(u: { username: any }) => u.username === decoded.username
```
AlCalzone, line 584: try to avoid inline function type declarations

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+		try {
+			// token auth, mostly used to restore sessions when user refresh the page
+			if (token) {
+				const decoded = await verifyJWT(token, sessionSecret)
+
+				// Successfully authenticated, token is valid and the user _id of its content
+				// is the same of the current session
+				const users = jsonStore.get(store.users)
+
+				user = users.find((u: User) => u.username === decoded.username)
```
AlCalzone, line 640: try to avoid inline function type declarations

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+			} else {
+				// credentials auth
+				const users = jsonStore.get(store.users)
+
+				const username = req.body.username
+				const password = req.body.password
+
+				user = users.find((u: User) => u.username === username)
```
AlCalzone, line 648: try to avoid inline function type declarations

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+	}
+)
+
+app.get('/health', apisLimiter, function (req, res) {
+	let mqtt: Record<string, any> | boolean
+	let zwave: boolean
+
+	if (gw) {
+		mqtt = gw.mqtt ? gw.mqtt.getStatus() : false
+		zwave = gw.zwave ? gw.zwave.getStatus().status : false
```
AlCalzone, line 765: TypeScript supports [nullish coalescing and optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html). You could turn constructs like these into
```suggestion
		mqtt = gw.mqtt?.getStatus() ?? false
		zwave = gw.zwave?.getStatus().status ?? false
```

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+})
+
+app.get('/health/:client', apisLimiter, function (req, res) {
+	const client = req.params.client
+	let status: any
+
+	if (client !== 'zwave' && client !== 'mqtt') {
+		res.status(500).send("Requested client doesn 't exist")
+	} else {
+		status = gw && gw[client] ? gw[client].getStatus().status : false
```
AlCalzone, line 785: This could be
```suggestion
		status = gw?.[client]?.getStatus().status ?? false
```

config/store.ts
```diff
@@ -0,0 +1,21 @@
...
+  passwordHash: string,
+  token?: string
+}
+
+const store: { [key: string]: StoreFile } = {
+  settings : { file: 'settings.json', default: {} },
+  scenes : { file: 'scenes.json', default: [] },
+  nodes : { file: 'nodes.json', default: [] },
+  users : { file: 'users.json', default: [] }
+}
```
AlCalzone, line 19: There's really no nice way to achieve what you're doing here.
`{ [key: string]: StoreFile }` is better written as the type `Record<string, StoreFile>`, but this does not prevent you from accessing non-existant keys. You can hard-code the shape of an object using `as const`, but that does not let you say "the value is a StoreFile".

I'd probably do this:
```
type StoreKeys = "settings" | "scenes" | "nodes" | "users";

const store: Record<StoreKeys, StoreFile> = {
  settings : { file: 'settings.json', default: {} },
  scenes : { file: 'scenes.json', default: [] },
  nodes : { file: 'nodes.json', default: [] },
  users : { file: 'users.json', default: [] },
}
```

There is another trick which involves a one-time validation function and corresponding type declaration, but that's overkill here.

hass/configurations.ts
```diff
@@ -0,0 +1,205 @@
+// List of Home-Assistant configuration for MQTT Discovery
+// https://www.home-assistant.io/docs/mqtt/discovery/
+
+import { HassDevice } from '../lib/ZwaveClient'
+
+const configurations: { [key: string]: HassDevice } = {
```
AlCalzone, line 6: See comment in config/store.ts

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+	on(event: 'nodeRemoved', listener: (node: Z2MNode) => void): this
+	on(
+		event: 'valueChanged',
+		listener: (valueId: Z2MValueId, node: Z2MNode) => void
+	): this
+	on(
+		event: 'valueWritten',
+		listener: (valueId: Z2MValueId, value: any) => void
+	): this
+}
```
AlCalzone, line 319: You might want to define the other EventEmitter functions too:
https://github.com/zwave-js/node-zwave-js/blob/4b0ce9b613e63a683b25536456cb04ec7e0bb646/packages/zwave-js/src/lib/driver/Driver.ts#L324-L355

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+
+	driver: Driver
+
+	server: ZwavejsServer
+	statelessTimeouts: Record<string, NodeJS.Timeout>
+	commandsTimeout: NodeJS.Timeout
+	reconnectTimeout: NodeJS.Timeout
+	healTimeout: NodeJS.Timeout
+	updatesCheckTimeout: NodeJS.Timeout
+	pollIntervals: Record<string, NodeJS.Timeout>
```
AlCalzone, line 346: class properties should have `public`, `protected` or `private` modifiers. This doesn't really do anything during runtime, but clearly states the intent of visibility.
Same goes for all methods.

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+		this.driverInfo = {}
+		this.healTimeout = null
+
+		this.status = ZwaveClientStatus.CLOSED
+	}
+
+	/**
+	 * Restart client connection
+	 *
+	 * @returns {Promise<void>}
```
AlCalzone, line 408: unnecessary

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+			this.sendToSocket(socketEvents.nodeUpdated, node)
+		}
+	}
+
+	/**
+	 * Used to update hass devices list of a specific node and store them in `nodes.json`
+	 *
+	 */
+	async storeDevices(
+		devices: { [key: string]: HassDevice },
```
AlCalzone, line 560: `Record<string, HassDevice>`

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+		}
+	}
+
+	/**
+	 * Method used to close client connection, use this before destroy
+	 *
+	 * @param {boolean} keepListeners Prevents to remove all ZwaveCLient listeners (used when restarting)
+	 * @memberof ZwaveClient
+	 */
+	async close(keepListeners = false) {
```
AlCalzone, line 590: move type from JSDoc into function signature

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+			config: this.cfg,
+		}
+
+		return status
+	}
+
+	/**
+	 * Popolate node `groups`
+	 *
+	 * @param {number} nodeId Zwave node id
```
AlCalzone, line 661:
```suggestion
	 * @param nodeId Zwave node id
```
I won't comment on this anymore.

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+								nodeId: a.nodeId,
+								targetEndpoint: a.endpoint,
+							} as Z2MGroupAssociation)
+						}
+					}
+				}
+			} catch (error) {
+				logger.warn(
+					`Error while looking for Node ${nodeId}
+          associations: ${error.message}`
```
AlCalzone, line 727: Is this a 10-space indentation on purpose?

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+		return toReturn
+	}
+
+	/**
+	 * Get neighbors of a specific node
+	 */
+	async getNodeNeighbors(
+		nodeId: number,
+		dontThrow: boolean
+	): Promise<Readonly<number[]>> {
```
AlCalzone, line 951: `Readonly<number[]>` should be `readonly number[]`

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+		} catch (error) {
+			logger.error(
+				`Node ${nodeId} error while updating Neighbors: ${error.message}`
+			)
+			if (!dontThrow) {
+				throw error
+			}
+		}
+
+		return neighbors
```
AlCalzone, line 964: Much simpler and without the duplicate type annotation:
```suggestion
		try {
			return await this.driver.controller.getNodeNeighbors(nodeId)
		} catch (error) {
			logger.error(
				`Node ${nodeId} error while updating Neighbors: ${error.message}`
			)
			if (!dontThrow) {
				throw error
			}
		}
```

AlCalzone: Your implementation can return `undefined` though (if an error happens and `dontThrow` is true). The return type does not reflect that.

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+	/**
+	 * Calls a specific `client` or `ZwaveClient` method based on `apiName`
+	 * ZwaveClients methods used are the ones that overrides default Zwave methods
+	 * like nodes name and location and scenes management.
+	 *
+	 * @param {string} apiName The api name
+	 * @param {any[]} args Array of arguments to use for the api call
+	 * @returns An object `{success: <success>, message: <message>, args: <args>, result: <the response>}`,  if success is false the message contains the error
+	 */
+	async callApi(apiName: string, ...args: any[]) {
```
AlCalzone, line 1806: Clearly advanced territory, but there is a way to define this in a typesafe way:
https://www.typescriptlang.org/play?#code/C4TwDgpgBAwghgGwQQQAoEkBKEDOBXBYAHgBUA+KAXigG8AoKKfAY2dxwC4pgAnPCADQMoAW3ZwA5hC45eASwB2EoYx64CwLiToBfKAB9awlmxycoAM0Q5BwsWcnSm8pbrp1mCOGagAxAPb+RowWgQAUcDwSAIwyLhIAlFwKeCIARhA8wYxQasB4PApQ0QDcwjrujGmREVEATMmpGTxJUGmBCBBwRfQ5uRD5hdx8EGWMFcLMiChgcqRQEAAewBAKACY4UACii554axBEANYQIP4WfoECUABEU0jIszdkZGEKcGJa1wB0v5ES5lQkQ+A0yOCIAX8AG0SABdMiteAPDDYfCEIjYQYKEjgQ6QmHwii9HJ5Ao9YR9EzsLi8fgqPqicRSLg3fxHG70vpqNGaKBhYAACzkm28UG6IASUPeYlhYV+33+OASFKgFXGbg8-gUsksgSoUAUEAA7pd-GEEmVmFqdXl9aF-N97jM5GEbvaObcbhbNdr-J1vgh-BIwnkLUA

lib/ZwaveClient.ts
```diff
@@ -0,0 +1,3276 @@
...
+
+	/**
+	 * Parse a zwave value into a valueID
+	 *
+	 * @param { ZWaveNode } zwaveNode
+	 * @param { TranslatedValueID } zwaveValue
+	 */
+	_parseValue(
+		zwaveNode: ZWaveNode,
+		zwaveValue: TranslatedValueID & { [x: string]: any },
```
AlCalzone, line 3014: What's this weird type change for?

lib/jsonStore.ts
```diff
@@ -0,0 +1,74 @@
...
+import * as utils from './utils'
+
+const logger = module('Store')
+
+/**
+Constructor
+**/
+class StorageHelper {
+	public store: any
+	public config: any
```
AlCalzone, line 16: `any`, really?

lib/utils.ts
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return true
+}
+
+/**
+ * Deep copy of an object
+ */
+export function copy(obj: any): any {
+	return JSON.parse(JSON.stringify(obj))
+}
```
AlCalzone, line 85: I hope you don't do this too often on large objects. There are more performant ways to deep copy.

lib/Constants.ts
```diff
@@ -0,0 +1,693 @@
...
+}
+
+interface IDeviceClassProps {
+	generic: string
+	specific: IGenericMap
+}
+
+interface IDeviceClass {
+	[genericClass: number]: IDeviceClassProps
+}
```
AlCalzone, line 31: This is just a `Record<number, IDeviceClassProps>`

lib/MqttClient.ts
```diff
@@ -0,0 +1,456 @@
...
+	): this
+	on(event: 'multicastRequest', listener: (payload: any) => void): this
+	on(
+		event: 'apiCall',
+		listener: (topic: string, apiNema: string, payload: any) => void
+	): this
+	on(event: 'connect', listener: () => void): this
+	on(event: 'brokerStatus', listener: (online: boolean) => void): this
+	on(event: 'hassStatus', listener: (online: boolean) => void): this
+}
```
AlCalzone, line 57: See my comment in (I think) app.ts

AlCalzone (commented):
Didn't look at everything, but there are some things I would do differently.

.eslintrc.js
```diff
@@ -1,30 +1,65 @@
...
+				sourceType: 'module',
+				project: './tsconfig.eslint.json',
+			},
+			extends: [
+				// https://github.com/vuejs/eslint-plugin-vue#priority-a-essential-error-prevention
+				// consider switching to `plugin:vue/strongly-recommended` or `plugin:vue/recommended` for stricter rules.
+				'plugin:@typescript-eslint/recommended',
+				'plugin:@typescript-eslint/recommended-requiring-type-checking',
+				// Enables eslint-plugin-prettier and displays prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
+				'plugin:prettier/recommended',
```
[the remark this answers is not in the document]
robertsLando: yeah, why?

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+import { createServer as createHttpsServer } from 'https'
+
+declare module 'express' {
+	interface Request {
+		user?: User
+	}
+}
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { createCertificate } = require('pem').promisified
```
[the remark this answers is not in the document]
robertsLando: Hummm it's not really the same, createCertificate is a member of promisified and seems promisified is not typed

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+const logger = loggers.module('App')
+
+const verifyJWT = promisify(jwt.verify.bind(jwt))
+
+const storeLimiter = rateLimit({
+	windowMs: 15 * 60 * 1000, // 15 minutes
+	max: 100,
+	handler: function (
+		req: any,
+		res: { json: (arg0: { success: boolean; message: string }) => void }
```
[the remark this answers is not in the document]
robertsLando: I think this is somethinig created by vs code infer type from context/usage function, I have missed this

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+	},
+})
+
+const loginLimiter = rateLimit({
+	windowMs: 60 * 60 * 1000, // keep in memory for 1 hour
+	max: 5, // start blocking after 5 requests
+	handler: function (
+		req: any,
+		res: { json: (arg0: { success: boolean; message: string }) => void }
+	) {
```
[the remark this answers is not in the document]
robertsLando: Yep same

lib/ZwaveClient.ts
robertsLando: There could be some additional properties in zwaveValue

lib/Constants.ts
robertsLando: Isn't this the same?

lib/utils.ts
robertsLando: Like what?

app.ts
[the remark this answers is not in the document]
robertsLando: Fixed by installing rate limiter types

.eslintrc.js
```diff
@@ -1,30 +1,65 @@
...
+			],
+			// required to lint *.vue files
+			plugins: ['@typescript-eslint'],
```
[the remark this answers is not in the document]
robertsLando: Yeah

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+})
+
+// apis response codes
+const RESPONSE_CODES = {
+	0: 'OK',
+	1: 'General Error',
+	2: 'Invalid data',
+	3: 'Authentication failed',
+	4: 'Insufficient permissions',
+}
```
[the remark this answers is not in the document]
robertsLando: Fixed :)

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+	socket: Socket & { user?: Record<string, any> },
+	next: (err?: unknown) => void
+) {
+	if (!isAuthEnabled()) {
+		next()
+	} else if (socket.handshake.query && socket.handshake.query.token) {
+		jwt.verify(
+			socket.handshake.query.token,
+			sessionSecret,
+			function (err: unknown, decoded: Record<string, any>) {
```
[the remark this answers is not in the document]
robertsLando: Fixed by adding jwt `@types`

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+
+// ### UTILS
+
+/**
+ * Start http/https server and all the manager
+ *
+ * @param {string} host
+ * @param {number} port
+ */
+export async function startServer(host: any, port: number | string) {
```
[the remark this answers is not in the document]
robertsLando: Removed all jsdocs in favor of ts signatures

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+		const bind =
+			typeof addr === 'string' ? 'pipe ' + addr : 'port ' + addr.port
+		logger.info(
+			`Listening on ${bind} host ${host} protocol ${
+				process.env.HTTPS ? 'HTTPS' : 'HTTP'
+			}`
+		)
+	})
+
+	server.on('error', function (error: any) {
```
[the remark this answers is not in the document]
robertsLando: Maybe, but the problem is it's missing `syscall` and `code` props 🤷🏼‍♂️

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+	reqPath = path.normalize(reqPath)
+
+	if (!reqPath.startsWith(storeDir) || reqPath === storeDir) {
+		throw Error('Path not allowed')
+	}
+
+	return reqPath
+}
+
+async function loadCertKey() {
```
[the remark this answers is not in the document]
robertsLando: Fixed

app.ts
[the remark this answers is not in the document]
robertsLando: I have created a custom error interface for that

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+				next()
+			}
+		)
+	} else {
+		next(new Error('Authentication error'))
+	}
+}
+
+let gw: Gateway // the gateway instance
+const plugins = []
```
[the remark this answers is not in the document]
robertsLando: Hummm nope, I have to create it

app.ts
```diff
@@ -0,0 +1,1160 @@
...
+function setupInterceptor() {
+	// intercept logs and redirect them to socket
+	const interceptor = function (
+		write: (buffer: string | Uint8Array, cb?: (err?: Error) => void) => void
+	) {
+		return function (...args: any[]): boolean {
+			socketManager.io.emit('DEBUG', args[0]?.toString())
+			return write.apply(process.stdout, args)
+		}
+	}
```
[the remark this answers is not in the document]
robertsLando: Nope just add a `.bind(process.stdout)` fixed it

app.ts
robertsLando: Fixed

app.ts
robertsLando: done

app.ts
robertsLando: Fixed

app.ts
robertsLando: done

app.ts
robertsLando: dopne

app.ts
robertsLando: Yep I didn't refactored it yet :) That's something I really missed on js

config/store.ts
robertsLando: Done! Thanks for the tip :)

hass/configurations.ts
robertsLando: Done

lib/ZwaveClient.ts
robertsLando: Humm wait but If I extend EventEmitter in my class I already have those methods. At least I see them now

lib/ZwaveClient.ts
robertsLando: Yep I know, as a first step I focused on make the things work then I planned to gradually fix this things

lib/ZwaveClient.ts
robertsLando: yep removed, missed some jsdocs now i have removed them all

lib/ZwaveClient.ts
robertsLando: Done

lib/ZwaveClient.ts
robertsLando: Fixed

lib/ZwaveClient.ts
robertsLando: Nopt working, see slack

lib/jsonStore.ts
robertsLando: fixed

lib/Constants.ts
```diff
@@ -325,7 +325,7 @@ export const _sensorMap: ISensorMap = {
 		86: 'particulate_matter',
 	},
 }
-export function sensorType(index: number): any {
+export function sensorType(index: number) {
 	const sensorType = {
```
AlCalzone, line 329: There's _probably_ a better type for this which also defines the props, but I don't think it's so important.

lib/Constants.ts
```diff
@@ -438,7 +438,7 @@ const _commandClassMap = {
 	0xf0: 'non_interoperable',
 }
 export function commandClass(cmd: number): string {
```
AlCalzone, line 440: You're basically duplicating the `CommandClasses` enum from `zwave-js` - not sure if there isn't a better way?

lib/CustomPlugin.ts
```diff
@@ -0,0 +1,24 @@
+import { Router } from 'express'
+import MqttClient from './MqttClient'
+import { ModuleLogger } from './logger'
+import ZwaveClient from './ZwaveClient'
+
+export type PluginContext = {
```
AlCalzone, line 6: This should be an interface, not a type

lib/CustomPlugin.ts
```diff
@@ -0,0 +1,24 @@
...
+	mqtt: MqttClient
+	app: Router
+	logger: ModuleLogger
+}
+
+export interface CustomPlugin {
+	zwave: ZwaveClient
+	mqtt: MqttClient
+	app: Router
+	logger: ModuleLogger
```
AlCalzone, line 17:
```suggestion
export interface CustomPlugin extends PluginContext {
```

lib/CustomPlugin.ts
```diff
@@ -0,0 +1,24 @@
...
+	name: string
+
+	// eslint-disable-next-line @typescript-eslint/no-misused-new
+	new (context: PluginContext): CustomPlugin
```
AlCalzone, line 21: This isn't correct. `CustomPlugin` defines the instance type, but the constructor lives on the prototype (the class itself). This definition makes it so that every plugin implementation literally needs a `new` function which is impossible to implement.

If you want to say a function can only accept a constructor of a certain type, this is the way to do it:
https://www.typescriptlang.org/play?#code/PTAEBEFcFtoT1AFzgBwKYGdQYPZIBZqgBGaANjgO4ECWAdgOaiU4BOA1hgFDLqgBalAIYA3NAGEyNNHUSgAvKABEg0RKkzESgNw9URALIBHRIknTZC5cdPnNOvXwBKOSIjSsrSl248PehjgAJpBkaAAyOAwMHl4GwaERUTGsDlwgtFgY+K5kQSREQnSg9O6sAGZCAMZoXGgAHihscqUelTWgAApkkAz04jiyDXIA3lygoABewmIAXAIz6haI46DQJojzNmYasqtCKCjzPmWrFNEeWwlhkResXAC+XHWNzSVDFdVE4pAYiDjQbq9eigYYyIJYIF9OgDIb1UarOhCaBoeZ-Vj0BirIKYRCsHBwAAUAEp5p18dAaBg0AAeEQ4GhBAB8j2eDSarDkAS6PWhsPRkCq-08ijoaGohKqg3c8LJvP60uGxIUTNAPz+AKh9F0z3KkDoQpog1AVVYaCE7i1dBpABVQfDwZD5TDBgKhWwmZLXXj5jbSWrfv9Ac7QGMJlK6H9QGa5KLxSbvaxCfqceV6Gh8kIsEU4MTdBMYwA6JEorzlHA4YhCVL50AZSD4UA4fAAGgKplidBwciqQjIZEyqyLOPRBJJtZjkFYxRjuieXCqZCzWAAYhWrSVoCgwijZFh1UGN2HQCWiIolA4JiO8WP-eSAVTafTGarjxNEPh8dQxdQAKKsfEkyUAw0A-YIT27TdtzQXd3CCQslDzVYngmaY1C8VQxDsWQlFALME0jRBa3WUw4g2bCtDwrAIz+WsDhQLwTj8KiCNos5klic94hCG4ONSFiaKI1kFyXDAsAAOW7ABBI9VgAfRCWAEHPBx51Nc1LWdQk1xwK08zrMAcHYBczQtNArUJSTEBk519IyDxAKAA

lib/MqttClient.ts
```diff
@@ -391,7 +389,7 @@ class MqttClient extends EventEmitter {
 	_onMessageReceived(topic: string, payload: Buffer) {
 		if (this.closed) return
 
-		let parsed: any = payload ? payload.toString() : payload
+		let parsed: string | number | Record<string, any> = payload?.toString()
```
AlCalzone, line 392: `payload?.toString()` can be `undefined` if `payload` is undefined!

lib/SocketManager.ts
```diff
@@ -9,41 +9,59 @@ const logger = module('Socket')
...
+	on(
+		event: inboundEvents.hass,
+		listener: (socket: Socket, data) => void
+	): this
+
+	on(
+		event: inboundEvents.mqtt,
+		listener: (socket: Socket, data) => void
+	): this
 }
```
AlCalzone: I suggest you take a look at https://github.com/zwave-js/node-zwave-js/blob/4d9d317edf5192bec654d229fb8bc86c1966df29/packages/shared/src/EventEmitter.ts#L4 and https://github.com/zwave-js/node-zwave-js/blob/4d9d317edf5192bec654d229fb8bc86c1966df29/packages/zwave-js/src/lib/driver/Driver.ts#L326-L340

lib/jsonStore.ts
```diff
@@ -11,15 +12,15 @@ const logger = module('Store')
...
-class StorageHelper {
-	public store: any
-	public config: any
+export class StorageHelper {
+	public store: Record<StoreKeys, any>
+	public config: Record<StoreKeys, StoreFile>
 
 	constructor() {
-		this.store = {}
+		this.store = {} as Record<StoreKeys, any>
```
AlCalzone, line 20: This shouldn't be necessary when you define the type next to the property.

lib/utils.ts
```diff
@@ -139,7 +147,7 @@ export function removeSlash(str: string | number): string {
 /**
  * Check if an object has a property
  */
-export function hasProperty(obj: any, prop: string): boolean {
+export function hasProperty(obj, prop: string): boolean {
```
AlCalzone, line 150: An explicit any is better than no type. In this case you should limit the argument to non-nullable values, because hasOwnProperty will throw otherwise:
```suggestion
export function hasProperty(obj: {}, prop: string): boolean {
```

AlCalzone: FYI, `{}` is NOT the empty object type, this is a bit confusing.
`{}` is basically `any` minus `undefined | null` and without disabling type checking.

app.ts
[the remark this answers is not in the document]
AlCalzone: You're right!

lib/Constants.ts
robertsLando: yeah but the problem is that using the zwave-js command class names could create breaking changes in mqtt topics

lib/CustomPlugin.ts
robertsLando: In some cases like this I don't get why I should use interface instead of a type, mine it's a newbe question but they seems to be quite the same. The only difference is that interface can be extended but the type can use `&` operator

lib/CustomPlugin.ts
robertsLando: Oh now I got it

lib/Constants.ts
robertsLando: Type added

lib/MqttClient.ts
robertsLando: added undefined to types

lib/SocketManager.ts
robertsLando: That's black magic 🤣

lib/jsonStore.ts
robertsLando:
```
Type '{}' is missing the following properties from type 'Record<StoreKeys, any>': settings, scenes, nodes, users
```

Whithout that I got this error

lib/utils.ts
robertsLando:
```
Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.eslint@typescript-eslint/ban-types
```

When using `{}` as a type I get this

AlCalzone: > `{}` actually means "any non-nullish value"

Hm, you might actually want to limit it to objects, so `Record<string, unknown>` would be better.

robertsLando: Also using `Record<string, unknown>` it's not working 🤷🏼‍♂️

robertsLando: Disabled the rule for that line for now

lib/Constants.ts
AlCalzone: Yeah, but unnecessary to make an interface definition for something that's basically built in. If you want to alias the type, I'd do
```
type IDeviceClass = Record<number, IDeviceClassProps>
```

robertsLando: But in that way how could I alias the key?

PR #1321, opened 2021-06-09
fix: plugins router must be recreated on restart

kvaster:
expressjs is using static config, but we have option to restart plugins without restarting the whole service (when restarting zwave after applying settings). Plugin will fail (silently) to register expressjs handler after such restart.

Problem may be tested with the following plugin:

```
class TestPlugin {
  constructor(ctx) {
    this.zwave = ctx.zwave
    this.mqtt = ctx.mqtt
    this.logger = ctx.logger
    this.app = ctx.app

    var text = new Date().toString()

    this.logger.info(`Starting at: ${text}`)

    this.app.get('/test', function(req, res) {
      res.send(`Started at: ${text}`)
    })
  }

  async destroy() {
    this.logger.info('Stopped')
  }
}

module.exports = function(ctx) {
  return new TestPlugin(ctx)
}
```

* Start zwavejs2mqgg, curl to `/test`.
* Go to settings page and click 'Save'
* curl to '/test' and you will see old message insteadof new

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -107,9 +107,11 @@ socketManager.authMiddleware = function (
 
 let gw: Gateway // the gateway instance
 const plugins = [] as CustomPlugin[]
-const pluginsRouter = express.Router()
+let pluginsRouter
 
-app.use(pluginsRouter)
+app.use(function (req, res, next) {
+	pluginsRouter(req, res, next)
+})
 
 // flag used to prevent multiple restarts while one is already in progress
 let restarting = false
@@ -277,6 +279,7 @@ async function startGateway(settings: Settings) {
 	await gw.start()
 
 	const pluginsConfig = settings.gateway?.plugins ?? null
+	pluginsRouter = express.Router()
 
 	// load custom plugins
 	if (pluginsConfig && Array.isArray(pluginsConfig)) {
--- a/app.ts
+++ b/app.ts
@@ -109,10 +109,6 @@ let gw: Gateway // the gateway instance
 const plugins = [] as CustomPlugin[]
 let pluginsRouter
 
-app.use(function (req, res, next) {
-	pluginsRouter(req, res, next)
-})
-
 // flag used to prevent multiple restarts while one is already in progress
 let restarting = false
 
@@ -422,6 +418,10 @@ app.use(
 		},
 	})
 )
+
+app.use(function (req, res, next) {
+	pluginsRouter(req, res, next)
+})
 
 // Node.js CSRF protection middleware.
 // Requires either a session middleware or cookie-parser to be initialized first.


robertsLando (commented):
Did you try to use latest version on master? I think that the change you made is useless, the pluginsRouter instance is never destroyed and is registered before any other route. What I just noticed is that I should move that `app.use` after the 

```
app.use(
	session({ ...
```

kvaster:
> Did you try to use latest version on master? I think that the change you made is useless, the pluginsRouter instance is never destroyed and is registered before any other route. What I just noticed is that I should move that `app.use` after the
> 
> ```
> app.use(
> 	session({ ...
> ```

Yes. I've tested all on latest master. And the problem is not in the session.
Without patch I'm receiving old data with curl after pressing 'save' on settings page.

With my change pluginsRouter is recreated each time before loading plugins. The problem is that expressjs app is not fully recreated each time and same pluginsRouter instance is passed to new plugins instances each time. And if we will try to register handler on already registered path (by previous instance of same plugin) then it will not work. My patch will not register pluginsRouter. Instead it will register functional which will use router variable:

```
app.use(function (req, res, next) {	pluginsRouter(req, res, next)})
```

This function will always use new verison of pluginsRouter.

robertsLando:
Oh ok now I got it! Makes sense. Could you just move that `app.use` after the app.use that inits the session?

robertsLando (approved):

robertsLando:
@kvaster Let me know if now it works as expexted :)

kvaster:
Unfortunatelly it doesn't work as expected (at least for me) right now. With simple curl I'm getting right now following:

```
<!DOCTYPE html>
<html>

<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width,initial-scale=1.0">
  <title>ZWave To MQTT</title>
  <base href="/" />
```
 
I need to add "`Accept: application/json, text/plain`" header for request to be working, but prometheus (for example) have no such option... I think plugins handlers should work without such header.

We need to register pluginRouter handler just before history handler:

```
app.use(function (req, res, next) {
  if (pluginsRouter !== undefined) {
    pluginsRouter(req, res, next)
  } else {
    next()
  }
})

app.use(
	history({
		index: '/',
	})
)
```

robertsLando:
I think the problem are CORS, maybe you should try to register the plugin before registering cors?

robertsLando:
connect-history you mean?

kvaster:
This one peace of code: https://github.com/zwave-js/zwavejs2mqtt/blob/a5b444a06afff5077c5434dbd15705d0b60fdb12/app.ts#L390-L394

pluginRouter handler registration (use) should be before theese lines in order to work without accept header.

robertsLando:
Or maybe we could move the history middleware right before error handlers, let me try this

PR #1323, opened 2021-06-10
fix: driver ready error and ts improvments

robertsLando:


lib/MqttClient.ts
```diff
@@ -49,7 +49,7 @@ declare interface MqttClient {
 	on(event: 'multicastRequest', listener: (payload: any) => void): this
 	on(
 		event: 'apiCall',
-		listener: (topic: string, apiNema: string, payload: any) => void
+		listener: (topic: string, apiName: string, payload: any) => void
 	): this
 	on(event: 'connect', listener: () => void): this
 	on(event: 'brokerStatus', listener: (online: boolean) => void): this
```
AlCalzone, line 55: The point of the TypedEventEmitter is that you don't need to define this for every class that extends EventEmitter. I showed you an example in the second link in that review comment.

PR #1328, opened 2021-06-11
fix: force allowed apis to be a valid method of ZwaveClient

robertsLando:


AlCalzone (approved):
LGMT, just keep in mind that this "force" only applies to the types, JS and the frontend can still do whatever it wants unless it respects your types.

robertsLando:
@AlCalzone Not working yet, this is the implementation of your suggestion but seems that isn't validating the props correctly

AlCalzone:
Let me check

AlCalzone:
Can you try my commit? Seems that the `any` in `extends (EventEmitter as any)` wasn't good for this kind of stuff.

robertsLando:
```
TSError: ⨯ Unable to compile TypeScript:
Error: lib/Gateway.ts(1950,31): error TS2344: Type 'string | boolean | (<TEvent extends keyof ZwaveClientEventCallbacks>(event: TEvent, callback: ZwaveClientEventCallbacks[TEvent]) => ZwaveClient) | ... 61 more ... | (<TEvent extends keyof ZwaveClientEventCallbacks>(event: TEvent, ...args: Parameters<...>) => boolean)' does not satisfy the constraint '(...args: any) => any'.
  Type 'string' is not assignable to type '(...args: any) => any'.
```

robertsLando:
Seems that you impoorted `applyMixin` from `lib/utils` but there is no implementation of such method in utils

robertsLando:
Ok I have added it but seems that `keyof` opperator doens't 'fetch' private methods from class, any suggestion?

AlCalzone:
> Seems that you impoorted applyMixin from lib/utils but there is no implementation of such method in utils

Weird... It was there when I tested. Here you go: https://github.com/zwave-js/node-zwave-js/blob/4d9d317edf5192bec654d229fb8bc86c1966df29/packages/shared/src/inheritance.ts#L12-L41

> `keyof` opperator doens't 'fetch' private methods from class, any suggestion?

Its not supposed to.

robertsLando:
@AlCalzone I have found a 'compressed' solution for the `applyMixin`:

```ts
export function applyMixin(derivedCtor: any, constructor: any) {
	Object.getOwnPropertyNames(constructor.prototype).forEach((name) => {
		Object.defineProperty(
			derivedCtor.prototype,
			name,
			Object.getOwnPropertyDescriptor(constructor.prototype, name) ||
				Object.create(null)
		)
	})
}
```

AlCalzone:
Two problems with that.
1.) `any` should really be replaced with a constructor type
2.) This only works for "shallow" mixins. when a mixin inherits from another class, this will break.

robertsLando:
Ok I have added yours

PR #1339, opened 2021-06-14
fix: place plugins route before history middleware

robertsLando:
cc @kvaster. I also tried to place history middleware right before error handlers but that's not working. I have placed the pluginsRouter before the history one as you suggested


kvaster:
Great! Thanks!

PR #1344, opened 2021-06-15
chore(docker): bump node to 16.3.0

### LINKED ISSUE #1342: [bug] Node.js 15 is unsupported, migrate to next version? (Docker)
### Version

**Checklist:**

- [X] I am **not** using HomeAssistant. **Or:** a developer has told me to come here.
- [X] I have checked the troubleshooting section and my problem is **not** described there.
- [X] I have read the changelog problem was **not** mentioned there.

Build/Run method

- [X] Docker
- [ ] PKG
- [ ] Snap package

zwavejs2mqtt version: 5.0.1
zwave-js version: 7.7.3

### Describe the bug

The official docker release uses Node.js 15.11.0. Node.js 15 became [unsupported](https://nodejs.org/en/about/releases/) as of June 1, 2021. Would it make sense to begin migrating to 16?

Another thing I've noticed is that `Dockerfile.contrib` use the image `node:erbium-buster`, AKA Node.js 12. Would it also make sense to match this version to the same one as the official image, e.g. `node:16.3.0-alpine` and `node:16.3.0-buster`?

robertsLando:
Fixes #1342

cc @kpine

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 # STEP: 1
-FROM node:15.11.0-alpine AS build-z2m
+FROM node:16.3.0-alpine AS build-z2m
 
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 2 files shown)


robertsLando:
Seems that with latest alpine image `python` apk has been removed in favor of `python3` and `python2` packages and that was causing the fail on the build

PR #1351, opened 2021-06-17
chore(docker): yarn fails to install plugins

billiaz:
Yarn add needs to be used for plugin installs, otherwise it will not
work.
 by this addition we fix this issue


robertsLando (approved):
Ops, missed that on the yarn refactor. LGTM and happy to see you back :)

Issue #294, opened 2021-01-20
[question] Are dimmers still scaled down from 0..255 to 0..99 ?


LordMike:
One of my pet peeves with z2m was that in some cases, it would scale the light values from a supplied `0..255` range to a `0..99` range. I was hoping this scaling wasn't present in zwavejs2mqtt, as it was a new project and it was possible to make breaking changes :)..

I'm seeing logs like this, so I guess it is. Is there a way to disable this scaling?

```
2021-01-20 22:16:08.328 INFO MQTT: Message received on zwavejs2mqtt/wallswitch_2/38/5/targetValue/set, '16'
2021-01-20 22:16:08.329 INFO ZWAVE: Writing 6 to 38-5-targetValue
```

LordMike:
I actually may have my own answer.. Gateway.js scales multilevels in those ranges, but unlikc z2m, I can disable that code path by not having hass discovery.. 

https://github.com/zwave-js/zwavejs2mqtt/blob/3c6b99ec307f09924a962c4bdfc29d0509a95462/lib/Gateway.js#L705-L715

So I've disabled hass discovery.. :)

jcam:
With the fix-dimmers branch, this scaling will be removed entirely :)

LordMike:
Fix dimmers branch?

Great if it disappears. You can indicate to HASS what the scale is, if needed, by the [`brightnes_scale`](https://www.home-assistant.io/integrations/light.mqtt/#brightness_scale) setting. Setting this to `99`, makes `99` be the maximum value.

jcam:
Yes, we did that in the fix_dimmers branch which is now merged to master.

jcam:
Unfortunately a side effect of switching from the template to the default schema (required to use brightness_scale), is that if brightness is not known, it will no longer send 255, which is the z-wave internal value that means turn back on to last brightness

LordMike:
You can probably achieve that by reporting the `payload_on` value as `255` .. which in turn would mess with the `on_command_type` setting ... Hmm.

PR #290, opened 2021-01-20
fix(hass): dimmers and rgb dimmers

robertsLando:
Fixes #280 
Fixes #285 
Fixes #288
Fixes #287 
Fixes #286

PR #1360, opened 2021-06-20
fix(hass): make dimmers turn on to previous brightness

brlodi:
This fixes a side-effect from the changes in zwave-js/zwavejs2mqtt/pull/290 that caused using the toggle to turn a dimmer on to make the light come on at full brightness instead of restoring the previous level (see discussion at end of #294). While there might be an argument for that as intentional behavior, this change also brings the behavior over MQTT in line with the behavior when using the direct ZwaveJS integration (where Zwave dimmers return to their previous brightness when the toggle is used).

The only caveat I can find in my testing is a very slight but noticeable delay on the brightness slider updating, because HA isn't sending a brightness so it doesn't preemptively move the slider.

### DIFF
--- a/hass/configurations.ts
+++ b/hass/configurations.ts
@@ -55,10 +55,9 @@ const configurations: Record<HassDeviceKey, HassDevice> = {
 			state_topic: true,
 			command_topic: true,
 			on_command_type: 'last',
-      payload_on: 255,
-      payload_off: 0,
-			state_value_template:
-				'{{ 0 if value_json.value == 0 else 255 }}',
+			payload_on: 255,
+			payload_off: 0,
+			state_value_template: '{{ 0 if value_json.value == 0 else 255 }}',
 			brightness_value_template: '{{ value_json.value }}',
 			brightness_scale: 99,
 			rgb_command_template:
@@ -73,15 +72,14 @@ const configurations: Record<HassDeviceKey, HassDevice> = {
 		discovery_payload: {
 			command_topic: true,
 			state_topic: true,
-			state_value_template:
-				'{{ 0 if value_json.value == 0 else 255 }}',
+			state_value_template: '{{ 0 if value_json.value == 0 else 255 }}',
 			brightness_command_topic: true,
 			brightness_scale: 99,
 			brightness_state_topic: true,
 			brightness_value_template: '{{ value_json.value }}',
 			on_command_type: 'last',
-      payload_on: 255,
-      payload_off: 0,
+			payload_on: 255,
+			payload_off: 0,
 		},
 	},
 	volume_dimmer: {
--- a/hass/configurations.ts
+++ b/hass/configurations.ts
@@ -54,9 +54,11 @@ const configurations: Record<HassDeviceKey, HassDevice> = {
 		discovery_payload: {
 			state_topic: true,
 			command_topic: true,
-			on_command_type: 'brightness',
+			on_command_type: 'last',
+      payload_on: 255,
+      payload_off: 0,
 			state_value_template:
-				'{{ "OFF" if value_json.value == 0 else "ON" }}',
+				'{{ 0 if value_json.value == 0 else 255 }}',
 			brightness_value_template: '{{ value_json.value }}',
 			brightness_scale: 99,
 			rgb_command_template:
@@ -72,12 +74,14 @@ const configurations: Record<HassDeviceKey, HassDevice> = {
 			command_topic: true,
 			state_topic: true,
 			state_value_template:
-				'{{ "OFF" if value_json.value == 0 else "ON" }}',
+				'{{ 0 if value_json.value == 0 else 255 }}',
 			brightness_command_topic: true,
 			brightness_scale: 99,
 			brightness_state_topic: true,
 			brightness_value_template: '{{ value_json.value }}',
-			on_command_type: 'brightness',
+			on_command_type: 'last',
+      payload_on: 255,
+      payload_off: 0,
 		},
 	},
 	volume_dimmer: {


robertsLando:
@brlodi Thanks for your PR, could youi fix lint issues? `npm run lint-fix`

brlodi:
Yep sorry. I blindly trusted my editor's Prettier integration but it enforced *my* preferences instead of the project's. Should be lint-error-free now.

Diaoul:
With this change I cannot change the brightness of my dimmers anymore and they stay stuck in some (previous? full?) brightness state.
When I change the brightness in HASS using the slider, it immediately goes back to whatever position it was. I notice 2 MQTT payloads being sent when I do this: one to the value I clicked on, one on the previous value.
When I change back to the previous `brightness` instead of `last` there is no problem anymore.
The docs says:
> Using last (the default) will send any style (brightness, color, etc) topics first and then a payload_on to the command_topic.

So I understand that 2 payloads are sent to the command topic (which explains why I see 2 values there) one is the value I clicked on and the other is the payload ON which is defined as 255 now.

Here are my MQTT logs of the command topic from the MQTT integration listener:
```
Message 2 received on zwave/some/some-light/38/1/targetValue/set at 6:30 PM:
255
QoS: 0 - Retain: false
Message 1 received on zwave/some/some-light/38/1/targetValue/set at 6:30 PM:
27
QoS: 0 - Retain: false
```

Here are zwavejs2mqtt logs:
```
2021-06-29 16:30:21.725 INFO MQTT: Message received on zwave/some/some-light/38/1/targetValue/set, <Buffer 32 37>
2021-06-29 16:30:21.726 INFO ZWAVE: Writing 27 to 3-38-1-targetValue
2021-06-29 16:30:21.735 INFO MQTT: Message received on zwave/some/some-light/38/1/targetValue/set, <Buffer 32 35 35>
2021-06-29 16:30:21.737 INFO ZWAVE: Writing 255 to 3-38-1-targetValue
2021-06-29 16:30:21.772 INFO ZWAVE: Node 3: value updated: 38-1-currentValue 82 => 27
2021-06-29 16:30:21.892 INFO ZWAVE: Node 3: value updated: 38-1-currentValue 27 => 82
```

I am running on HA 2021.6.6 with zwavejs2mqtt 5.1.0 docker image.

brlodi:
I did have a limited set of devices to test this on, and was relying in part on the Zwave specification, so it's not impossible that you might have one that behaves differently. If it's causing problems it's probably best @robertsLando reverts this change until we can (if possible) sort it out.

The bit of spec this change depends on is this:

> A supporting device MUST restore the most recent (non-zero) value in response to the “On” state control command.

To me, and apparently the designers of the devices I currently have (a mix of Zooz and Jasco dimmers) and have previously used, the "most recent non-zero value" of a dimmer that is already on means its current level. That means you can safely send an "On" command to a light without changing the current brightness.

With your device one of two things is happening instead:

1. Due to network delays and perhaps a race condition on the device itself, it's getting the command to change to level 27 and then while that's still happening (perhaps due to a transition time) getting the 255 and looking up its "current" level, which hasn't yet changed from 82, so it resets itself to 82.
2. The designers of the device have misunderstood what "most recent" means, and the device will always return to the previous level when receiving 255. That would mean if you continued sending 255 to the device in your example it would flip back and forth between 82 and 27 each time.

I don't think there's any way to work around either of those on the HA discovery side, except by using a device-specific discovery payload for known-misbehaving devices. I suppose there might also be some debouncing that could happen in the ZwaveJS driver if the problem is (1), but (2) will likely mean most users will have to manually configure their compliant dimmers for full functionality and we'll have to leave the default as the least-common-denominator.

Diaoul:
I do not remember having issues with the light turning full brightness instead of restoring previous state but it is very likely that I missed it. I think zwavejs2mqtt should rely on the underlying hardware, eventually implement a per-device workaround when the manufacturer does not implement the spec properly. Sending 2 payloads to a compliant device seems like a waste and leading to a whole world of trouble like you mentioned in 1.
My zwave is pretty limited, is there a way to trigger an ON command on the zwave protocol without specifying brightness so that it should restore to previous state? This is the default behavior when I press the switch: brightness is restore to previous on state.

Diaoul:
So just to summarize what I understood so far (please correct me!):
- this PR adds the effect that the brightness gets stuck due to 2 payloads being sent to the device and probably some concurrency issue involved here
- 2 payloads are sent when `on_command_type` is `first` or `last`, one with the `ON` payload and one with the `brightness`
- #290 seems to be the root cause, relying on `brightness_scale` makes the *turn on* command no different that the "set to max brightness" command, previous to that PR, 255 was a special value sent that would mean "turn on to previous value" to zwavejs2mqtt

I think there are 4 scenarios that need to be identified properly:
| Previous | Command | Expected behavior |
|-------------|----------------|--------------------------|
| Off          | On             | Light turns on to previous brightness |
| Off          | x %            | Light turns on to x % |
| x %         | y %            | Light turns to y % |
| Any         | Off             | Light turns off |

We could get away with 3 kinds of payloads: `ON`, `OFF` and `x %` and craft the proper command to send to the device.
Having `on_command_type` to `first` or `last` will necessarily involve some work with specific time windows to identify which one of the 3 it is or some internal intermediate state to issue a single command to the device.

| Payload 1 | Payload 2 | Maps to |
| ------------- | ------------- | ------------|
| ON | 24 | 24 % |
| ON | 99 | 99 % |
| ON |  -   | ON |
| OFF |  - | OFF |

(*) If `first` payload 1 will be sent first, in `last` mode, it will come last.

I think those modes were intended when setting the brightness does not turn on the device, just updates some internal brightness value (I use this for some of my ESP8266), hence the need to send the ON command after setting the brightness (this is why `last` is the default behavior I imagine).

We could use some time window where *if within the 50ms following the `ON` command we receive a brightness* we send the `x %` command. Or maintain the brightness state in memory and emulate that behavior.
This seem like a slippery slope, the best I think would be to have HASS send the 3 different payloads directly and not have to interpret the meaning of this.

I have no idea how to instruct HASS to do this using discovery :disappointed:

brlodi:
> I do not remember having issues with the light turning full brightness instead of restoring previous state but it is very likely that I missed it.

I think it's very likely the same concurrency problem that's making them misbehave now was covering up the effect before.

> My zwave is pretty limited, is there a way to trigger an ON command on the zwave protocol without specifying brightness so that it should restore to previous state? This is the default behavior when I press the switch: brightness is restore to previous on state.  
> 255 was a special value sent that would mean "turn on to previous value" to zwavejs2mqtt

Sending `255` *is* the special value to do that, and to clarify it isn't a matter of meaning something special to *Zwavejs2MQTT*, it means something special in the *Zwave* spec. Zwave dimmers can accept the values `0`, a number in `[1-99]` inclusive, or `255`, with `0` and `255` meaning 'off' and 'on' respectively.

Zwavejs2MQTT actually automatically translates `ON` and `true` coming in on MQTT into `255` to send to the actual device. I could have made the `on_command` one of those, but it would be equivalent.

> I think those modes were intended when setting the brightness does not turn on the device, just updates some internal brightness value (I use this for some of my ESP8266), hence the need to send the ON command after setting the brightness (this is why last is the default behavior I imagine).

I'm sure that's where it comes from, although it's pretty common behavior across the board. For example, HomeKit notoriously always sends both ON and a brightness level if it thinks the target is currently off, and not necessarily in a fixed order! I recall the SmartThings Zwave integration doing something similar.

___

It might be wasteful to send two commands, but it *should* be fine according to the Zwave spec. As long as the `255` arrives after the brightness it should do nothing. Unfortunately that only holds true if devices follow the spec and don't have bugs. This would have been a nice straightforward way to solve this with minimal overhead (i.e. no more than most Zwave implementations have).

I appreciate the effort you've put into charting out the payloads, but I don't think it is nor should be Zwavejs2MQTT's—nor ZwaveJS itself—responsibility. They should just faithfully pass whatever values come in and out with only the most minor formatting adjustments. (Personally I think turning ON into 255 might already be too much, but I have to admit it's handy)

What we're now up against is the limitations of _Home Assistant_'s MQTT integration. Originally Zwavejs2MQTT was performing scaling itself, which caused a lot of confusion, so the decision was made to switch from discovering as a "template schema" light with that custom scaling code and instead discovering as a "default schema" light which can take advantage of HA's built in scaling.

I believe that, without either modifying the Home Assistant integration itself or burdening Zwavejs2MQTT with a lot of complex logic to work around it, the only way to get this working on noncompliant/bugged devices would be to go back to discovering as a template schema.

I'm willing to take it on if @robertsLando agrees, but it is a mixed bag for sure.

Regardless, this PR should be reverted. I'm sorry for the difficulty @Diaoul.

robertsLando:
@brlodi Can you please open a PR with the revert?

Diaoul:
If we can get Home Assistant to send us only 3 different kind of payloads as I mentioned, zwavejs2mqtt will just send that to the zwave device and this should just work, shouldn't it?
From what I understand this will avoid the buggy behavior of 2 payloads (ON + brightness), and their order while still procuring the expected behavior.

>  They should just faithfully pass whatever values come in and out with only the most minor formatting adjustments.

I agree that the work (if any) should be done in the MQTT integration to provide better control during the discovery process about how to interact with the device and from zwavejs2mqtt this should just be a matter of configuring the discovery properly.

robertsLando:
If you guys could tell me the changes needed or even better send a PR for this it would be more then welcome as I don't use HASS so it's hard for me to test this

brlodi:
Sorry was traveling yesterday. Revert is in PR #1399

brlodi:
I'll see what can be done to resolve it another way and submit that separately. Let's get @Diaoul and everyone's dimmers working again in the sort term.

Diaoul:
I have put _brightness_ instead of _last_ and things are working at least for now (expect ON does not restore the previous state). There is no hurry.
Thank you both for your help on this.

virtualJonesie:
Is there any update on this? I recently moved from ZWave2MQTT because it was depreciated in lieu of ZWaveJS2MQTT and this is a disappointing discovery.

PR #1369, opened 2021-06-23
feat: preferred sensor scales

### LINKED ISSUE #1362: [feat] Set preferred scales
https://github.com/zwave-js/node-zwave-js/pull/2891/files

robertsLando:
Fixes #1362 

Peek 24-06-2021 09-41


app.ts
```diff
@@ -801,11 +807,39 @@ app.get(
...
+		for (const [, sensor] of sensorTypes) {
+			for (const [, scale] of sensor.scales) {
+				scales.push({
+					type: sensor.key,
+					label: sensor.label,
+					unit: scale.unit,
+					description: scale.description,
+				})
+			}
+		}
```
AlCalzone, line 835: I would unify how those are handled, so you can also use `configManager.namedScales` instead of hardcoding the named ones. Proposal:
- key = scale name or sensor type (key)
- label = scale label
- unit = scale unit
- description: seems unnecessary, unless you display it.

lib/ZwaveClient.ts
```diff
@@ -1022,41 +1044,44 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+					forceConsole: true,
+					nodeFilter:
+						this.cfg.nodeFilter && this.cfg.nodeFilter.length > 0
+							? this.cfg.nodeFilter.map((n) => parseInt(n))
+							: undefined,
+				},
+			}
+
+			if (this.cfg.scales) {
+				const scales: Record<string | number, string | number> = {}
```
AlCalzone, line 1067: The reason this is modeled as an object is because you cannot define multiple preferred scales for one sensor type (or per named scale group like "temperature"). From a UX perspective, the multiselect dropdown seems wrong. Instead it should be: select sensor type or named scale, then select one preferred scale for that.

AlCalzone (changes requested):
Also it should be made clear that this is a **preferred** sensor scale - zwave-js uses them when it queries the device without knowing which scale to use. This does not influence what the device sends.

app.ts
```diff
@@ -32,6 +37,7 @@ import {
 	CustomPlugin,
 	PluginConstructor,
 } from './lib/CustomPlugin'
+import { Scale } from 'zwave-js'
```
AlCalzone, line 40: Unused according to Github

app.ts
```diff
@@ -801,11 +808,42 @@ app.get(
...
+
+		for (const [key, group] of sensorGroups) {
+			for (const [, scale] of group) {
+				scales.push({
+					key: key,
+					sensor: group.name,
+					unit: scale.unit,
+					label: scale.label,
+					description: scale.description,
+					group: true,
```
AlCalzone, line 824: What do you need this flag for?

app.ts
```diff
@@ -801,11 +808,42 @@ app.get(
 	apisLimiter,
 	isAuthenticated,
 	async function (req, res) {
+		const sensorTypes = configManager.sensorTypes
+		const sensorGroups = configManager.namedScales
```
AlCalzone, line 812: I think the name is confusing - these are not sensor groups. These are named groups of scales.

app.ts
robertsLando: Just to distinguish groups from the others

app.ts
```diff
@@ -809,15 +808,15 @@ app.get(
...
 
 		const scales: SensorTypeScale[] = []
 
-		for (const [key, group] of sensorGroups) {
+		for (const [key, group] of sensorScalesGroups) {
 			for (const [, scale] of group) {
 				scales.push({
 					key: key,
-					sensor: group.name,
+					sensor: group.name || key,
```
AlCalzone, line 819: `group.name` is guaranteed to exist for `namedScales`, you could do this:
```suggestion
					sensor: group.name!,
```

app.ts
```diff
@@ -809,15 +808,15 @@ app.get(
...
 					unit: scale.unit,
 					label: scale.label,
 					description: scale.description,
```
AlCalzone, line 822: And I still don't understand for what you need `group: true`. I don't see where it is used.

robertsLando: I was using it previously but not anymore, removing it now

app.ts
robertsLando: It doesn't let me add the ! at the end as ts says it's useless. just removed the `||` condition

AlCalzone (approved):

PR #1383, opened 2021-06-25
fix(hass): add default to central_scene

### LINKED ISSUE #1020: [bug] Add default on all HA templates for attributes
### Version

**Checklist:**

- [ ] I am **not** using HomeAssistant. **Or:** a developer has told me to come here.
- [X ] I have checked the troubleshooting section and my problem is **not** described there.
- [ ] I have read the changelog problem was **not** mentioned there.

Build/Run method

- [ X] Docker
- [ ] PKG
- [ ] Snap package

zwavejs2mqtt version: 3.2.0
zwave-js version: 7.1.1

### Describe the bug

Similar to the issue for zigbee2mqtt described [here](https://github.com/Koenkk/zigbee2mqtt/pull/6988) 

### To Reproduce

1. start zwave2jsmqtt
2. home_assistant.log shows lots of 
`WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'value' when rendering '{{ {}[value_json.value] | default(value_json.value) }}'`
and
`WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'value' when rendering '{{ value_json.value}}'`

### Expected behavior

There should not be warnings.

robertsLando:
Fixes #1020

### DIFF
--- a/hass/configurations.ts
+++ b/hass/configurations.ts
@@ -44,7 +44,7 @@ const configurations: Record<HassDeviceKey, HassDevice> = {
 		object_id: 'scene_state',
 		discovery_payload: {
 			state_topic: true,
-			value_template: '{{ value_json.value}}',
+			value_template: "{{ value_json.value | default('', true) }}",
 		},
 	},
 	// Light https://www.home-assistant.io/components/light.mqtt


emichael:
I could be wrong, but I don't think `, true` is what you want. The OP in zigbee2mqtt got it backwards. If the optional second arg to `default` is `true`, then `default` will also replace any value that, when coerced to a boolean, translates to `false` with the default. If you leave off the second argument, it will only replace values when they're undefined.

robertsLando:
Thanks for pointing this out, will fix asap

robertsLando:
Fixed.

PR #1391, opened 2021-06-28
fix: clear node groups

robertsLando:


AlCalzone (commented):
LGTM, but maybe you'll want to clear it only when getting the new groups succeeds?

robertsLando:
@AlCalzone IMO it would be ok to just clear them before starting, if user is requesting an update it means that the actual cache could be wrong so better to have no groups at all then a wrong cache

PR #1394, opened 2021-06-30
fix: check updates after driver is ready

robertsLando:


lib/ZwaveClient.ts
```diff
@@ -2025,6 +2023,9 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		this._updateControllerStatus('Driver ready')
 
 		try {
+			// this must be done only after driver is ready
+			await this._scheduledConfigCheck()
```
AlCalzone, line 2027: I wouldn't await here, just fire and forget.
```suggestion
			this._scheduledConfigCheck().catch(() => { /* ignore */ });
```

Or at least do it after attaching the event handlers.

AlCalzone (changes requested):

AlCalzone (approved):

PR #1523, opened 2021-08-02
fix: catch "controller not ready" error in statistics event handler

AlCalzone:
For some reason, the controller no longer exists after its event handler is invoked.
I don't know why this happens. It should not, but it does and it bombs sentry with errors.


robertsLando (commented):
🤷🏼‍♂️ LGTM

Issue #1463, opened 2021-07-20
Dockerfile.contrib gets broken by upgrading to v8 of `zwave-js`


AlCalzone:
See https://zwave-js.github.io/node-zwave-js/#/development/installing-from-github for reference.

This should not be done
https://github.com/zwave-js/zwavejs2mqtt/blob/698958e55413af49f00ad93ec41b9612f7d20743/docker/Dockerfile.contrib#L27

This is unnecessary
https://github.com/zwave-js/zwavejs2mqtt/blob/698958e55413af49f00ad93ec41b9612f7d20743/docker/Dockerfile.contrib#L28

Yarn v3 no longer has a --network-timeout param:
https://github.com/zwave-js/zwavejs2mqtt/blob/698958e55413af49f00ad93ec41b9612f7d20743/docker/Dockerfile.contrib#L29
Use https://yarnpkg.com/configuration/yarnrc#httpTimeout instead or set it as an env variable:
```
YARN_HTTP_TIMEOUT=... yarn install
```

The `build:full` command is deprecated, use `build` instead
https://github.com/zwave-js/zwavejs2mqtt/blob/698958e55413af49f00ad93ec41b9612f7d20743/docker/Dockerfile.contrib#L30

The `--frozen-lockfile` parameter is now called `--immutable`
https://github.com/zwave-js/zwavejs2mqtt/blob/698958e55413af49f00ad93ec41b9612f7d20743/docker/Dockerfile.contrib#L31

This doesn't work:
https://github.com/zwave-js/zwavejs2mqtt/blob/698958e55413af49f00ad93ec41b9612f7d20743/docker/Dockerfile.contrib#L31
Creating a production install should be done with this `yarn` plugin instead:
https://gitlab.com/Larry1123/yarn-contrib/-/tree/master/packages/plugin-production-install

Not sure if versioning that way works:
https://github.com/zwave-js/zwavejs2mqtt/blob/698958e55413af49f00ad93ec41b9612f7d20743/docker/Dockerfile.contrib#L34-L36
but in any case, `yarn workspaces foreach --include ... ` should be used instead to run commands in the workspaces. Linking that way won't work either, better use the aforementioned plugin and overwrite the install path with `"resolutions"` as described here:
https://zwave-js.github.io/node-zwave-js/#/development/installing-from-github

/cc @larstobi

robertsLando:
@larstobi Could you apply the required changes to that file? I don't use it

[...]

PR #1585, opened 2021-08-23
feat: support for S2 inclusion

robertsLando:
- Added new ZwaveClient methods `grantSecurityClasses` `validateDSK` `abortInclusion`
- Changed signature of `startInclusion` and `replaceFailedNode` methods
- New socket events
- Added `securityKeys` object setting to Zwave settings (replaces the old `networkKey` setting, old settings are automatically updated to the new format so it's back compatible)
- Added node `security` property that shows the security class used by the node (S2, S0, None, ...)
- Completely refactored Add/Remove dialog to support new inclusion steps and replace failed nodes:

https://user-images.githubusercontent.com/11502495/130618833-ec2e13b1-38c0-4d1f-a959-ab70c1e0d990.mp4

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -472,6 +472,7 @@ function setupSocket(server: HttpServer) {
 		// eslint-disable-next-line @typescript-eslint/no-misused-promises
 		async (socket, data) => {
 			if (gw.zwave) {
+				if (!data.args) data.args = []
 				const result: CallAPIResult<any> & {
 					api?: string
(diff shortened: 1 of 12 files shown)


blhoward2:
I can't test this because of #1463. I spent two hours trying work through the changes but its a bridge too far. I can't get past the last few comments about `yarn versions` and `RUN yarn install --production --frozen-lockfile `

PR #1672, opened 2021-09-10
docs: add `refreshCCValues` to the list of APIs

karwosts:
I was looking for ways to just refresh specific values through the API, and realized that refreshCCValues already exists in the API, but it is missing from the list of supported API calls. I tested the API call from NodeRed and it worked as expected.

I see that getNodeNeighbors, driverFunction, checkForConfigUpdates, and installConfigUpdates are missing from the list as well, though I don't know as much about those and wasn't able to test them. Not sure if those should be added as well.


robertsLando:
I may have missed it :) Thanks

robertsLando (approved):

PR #1675, opened 2021-09-11
feat(ui): show power source + battery level in nodes table

### LINKED ISSUE #1638: [feat] Show battery status on nodes table
When doing device maintenance of battery-powered device you have to look into the details of every device to see if the battery is still in a good state.

Showing the battery state on the nodes table gives a quick overview and lets you sort by percentage to quickly see which have to be replaced.

Use scripts to extract the information from the device data (not user-friendly).

**Additional context**
-

ahochsteger:
This PR adds a power column to the nodes table showing the **power source** (mains-powered, battery) using an icon and the **battery level** for battery-powered devices.

Fixes #1638

**Screenshot:**
Screenshot_20210916_153645

**Changes under the hood:**

Now all over 5900 icons from **[MaterialDesignIcons.com](https://materialdesignicons.com/)** are available using <https://github.com/JamesCoyle/vue-icon> with low size (only used icons are packaged).

It is now possible to **Map node properties** (e.g. `batteryLevel`) to a `ValueID` (e.g. commandClass 'Battery', property 'level') in `ZwaveClient.ts`:

```typescript
// Define mapping of node properties to ValueIDs
const mappedNodeProps: Record<string, ValueID> = {
	batteryLevel: {
		commandClass: CommandClasses.Battery,
		property: 'level',
	},
}
```

The functionality is encapsulated in the function `_mapNodeValuesToProps()` that is triggered on node ready and value changed events.

The **node property definitions** for the table can now be **extended** using the following functions to be more dynamic:

* `customValue`: A function to extract the column value from a given node (as it already was). Note though, that grouping may have to be disabled for such columns, since it is not working in `VDataTable` (until <https://github.com/vuetifyjs/vuetify/issues/7066> is solved).
* `customFormat`: A function to format the value for displaying (e.g. add '%' to the battery level on grouping)
* `customInfo`: A function to provide more detailed information to be displayed as a tooltip
* `customSort`: A function to define a custom sorting for the column (used for `batteryLevel` to sort mains-powered devices like 100% battery).

Finally the **Zwave graph** has been **refactored** to directly use `node.batteryLevel` istead of iterating over the ZWave values.

**Known issues:**

- Grouping values in combination with `customValue` does not work. The implementation of `genGroupedRows()` may have to be adjusted to respect `customValue` to extract the battery value instead of directly accessing it using the property name: <https://github.com/vuetifyjs/vuetify/blob/4cc80ca7bea963be875b3ab0f9ba03a8b21c4d92/packages/vuetify/src/components/VDataTable/VDataTable.ts#L316>. This should be possible if <https://github.com/vuetifyjs/vuetify/issues/7066> is resolved.

**Related information:**

* <https://github.com/vuetifyjs/vuetify/issues/7066>
* <https://stackoverflow.com/questions/52678905/how-to-use-vutifys-custom-sort>


ahochsteger:
@robertsLando, I'd need some advice on the approach I've taken to deal with the battery level as well as the use of additional icons.

**Approach to support the battery level**

To display the battery level on the table I see two following **approaches**:
1. **Extend the table** to be able to **dynamically find** the required nested information (e.g. battery level)
2. **Extend the node object** to provide the **battery level** as top level **property**

I've implemented the 1st approach since I wanted the table to display as much information as possible without having to change the general node object structure. It works but for displaying and filtering but not yet for sorting and grouping.
I may solve these issues but I'm unsure if this is really the way to go, since extracting the battery level requires to iterate a lot to find the respective command class that holds the value which may have a **negative impact** to the overall **performance** of the table.

Before I continue to solve the remaining issues I'd like to get your opinion if it's not better to provide the battery directly as property of the node object.

**Using icons from materialdesignicons.com**

Another feedback I'd like to have is for the use of additional icons, since I find the available ones a bit limited.
I used Material Design Icons which recommends this way to include additional icons:
<https://dev.materialdesignicons.com/getting-started/vuejs>
This way I was able to show the battery level directly in the icon.
If I remember correctly we already had a discussion on increased file size because of additional icons I observed the size of the dist directory and it seems to have no negative impact. In fact it seems to take less space than using the built-in icons but I'm not sure, if that is really the case - an independent look would be helpful.
It seems that only the used icons are included into the final distribution due to tree shaking.

robertsLando:
Hi @ahochsteger , thanks for your PR! I love to keep the code as simple as possible, if it was me I would have choose to set the battery level as a node property instead. It is also used on mesh graph so it makes sense to have it there. Just remember to add the code to the value update in order to keep it updated

package.json
```diff
@@ -122,6 +122,8 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.12.1",
+    "@jamescoyle/vue-icon": "^0.1.2",
+    "@mdi/js": "^6.1.95",
```
robertsLando, line 126: Do we really need this packages?

src/modules/ManagedItems.js
```diff
@@ -108,7 +108,10 @@ export class ManagedItems {
 	get propValues() {
 		const values = {}
 		Object.keys(this.propDefs).forEach((propName) => {
-			values[propName] = this.getPropValues(propName)
+			let valueFn = this.propDefs[propName].valueFn
+				? this.propDefs[propName].valueFn
+				: (node) => node[propName]
+			values[propName] = this.getPropValues(propName, valueFn)
```
robertsLando, line 114: Cannot you just write:

```suggestion
			values[propName] = this.getPropValues(propName, this.propDefs[propName].valueFn)
```

robertsLando (changes requested):
As I said I would have move the battery level to node props, BTW I like your implementation here and I think that it could be useful also in future if we need to parse some other column fetched from node values. So keep this.

My only question is if we really need the mdi package as it is huge, also we should check if the icons are working also offline and how much the package impact to the final boundle size

package.json
ahochsteger: As I understand it, the webpack tree shaking should only include the few bytes for each used icon - that's why I'm using the recommended way for Vue and don't use the web font (which is huge).
It may take even less if existing icons would be used that way as well ...

ahochsteger: > Hi @ahochsteger , thanks for your PR! I love to keep the code as simple as possible, if it was me I would have choose to set the battery level as a node property instead. It is also used on mesh graph so it makes sense to have it there. Just remember to add the code to the value update in order to keep it updated

@robertsLando, thanks for your feedback, that confirms my thoughts while implementing this.
Is there already an implementation of a similar property that syncs the value property I could use as reference? Maybe you have a pointer for me ...

robertsLando: > Is there already an implementation of a similar property that syncs the value property I could use as reference? Maybe you have a pointer for me ...

Unfortunally nope, but I made my comment before to see your actual implementation and I think it looks good anyway and could be helpful in future for other addition on the table

robertsLando: About the icons if it doesn't add the entire icons font I'm ok for them :)

ahochsteger: > Unfortunally nope, but I made my comment before to see your actual implementation and I think it looks good anyway and could be helpful in future for other addition on the table

Ok, so I'll fix the sorting and grouping issue for the table and keep accessing the value using the `valueFn` for now. We have to keep an eye on the performance though, since every use of the function causes an iteration over the values to find the relevant command class. If it will cause problems in the future (currently I didn't see it with showing all 66 devices), we may have to go the node property way ...

ahochsteger:
Next steps:
* Fix sorting and grouping by using custom sorting functions (see <https://stackoverflow.com/questions/52678905/how-to-use-vutifys-custom-sort>)

ahochsteger:
@zwave-js-bot fix lint

package.json
robertsLando: > We have to keep an eye on the performance though, since every use of the function causes an iteration over the values to find the relevant command class. If it will cause problems in the future (currently I didn't see it with showing all 66 devices), we may have to go the node property way

Sure! 👍🏼

ahochsteger:
@zwave-js-bot fix lint

ahochsteger:
# Battery level PR

@robertsLando the PR is ready to be reviewed from my point of view.
I'll give you a quick summary of what changes after your last review ...

The approach using a function to extract the battery level dynamically worked but had the limitation that VDataTable does not provide a way to override the standard behavior for grouping, thus grouping was broken for this column.
We could easily disable grouping for such columns but I wanted to have it working for the battery level.

Therefore I tried the other approach to map the battery level directly to a node property.
It turned out to be much easier than I thought and solved it by:

**Map node properties** (e.g. `batteryLevel`) to a `ValueID` (e.g. commandClass 128, property 'level') in `ZwaveClient.ts`:

```typescript
// Define mapping of node properties to ValueIDs
const mappedNodeProps: Record<string, ValueID> = {
	batteryLevel: {
		commandClass: CommandClasses.Battery,
		property: 'level',
	},
}
```

The functionality is encapsulated in the function `_mapNodeValuesToProps()` that is triggered on node ready and value changed events.

The **node property definitions** for the table can now be **extended** using the following functions to be more dynamic:

* `valueFn`: A function to extract the column value from a given node (as it already was). Note though, that grouping may have to be disabled for such columns, since it is not working in `VDataTable` (until <https://github.com/vuetifyjs/vuetify/issues/7066> is solved).
* `formatFn`: A function to format the value for displaying (e.g. add '%' to the battery level on grouping)
* `sortFn`: A function to define a custom sorting for the column (used for `batteryLevel` to sort mains-powered devices like 100% battery).

Finally the **Zwave graph** has been **refactored** to directly use `node.batteryLevel` istead of iterating over the ZWave values.

It would be great if you could give me your final feedback to finalize the PR :-).

lib/ZwaveClient.ts
```diff
@@ -3293,6 +3309,46 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+							": mapping property '" +
+							k +
+							"' => commandClass=" +
+							valueId.commandClass +
+							' (' +
+							CommandClasses[valueId.commandClass] +
+							"), property='" +
+							valueId.property +
+							"'"
+					)
```
robertsLando, line 3341: Move this to `debug` level and use template string instead of concatenation

robertsLando (changes requested):
@ahochsteger Looks good! I like this approach :) I think this is the best and more efficent way to implement such feature, good job! Just fix that log and move it to debug level otherwise we could have too much useless logs :)

lib/ZwaveClient.ts
ahochsteger: @ahochsteger, thanks for your feedback and kind words :-).
A quick question about naming:
I'm thinking about changing the naming of the functions that can be defined in the nodes table property definition and would like to get your feedback. Esp. I'd like to get rid of the `Fn` suffix to be more flexible in the future, since I can test for a function anyway. E.g.:
* `valueFn` => `value`
* `sortFn` => `customSort` (named after the `customSort` function, or we may just name it `sort`)
* `infoFn` => `info`
* `formatFn` => `format`

What do you think?

robertsLando: I think that yes in some cases instead of just checking if them exists adding a `typeof 'function` check would be more clear. Go for it :)

ahochsteger: @robertsLando the PR is finished and ready to be merged. :-)

I decided to rename them to `customValue`, `customSort`, `customInfo` and `customFormat` to better communicate what it is and to prevent naming clashes, when they are put into the `headers` object of `VDataTable`.

robertsLando (approved):
LGTM. Feel free to squash and merge it if you are done here and thanks for the PR 🙏🏼

PR #1726, opened 2021-09-23
fix: workaround for pkg bug

ahochsteger:
@robertsLando is this what you intended as workaround for the pkg fix?
I guess this is the bug you've been referring to in Slack: https://github.com/vercel/pkg/issues/1134 
Unfortunately I could not reproduce the problem locally but hopefully it does not process the file node_modules/@jamescoyle/vue-icon/lib/svg-icon.vue if I rename it to *.bak and move it out of node_modules.


robertsLando (approved):
Looks good! Let's see if this is enough to fix the problem. Thanks 🙏🏻

robertsLando:
BTW the problem is exactly the one you have linked

PR #1777, opened 2021-10-03
feat: support nodes with multiple batteries

ahochsteger:
* add powerSource node property to improve power source detection
* add batteryLevels node property with multiple values
* rename batteryLevel to batteryLevelMin node property for minimum battery level

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -140,17 +140,61 @@ const allowedApis = validateMethods([
 	'abortInclusion',
 ] as const)
 
+export type CmdClsExistsMapping = {
+	commandClass: CommandClasses
+	existsValue: any
+	absentValue: any
+}
+
+export enum NodePropMappingAggFn {
+	ALL = 'all',
+	MIN = 'min',
+	MAX = 'max',
+	FIRST = 'first',
+	LAST = 'last',
+}
+
+export type NodePropMapping = {
+	hasCommandClass?: CmdClsExistsMapping
+	valueId?: ValueID
+	aggregationType?: NodePropMappingAggFn
+}
+
 // Define mapping of node properties to ValueIDs
-const mappedNodeProps: Record<string, ValueID> = {
-	batteryLevel: {
-		commandClass: CommandClasses.Battery,
-		property: 'level',
+const mappedNodeProps: Record<string, NodePropMapping> = {
+	batteryLevelMin: {
+		valueId: {
+			commandClass: CommandClasses.Battery,
+			property: 'level',
+		},
+		aggregationType: NodePropMappingAggFn.MIN,
+	},
+	batteryLevels: {
+		valueId: {
+			commandClass: CommandClasses.Battery,
+			property: 'level',
+		},
+		aggregationType: NodePropMappingAggFn.ALL,
+	},
+	powerSource: {
+		hasCommandClass: {
+			commandClass: CommandClasses.Battery,
+			existsValue: 'battery',
+			absentValue: 'mains',
+		},
 	},
 }
 // Extract mapped command classes for better mapping performance
-const mappedCommandClasses: Set<number> = new Set()
+const mappedValueIdCommandClasses: Set<number> = new Set()
+const mappedExistingCommandClasses: Set<number> = new Set()
 for (const k in mappedNodeProps) {
-	mappedCommandClasses.add(mappedNodeProps[k].commandClass)
+	if (mappedNodeProps[k].valueId) {
+		mappedValueIdCommandClasses.add(mappedNodeProps[k].valueId.commandClass)
+	} else if (mappedNodeProps[k].hasCommandClass) {
+		mappedExistingCommandClasses.add(
+			mappedNodeProps[k].hasCommandClass.commandClass
+		)
+	}
 }
 
 export type SensorTypeScale = {
@@ -318,7 +362,9 @@ export type Z2MNode = {
 	status?: keyof typeof NodeStatus
 	inited: boolean
 	healProgress?: string | undefined
-	batteryLevel?: number
+	batteryLevelMin?: number
+	batteryLevels?: { [key: string]: number }
+	powerSource?: string
 }
 
 export type ZwaveConfig = {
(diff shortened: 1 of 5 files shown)


ahochsteger:
@kvaster would you be able to test the implementation of this PR?
It would be good to have some feedback before merging it.

ahochsteger:
@robertsLando there were some changes under the hood to allow handling of multiple values for a certain command class.
May you have a look at this generic mapping definition in `ZwaveClient.ts`:

```typescript
// Define mapping of node properties to ValueIDs
const mappedNodeProps: Record<string, NodePropMapping> = {
	batteryLevelMin: { // provides a single value with the minimum of all
		valueId: {
			commandClass: CommandClasses.Battery,
			property: 'level',
		},
		aggregationType: NodePropMappingAggFn.MIN, // Defines how to aggregate multiple values (MIN, MAX, FIRST, LAST, ALL)
	},
	batteryLevels: { // provides a list of battery levels for every endpoint
		valueId: {
			commandClass: CommandClasses.Battery,
			property: 'level',
		},
		aggregationType: NodePropMappingAggFn.ALL, // Provide all values using the endpoint as key
	},
	powerSource: { // provides a more reliable powerSource property without interpreting battery levels in a complex way
		hasCommandClass: { // maps the property value based on the existence of a command class
			commandClass: CommandClasses.Battery,
			existsValue: 'battery',
			absentValue: 'mains',
		},
	},
}
```

The PR is still in draft mode, because there's some more refactoring I'd like to do to simplify some of the code.

src/components/nodes-table/index.vue
```diff
@@ -125,18 +125,7 @@
...
-						type="mdi"
-						:path="getPowerInfo(item).icon"
-						:style="getPowerInfo(item).style"
-					></svg-icon>
-					<span style="padding-top: 4px">{{
-						getPowerInfo(item).label
-					}}</span>
-				</v-layout>
-			</div>
+			<table-value :value="getPowerInfo(item)" />
```
ahochsteger, line 128: Maybe `custom-display` would be a better name instead of `table-value`.

src/components/nodes-table/nodes-table.js
```diff
@@ -47,10 +48,17 @@ export default {
...
+						if (node.powerSource === 'mains') return 'mains-powered'
+						let levelInfo = ''
+						for (const ep in node.batteryLevels) {
+							if (levelInfo !== '') {
+								levelInfo += '/'
+							}
+							levelInfo += `${node.batteryLevels[ep]}%`
+						}
+						return `Battery level(s): ${levelInfo}`
+					},
```
ahochsteger, line 61: Maybe providing a `customDisplay` function instead of separate `customFormat` and `customInfo` would be better.
The function would then expect an object like the one provided by `getPowerInfo()`.
This way any table column could be customized in a better way (e.g. to show icons instead of 'Yes'/'No').

src/components/nodes-table/index.vue
robertsLando: Agree 👍🏼

src/components/nodes-table/nodes-table.js
robertsLando: Agree 👍🏼

lib/ZwaveClient.ts
```diff
@@ -140,17 +140,61 @@ const allowedApis = validateMethods([
...
+			property: 'level',
+		},
+		aggregationType: NodePropMappingAggFn.ALL,
+	},
+	powerSource: {
+		hasCommandClass: {
+			commandClass: CommandClasses.Battery,
+			existsValue: 'battery',
+			absentValue: 'mains',
+		},
```
robertsLando, line 184: I think this could be made a bit cleaner, this could become a record of CC -> propMap like:

```js
const mappedNodeProps = {
[CommandClasses.Battery]: [
{
	nodeProp: 'batteryLevelMin',
	fn: (node, value) => {
		// the function to extract the value
	}
	
}
]
}
```

This would allow to remove the mappedValueIdCommandClasses and mappedExistingCommandClasses, you could just add a check if the valueId CC exists in the mappedNodeProps keys and if so for every definition in the array run the function to extract the value

robertsLando (changes requested):

kvaster:
> @kvaster would you be able to test the implementation of this PR? It would be good to have some feedback before merging it.

Yes. Will try it shortly! Thanks!

lib/ZwaveClient.ts
ahochsteger: @robertsLando , yes, that's a good idea - much cleaner and more flexible :-).
I may have to separate the function for value mapping from the function that checks for the existence of a command class.
E.g.:
```typescript
const mappedNodeProps = {
  [CommandClasses.Battery]: [
    {
      nodeProp: 'batteryLevelMin',
      valueFn: (node, value) => { ... }
    },
    {
      nodeProp: 'powerSource',
      existsFn: (node) => { ... }
    },
  ]
}
```

robertsLando: Or you can manage it by make the fn return an object, try and check what makes it more cleaner :)

kvaster:
Installed version from this PR and it's working!

ahochsteger:
@robertsLando, I'm finally done with the PR.
It took a bit more work to finally get everything working as expected and to not degrade the performance since there were still some places having to iterate values (-> please have a special look at the changes in ZWaveClient.ts). But I'm finally satisfied with the result and even the grouping by values of the power column works now :-).
@kvaster, it would be great, if you could give this PR one more try so we can finally merge it!

kvaster:
> @robertsLando, I'm finally done with the PR. It took a bit more work to finally get everything working as expected and to not degrade the performance since there were still some places having to iterate values (-> please have a special look at the changes in ZWaveClient.ts). But I'm finally satisfied with the result and even the grouping by values of the power column works now :-). @kvaster, it would be great, if you could give this PR one more try so we can finally merge it!

Compiling! Will get back very soon. And thanks!

ahochsteger:
Since the way has been paved to show enriched values in the table I used it to display boolean values for `beamingSupported` and `failed` as colored icons (using true/false/undefined).

kvaster:
Tried latest version and it works good. Thanks a lot!

ahochsteger:
@robertsLando the PR is now ready for review/approval.
Note, that I've also used the new `RichValue` feature to show icons for more columns to add more eye candy (see screenshot in the description) :-).

robertsLando:
@ahochsteger Just what's your progress on this? I'm going to make a new release but If we end this soon I can hold a bit for this to land :)

ahochsteger:
@robertsLando all Feedback seems done now - or is there something I've missed?

PR #1825, opened 2021-10-11
chore: switch to yarn 2

darkbasic:
Notes (blhoward2):
```
If zjs2mqtt is switching over to yarn v2, could you please take a look at the Dockerfile.contrib and make any changes necessary there?
We use that to build custom dockers off any branch on either project
```

### DIFF
--- a/lib/Gateway.ts
+++ b/lib/Gateway.ts
@@ -271,6 +271,8 @@ export default class Gateway {
 				this._zwave.on('event', this._onEvent.bind(this))
 			}
 
+			this._zwave.on('driverStatus', this._onDriverStatus.bind(this))
+
 			// this is async but doesn't need to be awaited
 			await this._zwave.connect()
 		} else {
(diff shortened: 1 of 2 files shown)


AlCalzone:
You need the .yarn folder. The yarn documentation mentions what should be in gitignore and what not.

darkbasic:
Turns out Dockerfile.contrib was already broken: I've fixed it and made it compatible with berry. By the way, will they ever make the equivalent of v1 --production flag for non-workspaces projects?

darkbasic:
Yes I tested it with darkbasic:feat/yarn2.

blhoward2:
Maybe this is better all around. Note the entry about a target: https://stackoverflow.com/questions/43654656/dockerfile-if-else-condition-with-external-arguments

We'd need to update the docs too and it's a breaking change, which I'm not sure is that big of a deal here.

darkbasic:
@blhoward2 turns out there was already a comment warning about buildkit but somehow I completely skipped it.
The Dockerfile pointed at the Readme which didn't contain any instructions, so I've linked the website instead.

PR #1917, opened 2021-11-02
feat: smart start

### LINKED ISSUE #1895: [feat] DSK input improvements

There seems to be a very short time-frame (down to 8 seconds) to enter the DSK and confirm it, when adding some device with S2_*.

Suggest the following UI improvements, to be able complete the process more quickly:

- Auto focus the DSK input field when it appear
- Allow submitting the DSK with the enter-key
- Display a numeric keyboard on mobile phones

### LINKED ISSUE #1916: [feat] Smart start support
Ref: https://zwave-js.github.io/node-zwave-js/#/getting-started/security-s2
https://zwave-js.github.io/node-zwave-js/#/api/controller?id=provisionsmartstartnode

robertsLando:
Fixes #1895
Fixes #1916


src/lib/utils.js
```diff
@@ -5,3 +5,63 @@ export function copy(o) {
 export function wait(ms) {
 	return new Promise((r) => setTimeout(r, ms))
 }
+
+export function parseSecurityClasses(securityClasses) {
+	const classes = {
+		s2AccessControl: false,
+		s2Authenticated: false,
+		s2Unhauntenticated: false,
```
kpine, line 13:
```suggestion
		s2Unauthenticated: false,
```

kpine:
Would it be possible to add another import option, by text? Most camera qr code scanners can convert the image to a text string, which is the QR code that is input to zwave-js.

kpine:
Hope you don't mind some comments in this current change. When I import a QR code by image, it seemed to be accepted (after having to crop it from the original). However, the screen does not proceed and the bar at the top is going back and forth as if its doing something. It's not obvious if I need to do anything. Closing it was fine. Perhaps showing the provisioning list or proceeding to the "Done" phase and indicating success would be better.

image

kpine:
Other possible enhancements:

- Allow a user to specify node name and location when adding the QR code. The PlannedProvisioningEntry allows properties to be attached to the entry, and those should be available in the provision entries for access after inclusion. zwavejs2mqtt could auto-name the node and set location if these are set.
- Show the Mfg & Product info in the provisioning list. The latter are available from the scanned QR code and would help quickly identify which entry is which. Also show the name/location if set.
- Include the provisioning list into the debug dump. Perhaps exclude the DSK or only include part of it.

blhoward2:
I like this idea:

“Show the Mfg & Product info in the provisioning list. The latter are available from the scanned QR code and would help quickly identify which entry is which. Also show the name/location if set.”

robertsLando:
> Would it be possible to add another import option, by text? Most camera qr code scanners can convert the image to a text string, which is the QR code that is input to zwave-js.

Sure!

> Closing it was fine. Perhaps showing the provisioning list or proceeding to the "Done" phase and indicating success would be better.

I added the progress bar as in reality it is waiting to the node to be added, if you stay there a while you will see also the node added dialog. I could remove the loading if you think it's not necessary

> Include the provisioning list into the debug dump. Perhaps exclude the DSK or only include part of it.

What If I add a button to export it from the smart start tab?

> Allow a user to specify node name and location when adding the QR code. The PlannedProvisioningEntry allows properties to be attached to the entry, and those should be available in the provision entries for access after inclusion. zwavejs2mqtt could auto-name the node and set location if these are set.

It's ok for me, would like to know @AlCalzone opinion on this

> Show the Mfg & Product info in the provisioning list. The latter are available from the scanned QR code and would help quickly identify which entry is which. Also show the name/location if set.

The provisioning list shows items returned from `getProvisioningEntries` method that returns an array of `SmartStartProvisioningEntry` that has only those info (securityClasses, dsk, nodeId) . Dunno How/if I can get the other informations that are present in qrCode parse `QRProvisioningInformation`

AlCalzone:
@robertsLando 
If the entry was scanned from a QR code and z2m didn't strip out the extra information, the required properties and some of the optional properties in [`QRProvisioningInformation`](https://zwave-js.github.io/node-zwave-js/#/api/utils?id=parse-s2-or-smartstart-qr-code-strings) will be included in the provisioning entry.
I would still test if they are present, just to be sure.

here you go for the manufacturer and label:
https://github.com/AlCalzone/ioBroker.zwave2/blob/1664a0a41c97bea44f9bf5806dde7a532624f379/src/main.ts#L1553-L1576
which could look like this:
grafik

You're free to add extra properties like `name` and `location`.

robertsLando:
Don't ask me why but vscode has automatically merged branch on master when doing a checkout to master branch, I have recreated the PR here: #1922

PR #1924, opened 2021-11-03
fix(ui): fix typo on add/remove dialog

luckyjay:
Add/remove dialog S2 Authenticated example mentioned "lightning". Assume that was supposed to be "lighting".


luckyjay:
Sorry for the goof-up on .gitpod.yml. Still learning the system.

robertsLando (approved):

PR #1926, opened 2021-11-04
feat: add node added dialog and fix replace node

### LINKED ISSUE #1923: [bug] Smart start fixes
- [x] Rename smart start in Inclusion by QR Code
- [x] Replacing a node by scanning a QR code should never add the QR code to the provisioning list. You'll end up with two nodes, one of which is dead and the other has a new ID. Instead just do an S2 inclusion even if the QR code is SmartStart
- [x] Can't replace a failed node that has a DSK. It will tell me that the node is already added, although I'm just trying to replace it
- [x] Add dialog that shows when a node is successfully added

robertsLando:
Fixes #1923


robertsLando:
cc @AlCalzone

PR #1929, opened 2021-11-05
chore(docker): use in built docker gha caching for build

chrisns:
caching is probably less relevant really for this now we're not waiting on ozw to compile, and with zwjs is changing so much, we probably never benefit from the cache, so you could remove this entirely but this might at least simplify the steps in the meantime


robertsLando (approved):
I remember last time we did so much work to improve caching and make ci faster. Let's see how this goes and if it's slower we can revert. LGTM

PR #1970, opened 2021-11-18
feat: optimize socket events and mqtt disabled events

robertsLando:
BREAKING CHANGE: Some socket and zwave client events names and content have changed. This will not affect normal users but could cause some breaking changes on users using custom Plugins:
- `NODE_UPDATED` socket event now contains the entire node or the props that have changed
- `nodeInited` new zwave client event emitted when node is inited
- `valueWritten` event signature has changed: `valueWritten: (valueId: Z2MValueId, node: Z2MNode, value: unknown) => void`

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -1316,9 +1316,9 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		if (node.ready && !node.inited) {
 			node.inited = true
 			this.emit('nodeInited', node)
-		} else {
-			this.emit('nodeStatus', node)
 		}
+
+		this.emit('nodeStatus', node)
 
 		if (changedProps) {
 			// we need it to have a reference of the node to update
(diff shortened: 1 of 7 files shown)


lib/ZwaveClient.ts
```diff
@@ -3695,6 +3750,7 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 				const values = Object.values(vIds)
 				const result = vMap.fn(node, values)
 				node[vMap.nodeProp] = result
+				updatedProps[vMap.nodeProp] = result
```
robertsLando, line 3753: cc @ahochsteger

lib/ZwaveClient.ts
```diff
@@ -3708,6 +3764,8 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 				}
 			}
 		)
+
+		this.emitNodeStatus(node, updatedProps)
```
robertsLando, line 3768: @ahochsteger  here is where props are sent to socket

PR #2005, opened 2021-11-28
feat(hass): expose binary sensors in Configuration command set in Fibaro TRVs

msbg:


msbg:
Latest 4.7 version of fibaro firmware for the FGT001 TRV exposes binary sensors for 'heat' demand in the configuration command class (https://github.com/zwave-js/node-zwave-js/commit/4b0ce9b613e63a683b25536456cb04ec7e0bb646).  This is my first attempt at exposing these as binary sensors to HASS (given that the 'heat demand' call can then be used to trigger the boiler).
If there is a better place/template for this code, or if I should be more guarded in making sure this logic is only entered in the case of an FGT001 device please let me know.  I think I also need someone to 'allow' me to run the ci pipeline steps, and I'm not sure how to fix the PR naming failure - any suggestions on how to test/re-trigger the 'Semantic PR Title' app now I've renamed both the commit and the PR appreciated.

Many thanks,

- monica

robertsLando:
Hi Monica, thanks for your PR. It looks good but the problem is this is something really device specific and cold also be easy to break as it's using properties that are actually coming from `propertyName` and colud change. For such applications with hass I generally suggest this: https://zwave-js.github.io/zwavejs2mqtt/#/homeassistant/homeassistant-mqtt?id=manually-create-entities

msbg:
Makes sense - thanks for the confirmation on the approach. Appreciated.

PR #2069, opened 2021-12-16
feat: show driver errors on UI and prevent restart when zwave options are not valid

### LINKED ISSUE #2063: [bug] Security key used multiple times isn't clearly identified as a fatal error
### Version

**Checklist:**

- [ ] I am **not** using Home Assistant. **Or:** a developer has told me to come here.
- [x] I have read the changelog and my problem is **not** mentioned there.

Build/Run method

- [x] Docker
- [ ] PKG
- [ ] Snap package

Zwavejs2Mqtt version: 6.1.0
Z-Wave JS version: 8.9.0-beta.0

### Describe the bug

Slight UX issue - If the same Zwave security key is used more than once, the Zwave driver won't start up, but it isn't clearly identified as a fatal error.  The message from the driver is logged at level INFO.  The logs indicate that the connection is being continually retried every 3 seconds. This gives the impression there is some sort of connection problem rather than a user configuration error.

```
2021-12-15 17:03:14.776 INFO ZWAVE: Controller status: Driver: The security key for class S2_Unauthenticated was used multiple times! (ZW0108)
2021-12-15 17:03:14.779 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:14.789 INFO APP: POST /api/settings 200 68.056 ms - 950
2021-12-15 17:03:14.850 INFO APP: GET /api/auth-enabled 304 6.210 ms - -
2021-12-15 17:03:17.783 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:20.787 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:23.794 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:26.800 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:29.803 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:32.809 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:35.813 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:38.818 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:41.830 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:44.834 WARN ZWAVE: Retry connection in 3 seconds...
2021-12-15 17:03:47.840 WARN ZWAVE: Retry connection in 3 seconds...
```

### To Reproduce

1. Enter the same network key in S0 Legacy and S2 Unauthenticated.
2. Click on Save
3. View the logs to see the error

### Expected behavior

It should be clearer that there is a fatal configuration error and that zwavejs will never successfully start.  Some ideas:

- ZwaveJ22MQTT UI shouldn't allow a save of the zwave config panel if the same key is used multiple times
- The error from the driver should be shown as a fatal error
- The server shouldn't keep trying to connect
- The driver error message in the UI should be highlighted as an error (red text or something like that)

image

robertsLando:
Fixes #2063
cc @AlCalzone

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -1286,12 +1286,20 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 						)
 					})
 				}
-				await this._onDriverError(error, true)
-				logger.warn('Retry connection in 3 seconds...')
-				this.reconnectTimeout = setTimeout(
-					this.connect.bind(this),
-					3000
-				)
+
+				if (error.code !== ZWaveErrorCodes.Driver_InvalidOptions) {
+					await this._onDriverError(error, true)
+					logger.warn('Retry connection in 3 seconds...')
+					this.reconnectTimeout = setTimeout(
+						this.connect.bind(this),
+						3000
+					)
+				} else {
+					logger.error(
+						`Invalid options for driver: ${error.message}`,
+						error
+					)
+				}
 			}
 		} else {
 			logger.info(`Driver already connected to ${this.cfg.port}`)
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -412,6 +412,7 @@ export type Z2MDriverInfo = {
 	appVersion?: string
 	zwaveVersion?: string
 	serverVersion?: string
+	error?: string | undefined
 	homeid?: number
 	name?: string
 	controllerId?: number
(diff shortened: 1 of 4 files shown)


AlCalzone (changes requested):
Isn't there a way to make this more obvious in the frontend?
Like an error dialog, or sending the user to the config page?

robertsLando:
@AlCalzone Yeah didn't get that far yet, I tagged you for your opinion about the invalid driver option check, is there some other error that could be thrown when initing driver that should prevent restart?

AlCalzone:
A bunch, but they all use the same error code:
https://github.com/zwave-js/node-zwave-js/blob/5575f4bc4aba4d9795a698e86e5c6f5bac5c8725/packages/zwave-js/src/lib/driver/Driver.ts#L227-L328

PR #2120, opened 2021-12-30
feat: use new `restoreNVM` API to restore NVM backups

### LINKED ISSUE #2119: [feat] Use new `restoreNVM` API to restore NVM backups

Z2M currently uses the `restoreNVMRaw` API to restore a backed up NVM. This API requires the the source and target controller NVMs are compatible. node-zwave-js v8.9.0 added a new [`restoreNVM` API](https://zwave-js.github.io/node-zwave-js/#/api/controller?id=nvm-backup-and-restore) which automatically converts the NVM backup to the correct target format. Using the new one would allow users to backup from a 500 and restore to a 700, for example.

Unless I'm missing something, there doesn't seem to be any reason for applications to use `restoreNVMRaw` anymore with this new API available.

robertsLando:
Fixes #2119

cc @kpine

### DIFF
--- a/docs/guide/mqtt.md
+++ b/docs/guide/mqtt.md
@@ -93,7 +93,7 @@ This are the available APIs:
   - `args`: array of arguments to pass to the command
 - `restart()`: restart client
 - `backupNVMRaw()`: Backup the NVM raw data
-- `restoreNVMRaw(data)`: Restore the NVM raw data
+- `restoreNVM(data)`: Restore the NVM data. If the given buffer is in a different NVM format, it is converted automatically. If the conversion is not supported, the operation fails.
 - `softReset`: Soft reset the controller (restart)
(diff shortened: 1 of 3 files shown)


AlCalzone (commented):
@robertsLando 
https://zwave-js.github.io/node-zwave-js/#/api/controller?id=nvm-backup-and-restore

You're not displaying the entire progress now... Only the NVM conversion part, but not the restore progress that begins afterwards.

```ts
restoreNVM(
    nvmData: Buffer,
    convertProgress?: (bytesRead: number, total: number) => void,
    restoreProgress?: (bytesWritten: number, total: number) => void, // <-- THIS is missing!
): Promise<void>
```

robertsLando:
Fixed on master, thanks

PR #2194, opened 2022-01-17
feat: set node name and location before inclusion and keep `nodes.json` up to date

### LINKED ISSUE #2129: [bug] nodes.json contains invalid/old data
### Version

**Checklist:**

- [x] I am **not** using Home Assistant. **Or:** a developer has told me to come here.
- [x] I have read the changelog and my problem is **not** mentioned there.

Build/Run method

- [ ] Docker
- [ ] PKG
- [x] Snap package

zwavejs2mqtt: 6.2.0.fa6b2ab
zwave-js: 8.9.3-0-pr-4011-940427a
home id: 3685395904
home hex: 0xdbaaadc0

### Describe the bug

`nodes.json` contain nodes that is removed from the network. Shouldn't cause big issues, but I guess when the controller reaches #232 it will start re-use node id's, then causing new nodes using old `nodes.json` data.

### Expected behavior

`nodes.json` to be in sync with the nodes in the network.

### LINKED ISSUE #2183: [feat] Allowing setting name/location before regular inclusion

HA parses and sets up entity domains based on the device's name and location at the time it receives a Node Ready. Subsequent name changes in zjs2mqtt are updated for the device name in HA (so long as it hasn't been renamed there), but the entity domains aren't updated after creation. 

Allow the user to preset the name/location prior to inclusion like they can with SmartStart. Set this as soon as possible post-inclusion but before the Node becomes ready.

robertsLando:
Fixes #2183
Fixes #2129

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -207,10 +207,7 @@ export type SensorTypeScale = {
 
 export type AllowedApis = typeof allowedApis[number]
 
-const ZWAVEJS_LOG_FILE = utils.joinPath(
-	process.env.ZWAVEJS_LOGS_DIR || storeDir,
-	'zwavejs_%DATE%.log'
-)
+const ZWAVEJS_LOG_FILE = utils.joinPath(storeDir, 'zwavejs_%DATE%.log')
 
 export type Z2MValueIdState = {
 	text: string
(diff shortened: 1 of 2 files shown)


robertsLando:
cc @blhoward2 @jmgiaever

PR #2259, opened 2022-02-14
Dependabot/npm and yarn/eslint plugin import 2.25.4

Multipas1:


robertsLando:
@Multipas1 Could you explain this?

PR #2294, opened 2022-03-01
feat: bump zwave-js@9 and @zwave-js/server@1.16.0 and fix breaking changes (#2294)

robertsLando:
cc @AlCalzone

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -3,7 +3,6 @@ import express, { Request, Response, RequestHandler, Router } from 'express'
 import morgan from 'morgan'
 import csrf from 'csurf'
 import { SerialPort } from 'serialport'
-import { PortInfo } from '@serialport/bindings-cpp'
 import jsonStore from './lib/jsonStore'
 import cors from 'cors'
 import ZWaveClient, {
@@ -879,15 +878,13 @@ app.get(
 			scales: scales,
 		}
 
-		let ports: PortInfo[]
 		if (process.platform !== 'sunos') {
 			try {
-				ports = await SerialPort.list()
+				data.serial_ports = (await SerialPort.list()).map((p) => p.path)
 			} catch (error) {
 				logger.error(error)
+				data.serial_ports = []
 			}
-
-			data.serial_ports = ports?.map((p) => p.path) ?? []
 			res.json(data)
 		} else res.json(data)
 	}
--- a/app.ts
+++ b/app.ts
@@ -2,7 +2,8 @@ import express, { Request, Response, RequestHandler, Router } from 'express'
 
 import morgan from 'morgan'
 import csrf from 'csurf'
-import SerialPort from 'serialport'
+import { SerialPort } from 'serialport'
+import { PortInfo } from '@serialport/bindings-cpp'
 import jsonStore from './lib/jsonStore'
 import cors from 'cors'
 import ZWaveClient, {
@@ -878,7 +879,7 @@ app.get(
 			scales: scales,
 		}
 
-		let ports: SerialPort.PortInfo[]
+		let ports: PortInfo[]
 		if (process.platform !== 'sunos') {
 			try {
 				ports = await SerialPort.list()
--- a/docker/Dockerfile.contrib
+++ b/docker/Dockerfile.contrib
@@ -42,8 +42,7 @@ RUN yarn
 RUN yarn build
 
 # Update the version info to match the branch built
-RUN yarn lerna version --allow-branch "*" --no-git-tag-version --no-push --ignore-scripts --yes \
-    $(echo $(yarn node -p 'require("semver").inc(require("zwave-js/package.json").version, "prerelease")+".dev"')-$(git rev-parse --short HEAD))
+RUN RUN yarn workspaces foreach version $(echo $(yarn node -p 'require("semver").inc(require("zwave-js/package.json").version, "prerelease")+".dev"')-$(git rev-parse --short HEAD)) --deferred && yarn version apply --all
 
 RUN yarn workspaces foreach pack
 
--- a/docker/Dockerfile.contrib
+++ b/docker/Dockerfile.contrib
@@ -54,6 +54,7 @@ WORKDIR /home/node/zwavejs2mqtt
 RUN cat package.json \
       | jq '.resolutions += { "@zwave-js/config": "file:../node-zwave-js/packages/config/package.tgz" }' \
       | jq '.resolutions += { "@zwave-js/core": "file:../node-zwave-js/packages/core/package.tgz" }' \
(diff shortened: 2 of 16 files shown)


app.ts
```diff
@@ -878,7 +879,7 @@ app.get(
 			scales: scales,
 		}
 
-		let ports: SerialPort.PortInfo[]
+		let ports: PortInfo[]
```
AlCalzone, line 882: You could avoid having to import `@serialport/bindings-cpp` at all if you did this:
```js
try {
  data.serialports = (await SerialPort.list()).map((port) => port.path);
} catch (error) {
  logger.error(error)
  data.serialports = [];
}

res.json(data)
```

package.json
```diff
@@ -92,8 +92,6 @@
     }
   },
   "resolutions": {
-    "serialport": "~9.0.7",
-    "@serialport/bindings": "~9.0.8",
     "colors": "1.4.0"
```
AlCalzone, line 95: AFAIK npm has reverted the changes to colors, so you can remove this too

docker/Dockerfile.contrib
```diff
@@ -41,9 +41,10 @@ ENV YARN_HTTP_TIMEOUT=300000
 RUN yarn
 RUN yarn build
 
+RUN yarn plugin import version && yarn plugin import workspace-tools
+
 # Update the version info to match the branch built
-RUN yarn lerna version --allow-branch "*" --no-git-tag-version --no-push --ignore-scripts --yes \
-    $(echo $(yarn node -p 'require("semver").inc(require("zwave-js/package.json").version, "prerelease")+".dev"')-$(git rev-parse --short HEAD))
+RUN yarn workspaces foreach version $(echo $(yarn node -p 'require("semver").inc(require("zwave-js/package.json").version, "prerelease")+".dev"')-$(git rev-parse --short HEAD)) --deferred && yarn version apply --all
```
AlCalzone, line 48: Does this work now?

AlCalzone (changes requested):

src/components/custom/StatisticsArrows.vue
```diff
@@ -48,7 +48,16 @@ export default {
...
 	methods: {
-		jsonToList,
+		jsonToList(item) {
+			item.indented = {
+				a: 'c',
+				b: 'd',
+			}
+			return jsonToList(item, {
+				rtt: 'ms',
+				rssi: 'dBm',
```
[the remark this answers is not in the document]
AlCalzone: New v9 test build coming up, this time with `rssiToString(...)` exported from `zwave-js` and `zwave-js/Utils`

docker/Dockerfile.contrib
robertsLando: Yep, at least the build completes successfully

AlCalzone (approved):

AlCalzone:
Bump to beta 6 please

raman325:
Published zwave-js-server 1.16.0-beta.1 which supports zwave-js 9 as well

robertsLando:
@AlCalzone @raman325 Done :) 

@AlCalzone Now I also display sdkVersion in table firmware version

src/components/nodes-table/index.vue
```diff
@@ -240,7 +240,12 @@
 						item.firmwareUpdate.progress
 					}}</span>
 				</v-progress-circular>
-				<div v-else>v{{ item.firmwareVersion }}</div>
+				<div v-else>
+					v{{
+						item.firmwareVersion +
+						(item.sdkVersion ? ` (${item.sdkVersion})` : '')
```
AlCalzone, line 246: 1.: `v` before the firmware version, but no `v` before the SDK version?
2.: I would clearly indicate which is which. firmware version and sdk version have no correlation on 500 series sticks. They do on 700 series sticks with stock firmware, but for Razberry I'd expect them to be different too - they compile their own.

How about:
```
FW: vA.B
SDK: vX.Y.Z
```

AlCalzone (changes requested):

kpine:
I'm testing this build (container metadata says `"org.opencontainers.image.revision": "4969aaeb9d58843396fd0c519a510b203526c545"`), but I'm not seeing any SDK version reflected in the Version column.

image

I also notice there's no `sdkVersion` entry in the node's Debug JSON. Should that be there? The driver logs show it being detected for the controller. I don't see anything in the driver cache either (besides `firmwareVersions`), not sure if it would be stored there anyways?

```
2022-03-17T21:14:09.203Z CNTRLR   received protocol version info:
                                    protocol type:             Z-Wave
                                    protocol version:          7.17.2
                                    appl. framework build no.: 406
                                    git commit hash:           30303030303030303030303030303030
```

docker/Dockerfile.contrib
```diff
@@ -16,7 +16,7 @@ ARG Z2M_REPOSITORY=https://github.com/zwave-js/zwavejs2mqtt
 USER node
 WORKDIR /home/node
 
-RUN git clone -b ${ZWJ_BRANCH} --depth 1 ${ZWJ_REPOSITORY}
+RUN git clone -b ${ZWJ_BRANCH} ${ZWJ_REPOSITORY}
```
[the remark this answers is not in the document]
robertsLando: Couldn't you use `npm pkg get version`?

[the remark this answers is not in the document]
AlCalzone: `yarn version` is used to update the versions in all `package.json`. There's probably a hacky way to do it with `jq` instead if you don't want to clone all history.

robertsLando:
Answering to @kpine 

@AlCalzone Could it be that sdkVersion value comes after node-ready event?

docker/Dockerfile.contrib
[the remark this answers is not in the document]
robertsLando: 👍🏼

AlCalzone:
Its queried during the controller interview which the driver does immediately after soft-reset/startup. The value isn't stored anywhere except the controller class instance, so it doesn't appear in the cache.

AlCalzone:
@robertsLando could it by that you're trying to read the `sdkVersion` from the controller node instead of `driver.controller`? There is no special handling for that yet, I forgot to add it.

robertsLando:
> could it by that you're trying to read the sdkVersion from the controller node instead of driver.controller? There is no special handling for that yet, I forgot to add it.

It's that for sure, I'm not reading sdkVersion from driver.controller anywhere as I saw you added it also to nodes. I think it would make sense to add it to controller node too for consistency

AlCalzone:
beta.7 coming up in a few minutes.

robertsLando:
Bumped

robertsLando:
Docker releases of this branch availabe with `zwavejs-9` branch

PR #2317, opened 2022-03-13
feat(ui): flatten UI when elevation is not necessary (#2317) by @floutchito

floutchito:
I guys.

I'm using this piece of code every day via home assistant, and I love it !

However, I noticed minor possible enhancements that I propose to share with this PR.
Comments and remarks are welcome.

Here's what is modified :

- [x] A "Home Assistant mode" for UI navigation (using tabs in the top app bar instead of a navigation drawer, to prevent dual drawer when integrated in Home Assistant)
- [x] Flatten UI when elevation is not necessary. There are places in the app where logical containers do not need elevation, this PR proposes to flatten these for a better readability

image

- [x] There is also a new UI section in the settings page where I added the new option "Use tabs for navitation" (aka "HA mode"), and moved the dark mode toggle that was previously in the nav drawer

image

- [x] Refactored controller statistics panel and actions menu

Statistics panel now shows all possible stats grayscaled when empty, and blue / red colors are applied when a value is set.
This makes the panel size, and the stats positions predictable over time, whereas it does not visually take more place.

I think it is useful when someone is quickly searching for a specific stat.
Stats are also grouped by logical sections: `MESSAGES`, `COMMANDS`, `COMMUNICATION`.

image
image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,9 +1,10 @@
 <template>
-	<v-app :dark="dark">
+	<v-app :dark="darkMode">
 		<div
 			v-if="$route.meta.requiresAuth && auth !== undefined && !hideTopbar"
 		>
 			<v-navigation-drawer
+				v-if="!navTabs || $vuetify.breakpoint.smAndDown"
 				clipped-left
 				:mini-variant="mini"
 				v-model="drawer"
@@ -29,7 +30,7 @@
 					<v-list-item
 						v-for="item in pages"
 						:key="item.title"
-						:to="item.path == '#' ? '' : item.path"
+						:to="item.path === '#' ? '' : item.path"
 						:color="item.path === $route.path ? 'primary' : ''"
 					>
 						<v-list-item-action>
@@ -42,13 +43,6 @@
 							>
 						</v-list-item-content>
 					</v-list-item>
-					<v-list-item v-if="!mini">
-						<v-switch
-							label="Dark theme"
-							hide-details
-							v-model="dark"
-						></v-switch>
-					</v-list-item>
 				</v-list>
 				<v-footer absolute v-if="!mini" class="pa-3">
 					<div>
@@ -58,10 +52,35 @@
 			</v-navigation-drawer>
 
 			<v-app-bar app>
-				<v-app-bar-nav-icon @click.stop="toggleDrawer" />
-				<v-toolbar-title v-if="$vuetify.breakpoint.smAndUp">{{
-					title
-				}}</v-toolbar-title>
+				<template v-if="!navTabs || $vuetify.breakpoint.smAndDown">
+					<v-app-bar-nav-icon @click.stop="toggleDrawer" />
+					<v-toolbar-title v-if="$vuetify.breakpoint.smAndUp">
+						{{ title }}
+					</v-toolbar-title>
+				</template>
+				<template v-else>
+					<v-tabs>
+						<v-tab
+							v-for="item in pages"
+							:key="item.title"
+							:to="item.path === '#' ? '' : item.path"
+							class="px-1"
+						>
+							<v-icon
+								left
+								:small="item.path === $router.currentRoute.path"
+							>
+								{{ item.icon }}
+							</v-icon>
+							<span
+								v-if="item.path === $router.currentRoute.path"
+								class="subtitle-2"
+							>
+								{{ item.title }}
+							</span>
+						</v-tab>
+					</v-tabs>
+				</template>
 
 				<v-spacer></v-spacer>
 
@@ -312,7 +331,6 @@ import io from 'socket.io-client'
 import ConfigApis from '@/apis/ConfigApis'
 import Confirm from '@/components/Confirm'
 import PasswordDialog from '@/components/dialogs/Password'
-import { Settings } from '@/modules/Settings'
 import { Routes } from '@/router'
 
 import { mapActions, mapMutations, mapGetters } from 'vuex'
@@ -326,7 +344,7 @@ export default {
 	},
 	name: 'app',
 	computed: {
-		...mapGetters(['user', 'auth', 'appInfo']),
+		...mapGetters(['user', 'auth', 'appInfo', 'navTabs', 'darkMode']),
 		updateAvailable() {
 			return this.appInfo.newConfigVersion ? 1 : 0
 		},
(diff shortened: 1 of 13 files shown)


floutchito:
This is a Draft, I'm searching for an easy way to use the HA zwave-js-server to continue this work on node views, but it seems I can't link the dev server to an existing prod instance embedded in HA.

My HA instance and its ZWJS2MQTT plugin are dockerized by Supervisor.

_**Do some of you guys know a way to do so ?**_

robertsLando:
Hiu @floutchito ! Really thanks for your PR, love it! I agree the UI can be improved a lot, like you can see I'm not an UI fun, I'm more a backend guy 😆

> I'm searching for an easy way to use the HA zwave-js-server to continue this work on node views, but it seems I can't link the dev server to an existing prod instance embedded in HA.

About your question, you should get the ip address of the docker container that is running the webserver and set it here:

https://github.com/zwave-js/zwavejs2mqtt/blob/master/src/apis/ConfigApis.js#L10

Let me know if you need something else! I will give a try to your PR soon

robertsLando:
I tested your PR. Everything looks good, I liked all the style you used. The only thing that seems not working is controller stats, when I enable it nothing is shown

src/App.vue
```diff
@@ -845,3 +860,9 @@ export default {
 	},
 }
 </script>
+
+<style scoped>
+.smaller-min-width-tabs {
```
robertsLando, line 865: In order to make this working with `scoped` you should use `>>>` operator

```
.v-tabs >>> .smaller-min-width-tabs
```

robertsLando (changes requested):

floutchito:
> Hiu @floutchito ! Really thanks for your PR, love it! I agree the UI can be improved a lot, like you can see I'm not an UI fun, I'm more a backend guy 😆
> 
> > I'm searching for an easy way to use the HA zwave-js-server to continue this work on node views, but it seems I can't link the dev server to an existing prod instance embedded in HA.
> 
> About your question, you should get the ip address of the docker container that is running the webserver and set it here:
> 
> https://github.com/zwave-js/zwavejs2mqtt/blob/master/src/apis/ConfigApis.js#L10
> 
> Let me know if you need something else! I will give a try to your PR soon

Thanks !

I'll give a try at setting the IP of my prod server.
Therefore, I would be able to test statistics and node views, I must say that I wasn't able to test it with real data.

Will work on this at night.

robertsLando:
> Will work on this at night.

Perfect! Thanks :)

floutchito:
Hi there.

<details>
  <summary>I finally managed to use my HA `zwave-js-server` linked to a dev instance of `zwjs2mqtt` !</summary>
  
I wasn't able to connect to the addon's IP, 'cause my HA is on a specific host, and addons are isolated in a specific Docker network.
I tryied to connect from HA's external URL, but I got trouble with CORS issues.
Even after I bypassed the CORS headers (with CORS anywhere or MITMProxy), I had `401 Unauthorized` errors, that was thrown by the addon's internal nginx.

I had to forward the internal port 44290 from inside of the `hassio` Docker network, through socat and expose the result to the host. After that I was able to bypass the nginx-level security built-in the HA add-on. 

Here's a docker-compose example :

```yaml
version: '3.7'

services:
      
  zwave-server-forwarder:
    image: "alpine/socat"
    command: "tcp4-listen:8099,fork,reuseaddr,su=nobody tcp-connect:172.XX.YY.ZZ:44920"
    networks:
      hassio:
    ports:
      - "8099:8099"
      
networks:
  hassio:
    external: true
```  
</details>

And after a bit of work, I ended with the following result :


### Working StatisticsCard
image


### Reworked Store view
I tried to use the available space instead of limiting the file browser to a max of 800px.
I tested it in many ways, and I couldn't find a situation where I missed controls or view.

Screenshot 2022-03-20 at 14-56-33 ZWave To MQTT

### Reworked inner Nodes view
Here, I tried to maximize the number of infomations available at the same time.
To do so, I switched the sections to vertical tabs, in order to always see the different available sections.
I also choosed to pull `Device ID` and the `PING` and `ADVANCED` actions buttons out of NodeDetails view.
I thought it should be useful to access these generic actions from anywhere in the opened node view.
These are obviously opinionated  choices, let me know if it makes sense to you.

Screenshot 2022-03-20 at 14-53-08 ZWave To MQTT

I also flattened the expansion panels to match the Settings view styles
Screenshot 2022-03-20 at 14-57-25 ZWave To MQTT

### Fixed The small overflow in AppBar

I didn't modified the graph view, and I do not use Scenes, so I can't see if there are view that I can work on.

Let me know if there are thing that needs to be rolledback or modified.

robertsLando (changes requested):
@floutchito Firstly thanks for this PR, just love it!

I tested your changes and everything looks good! Just two things:

1. On smaller screen I think it makes sense to have node tabs (in node details) horizontally instead of vertically, could you change that?
2. The controller stats button seems like it's not clickable when stats are hidden, I would change the disabled color to something more user friendly so the users can detect it as something clickable

The rest LGTM :)

floutchito:
Hi @robertsLando,

> 1.     On smaller screen I think it makes sense to have node tabs (in node details) horizontally instead of vertically, could you change that?

Done. I had to align left the AssociationGroups table to let users understand the structure on smaller screen.

> 2.     The controller stats button seems like it's not clickable when stats are hidden, I would change the disabled color to something more user friendly so the users can detect it as something clickable

Done. I also added a toggle for each individual node to show / hide statistics.

robertsLando (approved):
LGTM. Thanks @floutchito

AlanGreene:
@floutchito I just updated to release 6.6.0 in my Home Assistant instance and noticed the Dark Mode setting doesn't apply when I first load the page. To enable dark mode I must first navigate to settings, then to any other Z-Wave JS page. It remains enabled as long as I stay within Z-Wave JS. As soon as I navigate away to another page in HA and come back to Z-Wave JS it's reverted to displaying in light mode again.

Let me know if there's any particular info you need to help debug, happy to provide any logs / network traces, versions etc. that may be of use.

floutchito:
> @floutchito I just updated to release 6.6.0 in my Home Assistant instance and noticed the Dark Mode setting doesn't apply when I first load the page. To enable dark mode I must first navigate to settings, then to any other Z-Wave JS page. It remains enabled as long as I stay within Z-Wave JS. As soon as I navigate away to another page in HA and come back to Z-Wave JS it's reverted to displaying in light mode again.
> 
> Also the Save button in settings isn't reachable in the HA Android app, it's rendered off the right edge of the screen.
> 
> Let me know if there's any particular info you need to help debug, happy to provide any logs / network traces, versions etc. that may be of use.

Hello,

I can reproduce this bug. I'll look at it as soon as possible.

PR #2333, opened 2022-03-24
fix(ui): initialize darkMode at startup

floutchito:
Dark mode was not initialized to the stored value at startup

Fixes https://github.com/zwave-js/zwavejs2mqtt/pull/2317#issuecomment-1076825090


robertsLando (approved):
Thanks for the fix @floutchito I didn't catch that bug on my review

PR #2346, opened 2022-04-04
feat: add setting to disable zwave-js-server DNS-SD

raman325:
As discussed, adding a new option that can be passed into to `zwavejs2mqtt` to disable DNS service discovery in zwave-js-server. This option is primarily to be used by @Frenck's Home Assistant addon for zwavejs2mqtt: https://github.com/hassio-addons/addon-zwavejs2mqtt

With that being said it can be used by anyone if desired. I also bumped `zwave-js-server` to 1.16.0 (which is needed for the new server option) and `zwave-js` to 9.0.1

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -405,6 +405,7 @@ export type ZwaveConfig = {
 	enableSoftReset?: boolean
 	deviceConfigPriorityDir?: string
 	serverPort?: number
+	serverServiceDiscoveryDisabled?: boolean
 	logEnabled?: boolean
 	logLevel?: LogManager.LogLevel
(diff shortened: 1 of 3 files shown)


robertsLando:
Hi @raman325 , thanks for the PR, I was bumping zjs and zjs server locally so I have also added this changes to that commit (there were some conflicts after the bump if I would have merged this). I have also added an hidden input so the setting doesn't disappear after a save. I wonder if it makes sense to create a visible input for this or not? Just in case any user could need to edit it without pointing him on manually edit settings json file

robertsLando:
Thanks for the tip, wil add that on the hint :)

PR #2378, opened 2022-04-18
fix(ui): correct misspelling

XtremeOwnageDotCom:
Corrected misspelling of "Unhauthenticated" to "Unauthenticated"

### DIFF
--- a/src/components/SmartStart.vue
+++ b/src/components/SmartStart.vue
@@ -267,7 +267,7 @@ export default {
 						},
 						{
 							type: 'checkbox',
-							label: 'S2 Unhauthenticated',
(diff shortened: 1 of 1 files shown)


src/components/SmartStart.vue
```diff
@@ -267,7 +267,7 @@ export default {
 						},
 						{
 							type: 'checkbox',
-							label: 'S2 Unhauthenticated',
+							label: 'S2 UnAuthenticated',
```
robertsLando, line 270:
```suggestion
							label: 'S2 Unauthenticated',
```

robertsLando (changes requested):

robertsLando (approved):

PR #2386, opened 2022-04-22
fix: correctly parse MultilevelSwitchCC notifications

### LINKED ISSUE #2382: [Bug]: node_notification event parameter 2 is always null
### Checklist

### Deploy method

Docker

### Zwavejs2Mqtt version

6.7.2

### ZwaveJS version

9.0.2

### Describe the bug

I'm trying to handle MultilevelSwitchChange notifications from ZwaveJS by subscribing to the `node_notification` MQTT topic.

I receive an object like this (pruned for clarity):

```
[
  { /* tons of information about the node */ },
  {
    id: "57-38-0-5"
    nodeId: 57
    commandClass: 38
    commandClassName: "Multilevel Switch"
    property: 5
    propertyName: 5
   },
  null
]
```

That `null` is the problem.  It should be an object containing an object with eventType and direction fields as can be seen in the last parameter of the [call to `emit()` in ZwaveJS](https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/node/Node.ts#L2894).

Without having read too deeply into ZwaveJS2MQTT, I suspect this parameter is not being serialized properly when the MQTT message is generated, resulting in a null instead of an object.

### To Reproduce

Interact with a Zwave device that generates mutlilevel switch change messages.  In my case, press the up or down arrow buttons on a VRCS4 or VRCZ4.

### Expected behavior

The second parameter should contain an object with the event data that was supplied in the call to `emit()`, not null.

### Additional context

Here's the pull request that added this feature to ZwaveJS in case it's of use to you: https://github.com/zwave-js/node-zwave-js/pull/4282

robertsLando:
Fixes #2382

cc @AlCalzone

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -57,9 +57,6 @@ import {
 	ZWaveNodeValueRemovedArgs,
 	ZWaveNodeValueUpdatedArgs,
 	ZWaveNotificationCallback,
-	ZWaveNotificationCallbackArgs_EntryControlCC,
-	ZWaveNotificationCallbackArgs_MultilevelSwitchCC,
-	ZWaveNotificationCallbackArgs_NotificationCC,
 	ZWaveOptions,
 	ZWavePlusNodeType,
 	ZWavePlusRoleType,
@@ -3357,11 +3354,9 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 	 * Emitted when we receive a node `notification` event
 	 *
 	 */
-	private _onNodeNotification: ZWaveNotificationCallback = (
-		zwaveNode,
-		ccId,
-		args
-	) => {
+	private _onNodeNotification: ZWaveNotificationCallback = (...parms) => {
+		const [zwaveNode, ccId, args] = parms
+
 		const valueId: Partial<Z2MValueId> = {
 			id: null,
 			nodeId: zwaveNode.id,
@@ -3373,13 +3368,11 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		let data = null
 
 		if (ccId === CommandClasses.Notification) {
-			args = <ZWaveNotificationCallbackArgs_NotificationCC>args
 			valueId.property = args.label
 			valueId.propertyKey = args.eventLabel
 
 			data = this._parseNotification(args.parameters)
 		} else if (ccId === CommandClasses['Entry Control']) {
-			args = <ZWaveNotificationCallbackArgs_EntryControlCC>args
 			valueId.property = args.eventType.toString()
 			valueId.propertyKey = args.dataType
 			data =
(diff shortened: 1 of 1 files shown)


lib/ZwaveClient.ts
```diff
@@ -3372,16 +3372,16 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
-		) {
+		} else if (ccId === CommandClasses['Entry Control']) {
 			valueId.property = args.eventType as string
 			valueId.propertyKey = args.dataType as string
 			data =
 				args.eventData instanceof Buffer
 					? utils.buffer2hex(args.eventData)
 					: args.eventData
+		} else if (ccId === CommandClasses['Multilevel Switch']) {
+			valueId.property = args.eventType as string
```
AlCalzone, line 3383: `args.eventType`  is not a string, writing `as string` behind that doesn't change it.

It is one of these two enum values:
https://github.com/zwave-js/node-zwave-js/blob/3ef1368bd1f2fd1e41ceac43bc9f89e0d1ffe550/packages/zwave-js/src/lib/commandclass/MultilevelSwitchCC.ts#L50-L51

This way you'll end up with a property of `4`  or `5`  (number, not string!)

AlCalzone (changes requested):

lib/ZwaveClient.ts
AlCalzone: You might want to use `getEnumMemberName` here

robertsLando: done

lib/ZwaveClient.ts
```diff
@@ -3355,7 +3359,10 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 	private _onNodeNotification(
 		zwaveNode: ZWaveNode,
 		ccId: CommandClasses,
-		args: Record<string, unknown>
+		args:
+			| ZWaveNotificationCallbackArgs_NotificationCC
+			| ZWaveNotificationCallbackArgs_EntryControlCC
+			| ZWaveNotificationCallbackArgs_MultilevelSwitchCC
 	) {
```
AlCalzone, line 3366: I think you'd better use the official callback type for this. That way the dependencies between the `ccId`  and the `args`  parameters are preserved:
https://github.com/zwave-js/node-zwave-js/blob/3b5c91a19e09132900dfd7000fd53f8409f2cb0f/packages/zwave-js/src/lib/node/_Types.ts#L87-L93

```ts
private _onNodeNotification: ZWaveNotificationCallback = (zwaveNode, ccId, args) => {
  // ...
}
```

Leaves less room for type errors on your side, or forgetting to update the signature when the types change.

AlCalzone (changes requested):

lib/ZwaveClient.ts
robertsLando: are you sure the `this` is preserved correctly inside the function declared in that way?

robertsLando: done

lib/ZwaveClient.ts
```diff
@@ -3352,11 +3354,9 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 	 * Emitted when we receive a node `notification` event
 	 *
 	 */
-	private _onNodeNotification(
-		zwaveNode: ZWaveNode,
-		ccId: CommandClasses,
-		args: Record<string, unknown>
-	) {
+	private _onNodeNotification: ZWaveNotificationCallback = (...parms) => {
+		const [zwaveNode, ccId, args] = parms
```
AlCalzone, line 3358: This seems a bit weird, but destructuring here instead of having 3 separate arguments makes it so TS understands the connection between the parameter types and can deduce the type of `args` in the if branches.

robertsLando: 👍🏼  Thanks :)

PR #2418, opened 2022-05-09
fix: avoid showing git error on startup

markis:
Even though `execSync` is wrapped in a try...catch it will still output error messages in docker console.

This will introduce a check for the git command in the shell, if it doesn't exist, then `exit 0` else run the git command. This avoids throwing and catching errors and keeps the logs clean.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -140,13 +140,13 @@ export function num2hex(num: number): string {
  */
 export function getVersion(): string {
 	if (!VERSION) {
-		let revision = ''
 		try {
-			revision = execSync('git rev-parse --short HEAD').toString().trim()
-		} catch (error) {
-			// git not installed
+			const shellCmd = 'command -v git || exit 0; git rev-parse --short HEAD'
+			const revision = execSync(shellCmd).toString().trim()
+			VERSION = `${version}${revision ? '.' + revision : ''}`
+		} catch {
+			VERSION = version
 		}
-		VERSION = `${version}${revision ? '.' + revision : ''}`
 	}
 
 	return VERSION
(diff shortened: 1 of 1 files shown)


lib/utils.ts
```diff
@@ -140,12 +140,7 @@ export function num2hex(num: number): string {
  */
 export function getVersion(): string {
 	if (!VERSION) {
-		let revision = ''
-		try {
-			revision = execSync('git rev-parse --short HEAD').toString().trim()
-		} catch (error) {
-			// git not installed
-		}
+		const revision = execSync('command -v git || exit 0; git rev-parse --short HEAD').toString().trim()
```
robertsLando, line 143: Change looks good, BTW I would wrap it in a try catch anyway, of an error happens here for any reason the app wouldn't start at all

robertsLando (changes requested):

lib/utils.ts
```diff
@@ -140,13 +140,13 @@ export function num2hex(num: number): string {
  */
 export function getVersion(): string {
 	if (!VERSION) {
-		let revision = ''
 		try {
-			revision = execSync('git rev-parse --short HEAD').toString().trim()
-		} catch (error) {
-			// git not installed
+			const shellCmd = 'command -v git || exit 0; git rev-parse --short HEAD'
+			const revision = execSync(shellCmd).toString().trim()
```
markis, line 145: Having the shell command and this code on the same line, seemed to make it too long/complex.  So, I split them up, if that's an issue, I can put them back together.

robertsLando (dismissed):

robertsLando:
@zwave-js-bot fix lint

PR #2412, opened 2022-05-06
fix: correctly parse buffers in MQTT payload

robertsLando:
Fixes #2303

PR #2420, opened 2022-05-10
docs: describe format for Buffers (#2412)

j9brown:


robertsLando (changes requested):
@j9brown Thanks for your PR! I would add this to a dedicated section as this is not only working for sendcommand api but for all mqtt payloads sent.

`## Buffer payload` could be good IMO

j9brown:
Sounds good, how about this?

robertsLando (approved):
LGTM. Thanks @j9brown ! :)

PR #2439, opened 2022-05-25
feat(ui): node events logs queue

robertsLando:
This can be useful expecially to detect node notifications and all other events that are not shown on UI

https://user-images.githubusercontent.com/11502495/170527755-33a39042-23d5-4aec-a324-bb068d54dd64.mp4

### DIFF
--- a/lib/SocketEvents.ts
+++ b/lib/SocketEvents.ts
@@ -16,6 +16,7 @@ export enum socketEvents {
 	grantSecurityClasses = 'GRANT_SECURITY_CLASSES',
 	validateDSK = 'VALIDATE_DSK',
 	inclusionAborted = 'INCLUSION_ABORTED',
+	nodeEvent = 'NODE_EVENT',
 }
 
 // events from client ---> server
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -78,6 +78,7 @@ import { GatewayValue } from './Gateway'
 
 import { ConfigManager, DeviceConfig } from '@zwave-js/config'
 import { socketEvents } from './SocketEvents'
+import { ZWaveNodeEventCallbacks } from 'zwave-js/build/lib/node/_Types'
 
 export const deviceConfigPriorityDir = storeDir + '/config'
 
@@ -163,6 +164,10 @@ const allowedApis = validateMethods([
 	'checkLifelineHealth',
 	'checkRouteHealth',
 ] as const)
+
+export type ZwaveNodeEvents =
+	| keyof ZWaveNodeEventCallbacks
+	| 'statistics updated'
 
 // Define mapping of CCs and node values to node properties:
 const nodePropsMap = {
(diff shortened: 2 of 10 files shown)


robertsLando:
cc @AlCalzone @kpine @jmgiaever

AlCalzone:
First draft looks good. A few things I'd change:

- [ ] Filter out `statistics updated` events. They get NOISY and are exposed through the statistics anyways.
- [ ] Filter out `metadata updated` events. Most users won't understand what to do with them, and they are indirectly exposed through the values anyways
- [ ] Put multiline arguments below the event title. Currently the first line is behind the title, making it very hard to grasp
- [ ] Indent the stuff that belongs to each event slightly, so it is easier to grasp visually
- [ ] Distuinguish separate events visually: A bit of padding, alternating BG colors, horizontal lines (?)

blhoward2:
Stylistically, there should only be a : if there is more following. The READY line in your example shows a : without anything else, which could be confusing.

kpine:
The hover styling is a bit too strong and distracts from viewing the data. I'd rather see something like a border around the contents, or just nothing at all. Perhaps a much lighter color if you prefer this still.

The event data looks better with a smaller line height.

The 'alive' event (and 'sleep', going by your screenshot) is showing an argument of `0`, which doesn't seem right. In comparison, the 'ready' event shows nothing. I'd expect the former to be the same as 'ready', since they have the same structure. Or are the docs incorrect?

[dead/alive](https://zwave-js.github.io/node-zwave-js/#/api/node?id=quotdeadquot-quotalivequot): `(node: ZWaveNode) => void`
[ready](https://zwave-js.github.io/node-zwave-js/#/api/node?id=quotreadyquot): `(node: ZWaveNode) => void`

I'm wondering if the max queue size of 100 is too large as a default value.

kpine:
> I noticed this but I dunno why it happens

Oh, I guess the docs are out of date, but the answer is already here. 😅 It's the old status.

https://github.com/zwave-js/zwavejs2mqtt/blob/4982a2fe1fc58e4519e65d33bb8ccc9093016a94/lib/ZwaveClient.ts#L3202

> Can be configured in settings but I could limit it more, maybe 50 ?

Just thinking of someone with a lot of nodes and events. What impact would saving N nodes * M events in cache be (50 nodes, 5000 max events; 100 nodes, 10,000 max events)? Might be nothing to worry about. Also how does the UI function with 100 events shown? Easy and fast to use? The default should be the "safest" option. I don't have the answer, just posing the question. 😃

robertsLando:
>  Also how does the UI function with 100 events shown? Easy and fast to use

I should try but I don't think it would be a problem

robertsLando:
Just tried with 200 items the UI has zero problems to show them, I think the same for 500 items but 100 I think it's enought

AlCalzone:
I'd do a few stylistic changes, here's an example for dark mode:
grafik

Here are the corresponding CSS changes I made for reference. BG color might need to be tuned for light mode:
grafik

PR #2443, opened 2022-05-27
fix(ui): show warning when zwave or general logs are disabled

### LINKED ISSUE #2442: Display message on debug page if logging is disabled
Sometimes people get confused because the debug logging screen only works if logging is enabled and log to file is disabled. If they are in any other configuration (disabled or log to file) we should display a message on the debug screen explaining that it won't work and how to enable it. 

Note that we should still allow zjs2mqtt logging to be disabled but zwavejs logging to be on.

Addon: Is there any reason why we can't allow the log level and such to also be changed on the debug logging screen? Alternatively, that screen become "Logging" and you could make it a one stop for the log settings and the debug logger and remove them from settings.

@robertsLando

robertsLando:
Fixes #2442

### DIFF
--- a/src/components/Debug.vue
+++ b/src/components/Debug.vue
@@ -21,9 +21,26 @@
 				>
 			</v-col>
 
-			<v-alert v-if="logDisabled" dense text type="warning">
-				Logging is disabled. Please enable it on settings in order to
-				see logs here
+			<v-alert
+				v-if="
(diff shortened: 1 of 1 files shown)


robertsLando:
cc @blhoward2

blhoward2 (approved):

blhoward2:
I haven't built it to test, but it looks good to me!

PR #2457, opened 2022-06-08
Add git to docker container

disconn3ct:
Fixes (non-fatal) startup error `/bin/sh: git: not found`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -56,7 +56,8 @@ RUN apk add --no-cache \
     libgcc \	
     libusb \	
     tzdata \	
-    eudev	
+    eudev \
+    git
 
 
 # Copy files from previous build stage	


robertsLando:
@chrisns I don't remember now, there was a reason why we didn't added it?

chrisns:
When does this error show?
Would like to resist bloating the container image
Need to know what leads up to getting the error for us to then decide if adding a dependency is the right answer

robertsLando:
@chrisns It happens because I'm actually using git to get the commit sha to append it to package.json version

chrisns:
ah i see, i'd probably suggest `cat .git/HEAD` then

robertsLando:
@chrisns I don't think `.git` folder will be there

chrisns:
then how is `git` working? :)

robertsLando:
@disconn3ct Check https://github.com/zwave-js/zwavejs2mqtt/pull/2458

PR #2458, opened 2022-06-09
fix: show commit sha whitout using git

robertsLando:
cc @chrisns

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,7 @@
 .git
+!/.git/HEAD
+!/.git/master
+
 .github
 *.md
 docs
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -5,6 +5,7 @@ import { ValueID } from 'zwave-js'
 import path from 'path'
 import crypto from 'crypto'
 import { execSync } from 'child_process'
+import { readFileSync } from 'fs'
 
 let VERSION: string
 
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -1,4 +1,7 @@
 .git
+!/.git/HEAD
+!/.git/master
```
chrisns, line 3: should be 
```suggestion
!/.git/refs/heads
```

chrisns: i think

chrisns (changes requested):

.dockerignore
robertsLando: Yeah I was testing and I just noticed it, also added the test branch that is usually used in tests releases

.dockerignore
```diff
@@ -1,4 +1,8 @@
 .git
+!/.git/HEAD
+!/.git/refs/heads/master
+!/.git/refs/heads/test
```
chrisns, line 4:
```suggestion
!/.git/refs/heads
```
no harm in having all refs, which will do other branches e.g. on prs

chrisns (changes requested):

.dockerignore
```diff
@@ -1,4 +1,7 @@
 .git
+!/.git/HEAD
+!/.git/refs/heads/*
+
```
chrisns, line 4:
```suggestion
!/.git/refs/heads/*
```
don't need the empty line, but whatevs :)

chrisns (approved):

PR #2469, opened 2022-06-17
feat: backup restore

robertsLando:
Ref #2068

https://user-images.githubusercontent.com/11502495/175021176-937c8e6f-a986-4cae-a795-8df8f395cb01.mp4

https://user-images.githubusercontent.com/11502495/175021182-ad62224f-4fdc-4096-89a0-3dea3b4677a3.mp4


robertsLando:
TODOs:

- [ ] Scheduled backup settings
- [ ] NVM backups
- [ ] Fix backups not saved on backups dir
- [ ] Backups dir env var

AlCalzone:
Couple of thoughts:

- [ ] Show "Please wait, backing up NVM..." instead of "Inclusion starting..." or "Exclusion starting..." while the NVM backup is being made. Ideally show a progress bar in the dialog.
- [ ] Give the NVM backup (on event) a name that helps remember what was done, e.g. `NVM_12345_before_exclusion` or  `NVM_12345_before_replace_node_2`
- [ ] Add a backup/restore everything button that does both the store and NVM backup/restore

robertsLando:
@AlCalzone Should I wait for a new zwavejs release befgore merging this?

mike-csis:
Awesome. From the preview:

- settings. Regular backups are enable/disable and have cron expressions. Makes sense. 
- nvm backup has both enable disable, cron and then additionally “on event” - will I disable on event, if I disable backups, do i disable them in events as well?
- It says nvm backups are before events. Will it make sense to do them after as well?
  - the nvm backup takes seconds last I did it, does this change over time - will we encounter minutes waiting while clicking include?
- are all backups in a backup folder I can point my backup tool at?
- How is the restore process - simple unpack zip and point at nvm?

LordMike:
Awesome. From the preview:

* Settings. Regular backups are enable/disable and have cron expressions. Makes sense.
* nvm backup has both enable disable, cron and then additionally “on event” - will I disable on event, if I disable backups, do i disable them in events as well?
* It says nvm backups are before events. Will it make sense to do them after as well?
  * the nvm backup takes seconds last I did it, does this change over time - will we encounter minutes waiting while clicking include?
* are all backups in a backup folder I can point my backup tool at?
* How is the restore process - simple unpack zip and point at nvm?

robertsLando:
> If I disable backups, do i disable "on events" as well?

NVm scheduled backups and NVM backup on event are indipendent each others but yeah maybe add 'scheduled' on the switch could make this more obivious

> Will it make sense to do them after as well?

I think it could makes sense but IMO the best woould be to have also it enabled scheduled so at worst you loose the last operation if it fails before the scheduled backup, making one before and one after it's too much (takes some time)

> The nvm backup takes approx. 30 seconds on my end, does this change over time (more devices = more time?)

I have no clue, I don't think so as I think it backups all the memory but I let @AlCalzone answer to this.

> Maybe when some events occur, there could be a checkbox to "do backup before (event)" - its default state (checked/unchecked) will depend on this setting

Could make sense but IMO if someone has it enabled could also disable it if he has lot of nodes to add. Would just be an (useless) additional step in inclusion/esclusion process

> How is the restore process - simple unpack zip and point at nvm?

The store restore process simply asks you to upload a zip file and it unzips the things he has backuped in the same place them was. The nvm restore is done from the adbanced dialog in control panel

LordMike:
Nice!... Sorry I didn't see your reply, must've missed an email, but it was merely thoughts after all :)
I'm on the snap train now, so I'm not sure how to test out stuff anyways. 

Looking forward to getting some backups in place :)

rezzalex:
Hello,

thank you for this new functionality.

So NVM backup will be 1 more file inside the "store".

the video shows a manual store backup, from the UI, where you can download your backup file and do place it where you want. But where is going to be stored the Automatic scheduled "store Backup" ? 

Using the Zwavejs2mqtt container, should I Bind all the container "store" folders to a folder naturally accessible from my Host (for back-up purposes mainly) ?

robertsLando:
Both NVM and store backups are stored inide store/backups folder.

robertsLando:
@rezzalex I would map the entire store dir to a local folder, not only the backup one :) Like suggested in docs

rezzalex:
It sounds safe to do so, but I thought the documentation told me not to do so : 
https://zwave-js.github.io/zwavejs2mqtt/#/getting-started/docker
I used the docker-compose.yml, where there is no bind at all.
Now I need to "move" the "store" folder content to a local folder, without losing everything...

robertsLando:
https://zwave-js.github.io/zwavejs2mqtt/#/getting-started/docker?id=run-using-a-local-folder

rezzalex:
yes, but if I do so right now, I could lose the content of the current volume. I will try to find a solution.

What do you think about modifying the docker-compose file in the doc so that it include the bind ? that would prevent my case to happen again

src/components/Settings.vue
```diff
@@ -299,7 +299,7 @@
 						<v-row class="mb-5">
 							<v-col cols="12" sm="6">
 								<v-switch
-									hint="Enable/Disable backup before node add/remove/replace operations. DISCLAIMER: After an NVM backup controller will be soft-resetted, if you disabled soft-reset you should NOT use this feature"
+									hint="Enable/Disable backup before node add/remove/replace operations. DISCLAIMER: After an NVM backup controller will be soft-resetted, please make sure your stick can be soft-reset before enabling this feature"
```
marleyjaffe, line 306: Does this warning/disclaimer only apply to the `Backup on Event` functionality or is it applicable to all NVM backup actions?

marleyjaffe (commented):
I'm looking to start using this new functionality and looking for clarity. I looked in the documentation but this disclaimer isn't mentioned there.

LordMike:
It will probably be for all’s nvm backups and restores.

As an aside. I had a thought today. Maybe a scheduled nvm backup will hide issues that come from running for a long period of time. If I backup each week, then the driver will be reset every week, potentially hiding an issue that could arise from running - think a millisecond counter that overflows or a list of things that becomes too large.

Fun thought. :P
________________________________

marleyjaffe:
That's what I figured. This disclaimer should probably be moved to underneath the `Controller (NVM) Backup` header instead of on the `Backup on Event` toggle so it's clear its for all NVM backup actions.

AlCalzone:
Every NVM backup currently triggers a bug in the 700 series controller firmware. Silabs confirmed the issue, hope we'll see a fix soon. For now the soft-reset is a workaround.

robertsLando:
> That's what I figured. This disclaimer should probably be moved to underneath the Controller (NVM) Backup header instead of on the Backup on Event toggle so it's clear its for all NVM backup actions.

I'm going to fix this now, thanks

src/components/Settings.vue
robertsLando: All NVM backups

PR #2476, opened 2022-06-21
fix: log rotation

### LINKED ISSUE #2471: [Bug]: Log rotation not working
### Checklist

### Deploy method

Docker

### Zwavejs2Mqtt version

6.12

### ZwaveJS version

9.3.0

### Describe the bug

A couple of things here I notice:
1 - Expecting a 7-day (?) rotation but the logs are accumulating daily
2 - There's a gap or two in daily log name zips (and I haven't manually deleted any) e.g. June 1 and June 8 in attached file list.

The zwavejs log is rotating every 7 days so this is restricted to zwavejs2mqtt only.

Running from HA with Frenk add-on, if that matters

Gaps in daily log list
Configuration
.

### To Reproduce

Configure as per attached configure file.  Let it run > 1 week.

### Expected behavior

Logs rotate, directories don't grow with files unmanaged. One archive per day, no gaps.

### Additional context

_No response_

robertsLando:
Fixes #2471

### DIFF
--- a/lib/logger.ts
+++ b/lib/logger.ts
@@ -112,6 +112,7 @@ export function customTransports(config: LoggerConfig): winston.transport[] {
 		} else {
 			fileTransport = new DailyRotateFile({
 				filename: config.filePath,
+				auditFile: 'z2m-logs.audit.json',
 				datePattern: 'YYYY-MM-DD',
 				zippedArchive: true,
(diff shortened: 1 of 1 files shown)


robertsLando:
cc @AlCalzone

PR #2485, opened 2022-06-24
feat: zwave radio configurations

### LINKED ISSUE #2319: [feat] Zwave-radio configurations
https://zwave-js.github.io/node-zwave-js/#/api/controller?id=configuring-the-z-wave-radio

- RF Region
- Power Level

robertsLando:
Fixes #2319
Schermata da 2022-06-27 15-55-22


AlCalzone (changes requested):
Very nice to see this landing. We shouldn't just put the user in front of these options though without explanation. This is a very advanced feature that nobody should use without knowing what they are doing.

Some points we need to mention:
1. Increasing the TX power (normal power level) may be **illegal**, depending on where the user is located
2. Increasing the TX power will only make the nodes "hear" the controller better, but not vice versa
3. The default values for 700 series sticks (at least the ones I've seen) are:
    TX power: 0.0
    measured: +3.3

PR #2541, opened 2022-07-20
fix: issues with controller powerlevel settings

### LINKED ISSUE #2532: Issues with controller powerlevel settings
While doing a range test, I found some weird behavior for the powerlevel settings:

1. The values should be set together, not separate. I think this happens behind the scenes, but it should really not be necessary to send both changes separately:
    grafik

2. That's more for me: Z-Wave JS does an automatic refresh after setting the values. I think the API should be updated to reflect the verified values. Some sticks do not accept all settings and revert:
    powerlevel

3. Refreshing the values does not work. They get queried but never show up in the UI:
    powerlevel2

4. If you look closely at the log, refreshing the powerlevels also queries the region for some reason.

robertsLando:
Fixes #2532

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -1941,7 +1941,7 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 				measured0dBm
 			)
 
-			await this.updateControllerNodeProps()
+			await this.updateControllerNodeProps(null, ['powerlevel'])
 
 			return result
 		}
@@ -1952,7 +1952,7 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 	async setRFRegion(region: RFRegion): Promise<boolean> {
 		if (this.driverReady) {
 			const result = await this._driver.controller.setRFRegion(region)
-			await this.updateControllerNodeProps()
+			await this.updateControllerNodeProps(null, ['RFRegion'])
 			return result
 		}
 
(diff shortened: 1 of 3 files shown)


robertsLando:
cc @AlCalzone

AlCalzone (changes requested):
image
Reload button vanishes when the textboxes are empty.

---

Refreshing the powerlevel also refreshes the region (see log):
```
17:36:02.411 DRIVER » [REQ] [SerialAPISetup]
                        command: GetPowerlevel
                        payload: 0x08
17:36:02.415 SERIAL « [ACK]                                                                   (0x06)
17:36:02.417 SERIAL « 0x0106010b0863e971                                                   (8 bytes)
17:36:02.417 SERIAL » [ACK]                                                                   (0x06)
17:36:02.418 DRIVER « [RES] [SerialAPISetup]
                        command:               GetPowerlevel
                        normal powerlevel:     9.9 dBm
                        output power at 0 dBm: 23.3 dBm
17:36:02.426 SERIAL » 0x0104000b20d0                                                       (6 bytes)
17:36:02.426 DRIVER » [REQ] [SerialAPISetup]
                        command: GetRFRegion
                        payload: 0x20
17:36:02.433 SERIAL « [ACK]                                                                   (0x06)
17:36:02.435 SERIAL « 0x0105010b2000d0                                                     (7 bytes)
17:36:02.435 SERIAL » [ACK]                                                                   (0x06)
17:36:02.436 DRIVER « [RES] [SerialAPISetup]
                        command: GetRFRegion
                        region:  Europe
```

robertsLando:
> Refreshing the powerlevel also refreshes the region (see log):

I know that, it uses a general method that updates all those properties, is that a problem?

AlCalzone:
Seems unnecessary since both have the refresh button and I would expect that I have to refresh both separately, but it shouldn't matter.

PR #2542, opened 2022-07-20
feat: zwave-js v10 support

robertsLando:


lib/ZwaveClient.ts
```diff
@@ -65,6 +65,9 @@ import {
 	SerialAPISetupCommand,
 	FirmwareUpdateFileInfo,
 	FirmwareUpdateInfo,
```
AlCalzone, line 67: You got a linter warning here

AlCalzone: 'FirmwareUpdateInfo' is defined but never used

AlCalzone (changes requested):

lib/ZwaveClient.ts
robertsLando: fixed

robertsLando:
@AlCalzone I have also introduced `node found` event, when you add a node it should be immediatly shown in control panel as soon as it is found

AlCalzone:
That seems to work, but is hard to test

AlCalzone (dismissed):

PR #2543, opened 2022-07-22
docs: adjust period, to prevent hitting ratelimit

kenthua:
- The liveness check is too often, that it hits the rate limit threshold and prevents access to the UI
- Adjusted the health check path


kubernetes/deployment.yaml
```diff
@@ -21,10 +21,10 @@ spec:
             httpHeaders:
             - name: Accept
               value: text/plain
-            path: /health
+            path: /health/zwave
```
robertsLando, line 24: `/health` api is more generic, it also checks for mqtt status (if it's not disabled)

robertsLando (changes requested):

kenthua:
updated, thanks @robertsLando

robertsLando (approved):

PR #2562, opened 2022-08-01
feat: allow to force disable SSL

### LINKED ISSUE #2480: [feat]: Allow to force disable SSL
### Checklist

### Deploy method

Docker

### Zwavejs2Mqtt version

6.12

### ZwaveJS version

9.3.0

### Describe the bug

When trying to scan QR code for DSK on Android and on Windows 11 (Chrome 102.0.5005.115) it fails with a "Socket Disconnected" error on Android and nothing on Windows. I also tried adding the full code typing it in with the same result.

Home Assistant OS
core-2022.6.7
ZwaveJS2MQTT installed using addon/integration method

### To Reproduce

ZwaveJS2MQTT >> Smart Start >> + >> SCAN
Scan QR and window closes immediately (no legible error on windows, but on Android is says "Socket Disconnected")
I have been able to provision entities before but had deleted the couple attempts

The same happens when you type everything instead of using camera/webcam to scan QR

### Expected behavior

Provisioned Entity gets added to list.

### Additional context

_No response_

robertsLando:
Fixes #2480

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -122,6 +122,10 @@ const apisLimiter = rateLimit({
 	},
 })
 
+function sslDisabled() {
+	return process.env.FORCE_DISABLE_SSL === 'true'
+}
+
 // apis response codes
 enum RESPONSE_CODES {
 	OK = 'OK',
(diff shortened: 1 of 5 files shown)


robertsLando:
cc @kpine

robertsLando:
@frenck based on what @kpine told me this env var could make sense to be set to true by default in the addon to prevent users enabling https

PR #2584, opened 2022-08-10
fix(ui): firmware update changelog overflow

### LINKED ISSUE #2583: [feat] suggestions for OTA UI improvements
The OTA UI has a couple issues that could use improvement, IMO.

First, it's more of an issue, but I'm lumping it together. The text in the version box overflows.

image

The main issue, is that the UI makes it too easy to accidentally perform a downgrade, if you aren't careful.

image

My switch is at v5.24, but z2m is prominently showing a version 5.22 that is available. I am making an assumption that z2m will show all versions, except for the current one (or maybe the API handles that, whatever...). I think there should be some sort of visual indication of older versions, versus new upgrades, as upgrading will be the more common action. Since the current version was far off on the right, I nearly tried to perform a downgrade.

I'm not proposing any particular UI design, just suggesting that there could be some improvement to distinguish older and newer upgrades, to support the most common use case of upgrading to new versions, not downgrading.

N/A

**Additional context**

N/A

robertsLando:
Fixes #2583

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -1887,26 +1887,28 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 					nodeId
 				)
 
-			// result = [
-			// 	{
-			// 		version: '1.13',
-			// 		changelog: '* Fixed some bugs\n* Added other bugs',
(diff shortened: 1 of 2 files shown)


robertsLando:
cc @kpine

PR #2610, opened 2022-08-29
feat: move logs to `logs` folder in store

### LINKED ISSUE #2600: [feat] Move logs to an subfolder of the store
The store folder is starting to pile up and quite large, with a mix of files. 

A log folder where logs are stored. Would be great if the corresponding `<hex>.audit.json` files could be moved out from the root of the store folder too.

This could possibly be set by the user itself with an env-var, if such a feature might break compability.

robertsLando:
Fixes #2600

cc @jmgiaever @AlCalzone


AlCalzone:
Don't forget to update the documentation.

PR #2627, opened 2022-09-06
fix(ui): get/set configuration CC

robertsLando:


src/components/nodes-table/NodeDetails.vue
```diff
@@ -281,7 +281,10 @@
 															commandClass: 112,
 														},
 														'get',
-														[configCC.parameter],
+														{
+															parameter:
+																configCC.parameter,
+														},
```
AlCalzone, line 287: the get method did not change in v10

AlCalzone (changes requested):

AlCalzone (approved):

PR #2644, opened 2022-09-12
feat: set zwave-js `userAgent` option

### LINKED ISSUE #2636: [feat] Set user-agent in driver
The driver recently added support for configuring the user agent that is used to make requests to our web services.
Z2m should register itself there, either using the https://zwave-js.github.io/node-zwave-js/#/api/driver?id=updateuseragent method (if it is NOT creating the driver instance itself) or using the `userAgent` driver option: https://zwave-js.github.io/node-zwave-js/#/api/driver?id=zwaveoptions

robertsLando:
Fixes #2636


AlCalzone (approved):
z2m always creates the driver itself? Then 👍️

robertsLando:
> z2m always creates the driver itself? Then +1

Sure!

PR #2705, opened 2022-10-04
chore(docker): bump node+alpine versions

chrisns:


robertsLando (approved):

robertsLando:
Is this good to merge? I saw that comment but I don't remember why I put it 😆

chrisns:
heh, maybe it didn't work with a version?
I built and tested and it worked ¯\_(ツ)_/¯

robertsLando:
I think it was an error with armv7 builds but AFAIK that should be fixed with serialport 10. Something about `musl`

robertsLando (approved):

PR #2706, opened 2022-10-04
chore(deps): enable renovate for dependency management

chrisns:
see https://github.com/chrisns/zwave-js-ui/pulls for its suggested updates
note it handles updating the docker images etc
can configure it to do loads (obvs)

not all the suggestions its making can be merged, but I'm happy to work through making current versions of our deps work as best i can


robertsLando:
It's ok for me, I would like to know if @AlCalzone agree as we temporary disabled dependabot but renovate should work well

AlCalzone:
Renovate Bot supports yarn, so this seems Ok.
Although dependabot support seems to be just around the corner now. And personally, I prefer the dependabot workflow, even though renovate is more configurable.

chrisns:
agree dependabot is nice since it is github, but sadly not as feature rich :(
renovate I'd say would still have a limited place for maintaining base docker images
so if/when dependabot supports yarn we might find ourselves wanting both but for different sets of deps perhaps?

chrisns:
What if I scope it down to just docker/* which dependabot doesn't do so well

robertsLando:
Ok for me

robertsLando:
@AlCalzone should we enable dependabot again now?

AlCalzone:
Dependabot seems to work with Yarn v3 now. I've gotten some security fixes already and other repos using yarn are completely managed by Dependabot again.

chrisns:
@robertsLando have limited this to just docker+nvm now, so dependabot can carry on doing the npm/yarn stuff

robertsLando (approved):

robertsLando:
Thanks @chrisns ! 🙏🏼

PR #2716, opened 2022-10-10
feat(ui): better snackbars and import provisioning json

### LINKED ISSUE #1942: [feat] Allow import of exported provisions info
Presently, you can export the provision information but there is no way to re-import that, such as if you were changing sticks. Suggest renaming Import (for photos) to upload, then making an import button for this.

robertsLando:
Fixes #1942

Schermata da 2022-10-10 18-14-34


robertsLando:
cc @blhoward2

PR #2719, opened 2022-10-11
feat(ui): switch from vuex to pinia

robertsLando:


robertsLando:
TODOs:

- [ ] Fix auth not persisted on page refresh
- [ ] Test all pages/dialogs

PR #2728, opened 2022-10-12
chore: pre-build server to speed up docker buildx

robertsLando:


robertsLando:
cc @chrisns this should speed up builds and prevent some strange issues with armv7 builds causing actions to freeze

PR #2751, opened 2022-10-23
chore(docker): add tags for `major` and `minor` version to container images

esolitos:
## What is the objective of this PR

On release this adds tags `${major}` as well as `${major}.${minor}` to the container images.

This is quite common practice and allows users to follow minor or major releases instead of only `latest` _(typically not recommended)_ or specific patch versions, requiring to constantly monitor for new patches releases.

Any feedback is welcome. :)

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -13,7 +13,7 @@ jobs:
     permissions:
       packages: write
       contents: read
-      
+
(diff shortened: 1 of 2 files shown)


robertsLando (approved):
Makes sense to me. Thanks for this! Would like to have @chrisns feedback too

chrisns (approved):
Lgtm
Thanks!

PR #2817, opened 2022-11-23
docs: correcting typo on URL and updating to newer dashboard yaml

tflatebo:
The URL only had one forward slash in it, and when I added this to the version of homeassistant I just installed, it had an additional "views" stanza in it so I included that full yaml.


robertsLando:
@zwave-js-bot fix lint

robertsLando (approved):

robertsLando:
Thanks @tflatebo

PR #2820, opened 2022-11-25
fix(mqtt): node added on fly not subscribed to changes

### LINKED ISSUE #2819: No ValueIds available trough mqtt since 7.0.0
### Checklist

### Deploy method

Manually built (git clone - yarn build - yarn install

### Z-Wave JS UI version

All since 7.0.0

### ZwaveJS version

Each test was done with the built in ZwaveJs version from ZwaveJSUI

### Describe the bug

I discoverer that when a node is included and inited. Any actions made trough mqtt results in a

No writeable valueId 

LOG  
0000|[90m2022-11-24 18:09:09.176[39m [32mINFO[39m [1mMQTT[22m: Message received on zwave/196/37/0/targetValue/set: 'true'
0001|[90m2022-11-24 18:09:09.176[39m [34mDEBUG[39m [1mGATEWAY[22m: No writeable valueId found for 196/37/0/targetValue

the only way to solve that is restart ZwaveJS UI after that no issue

or  go on zwavejs UI itself and trigger the action then the valueId becomes writable

### To Reproduce

Include a fresh node that has some actions (a switch is enough)

Once included try to control it trough mqtt using  id based path

Notice the issue

Exclude the device 

Use a version before 7.0.0

Do the same thing

The node is controlable directly after its inclusion and there is no valueId error

### Expected behavior

Control a device after inclusion with mqtt without having to restart zwavejs or to have to make a first action from zwavejs ui

### Additional context

Discussed on slack

robertsLando:
Fixes #2819

### DIFF
--- a/lib/Gateway.ts
+++ b/lib/Gateway.ts
@@ -1708,6 +1708,13 @@ export default class Gateway {
 				delete this.discovered[id]
 			}
 		}
+
+		// clean topicValues
+		for (const topic in this.topicValues) {
+			if (this.topicValues[topic].nodeId === node.id) {
+				delete this.topicValues[topic]
+			}
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


robertsLando:
cc @sarakha63 

I also added missing node props in events

PR #2852, opened 2022-12-21
docs: Fix broken links in Quick Start guide

uid1000:


robertsLando:
@uid1000 I think you are exploring docs from github .md files, them are compiled using docsify so the links are good, check: https://zwave-js.github.io/zwave-js-ui/#/getting-started/quick-start

PR #2861, opened 2022-12-27
feat: encrypt passwords in settings

### LINKED ISSUE #2857: Security : Cypher mqtt password
The configuration file settings.json contains mqtt password as uncypher texte. `"password":"mypasswordhere"`
With the store browsing it's really simple to get it.

Cypher the password in settings.json. Or perharps add login/password in users.json

**Additional context**
I'm using zwave-js-ui 8.6.2 with MQTT (mosquitto) dans domoticz

robertsLando:
Fixes #2857

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -49,6 +49,7 @@ import { inboundEvents, socketEvents } from './lib/SocketEvents'
 import * as utils from './lib/utils'
 import backupManager from './lib/BackupManager'
 import { readFile, realpath } from 'fs/promises'
+import PasswordManager from './lib/PasswordManager'
 
 declare module 'express-session' {
 	export interface SessionData {
@@ -168,6 +169,10 @@ let restarting = false
 
 // ### UTILS
 
+function updateSettings(settings: Settings) {
+	return jsonStore.put(store.settings, settings)
+}
+
 /**
  * Start http/https server and all the manager
  */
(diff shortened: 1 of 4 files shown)


chrisns (changes requested):
You appear to just be moving the problem around, better would be, if its defined in an env var and keep it out the `settings.json` entirely.
best would be if that env var could be a file path to where you mount a file with the 
so something like
```js
const mqtt_password = process.env.MQTT_PASSWORD_FILE ? fs.readFileSync(process.env.MQTT_PASSWORD_FILE) : process.env.MQTT_PASSWORD 
|| settingsjson.mqtt_password
```

more readable syntax is available :)

or if you just wanted a lighter weight, you could maybe have an env var that allowed you to exclude files in the store by a match, e.g. `**settings.json**` (or use regex)

that all said I'm not sure of the use case, the app needs the creds in the clear to talk to the mqtt service, someone accessing the gui has some presumably quite administrative access already since zwjs-ui doesn't have any form of RBAC.

you could probably subvert via the gui to subscribe to arbitrary things, and publish arbitrary things, so what is the threat model that you are concerned about here?

robertsLando:
> that all said I'm not sure of the use case, the app needs the creds in the clear to talk to the mqtt service, someone accessing the gui has some presumably quite administrative access already since zwjs-ui doesn't have any form of RBAC.

Yeah that's the same reason I was telling that guy this isn't really useful, also he complains about the file that stores the password in clear but also in that case protecting the device should be the solution...

chrisns:
so, close 'wont-fix'?

robertsLando:
> so, close 'wont-fix'?

👍🏼

PR #2886, opened 2023-01-16
feat: support otw update

robertsLando:


robertsLando:
cc @AlCalzone

AlCalzone (changes requested):
I'm missing a BIG FAT WARNING here:
grafik

If a user uploads the wrong file, the controller can be bricked. 700 series may be recoverable via the bootloader, see also https://zwave-js.github.io/node-zwave-js/#/api/driver?id=driver-events.
500 series is likely gone if the wrong file gets uploaded.

Also it would be good if the bootloader/fallback mode of the controller could be exposed - maybe in a separate PR though. See https://zwave-js.github.io/node-zwave-js/#/api/driver?id=zwaveoptions -> `allowBootloaderOnly` description.

AlCalzone:
Testing result:
I didn't see any progress and the update seems to have been done with an empty file.

PR #2888, opened 2023-01-17
chore: fix deprecation on GH actions

varet80:
Changes:

- Fix actions for set-output and save-state deprecations.
- update `actions/github-script` to new version from v3 to v6 (takes care of node v12 issue)
- update `actions/node-setup` from v3.4.1 to v3.6.0 (takes care of save-state issues)


robertsLando (approved):
LGTM

PR #2910, opened 2023-01-25
chore: prevent dups runs on releases

robertsLando:
This could fix the bug that for some reason stucks the release workflow and requires me to retrigger it


robertsLando:
Schermata da 2023-01-25 08-35-39

PR #2941, opened 2023-02-03
feat: zwave-js@10.7.0, heal network options, endpoints labels (#2941

robertsLando:
Removed `endpointIndezes` in favor of `endpoint` property from ZUINode

### DIFF
--- a/src/components/dialogs/DialogAssociation.vue
+++ b/src/components/dialogs/DialogAssociation.vue
@@ -188,9 +188,12 @@ export default {
 				endpoints.unshift({ text: 'No Endpoint', value: null })
 			}
 
-			if (node && node.endpointIndizes) {
-				for (const i of node.endpointIndizes) {
-					endpoints.push({ text: 'Endpoint ' + i, value: i })
+			if (node && node.endpoints) {
+				for (const endpoint of node.endpoints) {
+					endpoints.push({
+						text: endpoint.label || 'Endpoint ' + endpoint.index,
+						value: endpoint.index,
+					})
 				}
 			}
 
--- a/src/components/nodes-table/AssociationGroups.vue
+++ b/src/components/nodes-table/AssociationGroups.vue
@@ -138,10 +138,12 @@ export default {
 			const node = this.nodes[this.nodesMap.get(nodeId)]
 			if (node) {
 				const ep = node.endpoints.find((e) => e.index === endpoint)
-				return ep?.label ? ep.label : 'Endpoint ' + endpoint
-			} else {
-				return 'Endpoint ' + endpoint
+				if (ep) {
+					return ep.label
+				}
 			}
+
+			return 'Endpoint ' + endpoint
 		},
 		apiRequest(apiName, args) {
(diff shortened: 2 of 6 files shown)


src/views/ControlPanel.vue
```diff
@@ -413,6 +409,28 @@ export default {
...
+							'Healing network causes a lot of traffic, can take minutes up to hours and users have to expect degraded performance while it is going on',
+							'info',
+							{
+								confirmText: 'Heal',
+								inputs: [
+									{
+										type: 'checkbox',
+										label: 'Include sleeping nodes',
+										key: 'includeSleeping',
+										value: false,
```
AlCalzone, line 425: Default is `true` in the driver for compatibility reasons. I guess default `false` here is okay.

src/components/nodes-table/AssociationGroups.vue
```diff
@@ -128,6 +134,15 @@ export default {
 			const node = this.nodes[this.nodesMap.get(nodeId)]
 			return node ? node._name : 'NodeID_' + nodeId
 		},
+		getEndpointLabel(nodeId, endpoint) {
+			const node = this.nodes[this.nodesMap.get(nodeId)]
+			if (node) {
+				const ep = node.endpoints.find((e) => e.index === endpoint)
+				return ep?.label ? ep.label : 'Endpoint ' + endpoint
```
AlCalzone, line 141:
```suggestion
				return ep?.label || 'Endpoint ' + endpoint
```

AlCalzone (changes requested):

src/components/dialogs/DialogAssociation.vue
```diff
@@ -188,9 +188,12 @@ export default {
 				endpoints.unshift({ text: 'No Endpoint', value: null })
 			}
 
-			if (node && node.endpointIndizes) {
-				for (const i of node.endpointIndizes) {
-					endpoints.push({ text: 'Endpoint ' + i, value: i })
+			if (node && node.endpoints) {
+				for (const endpoint of node.endpoints) {
+					endpoints.push({
+						text: endpoint.label || 'Endpoint ' + endpoint.index,
```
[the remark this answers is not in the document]
AlCalzone: this is the actual change request, the other comments are just information/hints

src/components/nodes-table/AssociationGroups.vue
robertsLando: Yeah, didn't noticed this, even better:

```
if (node) {
				const ep = node.endpoints.find((e) => e.index === endpoint)
				if (ep) {
					return ep.label
				}
			}

			return 'Endpoint ' + endpoint
```

src/components/dialogs/DialogAssociation.vue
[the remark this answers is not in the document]
robertsLando: I think we could do this on backend side directly, I mean set the label to that on root endoint and if not set

AlCalzone (approved):

PR #2947, opened 2023-02-07
fix(ui): prismjs editor line numbers (#2947

### LINKED ISSUE #2945: File editor in store does not handle long lines correctly on narrow screens
They get wrapped, but the line counter increases:
grafik
(the highlighted text is actually part of line 47)

and at the end, some line numbers are missing due to this: 
grafik

When I make the screen wider, this does not happen:
grafik

robertsLando:
Fixes #2945

### DIFF
--- a/src/views/Store.vue
+++ b/src/views/Store.vue
@@ -234,6 +234,18 @@
 	outline: none !important;
 }
 
+.prism-editor-wrapper :deep(.prism-editor__editor) {
+	white-space: pre !important;
+}
+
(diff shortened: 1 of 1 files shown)


robertsLando:
Ref: https://github.com/koca/vue-prism-editor/issues/87

robertsLando:
@AlCalzone this disables soft wrap, it means you will have to scroll X but at least line numbers will be ok, don't have a better solution for now, what do you suggest? keep wrong line numbers or make them look ok but without wrap?

AlCalzone:
yeah I think this is better

PR #2963, opened 2023-02-14
fix: enable "real" client IP visibility even behind proxies

999LV:
Complements https://github.com/zwave-js/zwave-js-ui/commit/b1dc1a4f535fbba65020d8e45943bb1c8e7e4e33

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -512,6 +512,9 @@ function sortStore(store: StoreFileEntry[]) {
 logger.info(`Version: ${utils.getVersion()}`)
 logger.info('Application path:' + utils.getPath(true))
 
+// ensure the server sees the real client ip even behind a proxy (eg. for rate limiting or authentication checks)
+app.set('trust proxy', true)
+
 // view engine setup
 app.set('views', utils.joinPath(false, 'views'))
 app.set('view engine', 'ejs')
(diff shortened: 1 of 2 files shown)


robertsLando:
I think it would be better to enable this based on an env var. @chrisns agree?

chrisns:
I don't think we actually care in zwjs-ui what the ip is, apart from maybe logging?

agree, if we're going to define it must be an env var / config

setting trust to `true` is a really bad idea (though at the user's discretion if its configurable), it should really be more specific, and must **only** enabled in situations where a proxy is present, otherwise an actor with direct access could use it to masquerade as any ip they like.

I suspect most users don't have this behind a proxy (which is why auth exists within the app etc)

https://expressjs.com/en/guide/behind-proxies.html

chrisns:
moved to draft, if its changed to be configurable, somewhere then it can be ready to review again.

999LV:
Got your points, thank you. I'll resubmit based on an environment variable configuration (with default to 'false' so that it is really a user conscious decision)

999LV:
On thinking further about this « origin ip » topic, I think it is better to withdraw that pull request indeed since it does not bring much but creates other risks. Apologies.

PR #2968, opened 2023-02-21
feat: bump zwave-js-server@1.26.0

raman325:
Update zwave-js-server to version 1.26.0

@robertsLando can we release for this?


robertsLando (approved):
@raman325 Let me checck some minor fixes then I will do a new release :)

PR #2981, opened 2023-02-26
chore(deps): remove support for node 14

chrisns:
node 14 LTS security patches end in [2 months](https://endoflife.date/nodejs), so getting ahead of this and removing the references for node 14 and see what CI things it breaks


robertsLando (approved):
LGTM! Thanks 🙏🏼

PR #2985, opened 2023-02-27
fix: toSubscribe being set to bool instead of map

lightswitch05:
I setup [sonarcloud](https://sonarcloud.io/summary/overall?id=lightswitch05_zwave-js-ui) on my fork, and this was reported as a bug. I don't use the MQTT gateway on my local setup, so I'm not sure what type of testing is required, but it seems like a pretty obvious bug.


robertsLando (approved):
Strange nothing on my ide ever highlighted this, thanks!

PR #3018, opened 2023-03-24
feat: bump zwave-js-server@1.27.0

raman325:
as usual, would love a release for this so that we can include support for 10.11 and 10.12 features


robertsLando:
Sure thanks @raman325 :)

PR #3052, opened 2023-04-17
chore: add launch task to start webpack and z-wave with F5

AlCalzone:
With this PR, one can just press F5 to start both the `dev` script (if not running yet) and the `dev:server` script for debugging, instead of having to start both manually via the CLI.

grafik


robertsLando (approved):
I also added the HTTPS scipt, Thanks ❤️

PR #3065, opened 2023-04-26
feat(ui): support for Schedule Entry Lock CC

### LINKED ISSUE #2871: Expose Schedule Lock CC to UI
[ScheduleLock CC](https://zwave-js.github.io/node-zwave-js/#/api/CCs/ScheduleEntryLock) is supported but the only way to use it is by using APIs.

Types are: https://github.com/zwave-js/node-zwave-js/blob/75feccb134029908419fcb041ff29dc24395dc8a/packages/cc/src/lib/_Types.ts#L1168

Essentially there are 3 modes: Daily Weekly and Yearly

Each mode has a fixed number of slot.

Slots rappresents a specific schedule in a specific mode and can be assigned to a specific userId or to all users

```ts
export interface ScheduleEntryLockSlotId {
	userId: number;
	slotId: number;
}

export enum ScheduleEntryLockWeekday {
	// Yay, consistency!
	Sunday = 0x00,
	Monday = 0x01,
	Tuesday = 0x02,
	Wednesday = 0x03,
	Thursday = 0x04,
	Friday = 0x05,
	Saturday = 0x06,
}

export interface ScheduleEntryLockDailyRepeatingSchedule {
	weekdays: ScheduleEntryLockWeekday[];
	startHour: number;
	startMinute: number;
	durationHour: number;
	durationMinute: number;
}

export interface ScheduleEntryLockYearDaySchedule {
	startYear: number;
	startMonth: number;
	startDay: number;
	startHour: number;
	startMinute: number;
	stopYear: number;
	stopMonth: number;
	stopDay: number;
	stopHour: number;
	stopMinute: number;
}

export interface ScheduleEntryLockWeekDaySchedule {
	weekday: ScheduleEntryLockWeekday;
	startHour: number;
	startMinute: number;
	stopHour: number;
	stopMinute: number;
}
```

robertsLando:
Fixes #2871

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -80,6 +80,8 @@ import {
 	ScheduleEntryLockYearDaySchedule,
 	ScheduleEntryLockDailyRepeatingSchedule,
 	ScheduleEntryLockSlotId,
+	ScheduleEntryLockCC,
+	UserCodeCC,
 } from 'zwave-js'
 import { getEnumMemberName, parseQRCodeString } from 'zwave-js/Utils'
 import { nvmBackupsDir, storeDir, logsDir } from '../config/app'
@@ -195,6 +197,7 @@ export const allowedApis = validateMethods([
 	'syncNodeDateAndTime',
 	'manuallyIdleNotificationValue',
 	'getSchedules',
+	'setSchedule',
 ] as const)
 
 export type ZwaveNodeEvents = ZWaveNodeEvents | 'statistics updated'
@@ -992,14 +995,30 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 			)
 		}
 
-		const userCodes = await zwaveNode.commandClasses[
-			'User Code'
-		].getUsersCount()
+		const userCodes = UserCodeCC.getSupportedUsersCached(
+			// @ts-expect-error https://github.com/zwave-js/node-zwave-js/issues/5602
+			this.driver,
+			zwaveNode
+		)
 
-		const numSlots = await zwaveNode.commandClasses[
-			'Schedule Entry Lock'
-		].getNumSlots()
-
+		const numSlots = {
+			numWeekDaySlots: ScheduleEntryLockCC.getNumWeekDaySlotsCached(
+				// @ts-expect-error https://github.com/zwave-js/node-zwave-js/issues/5602
+				this.driver,
+				zwaveNode
+			),
+			numYearDaySlots: ScheduleEntryLockCC.getNumYearDaySlotsCached(
+				// @ts-expect-error https://github.com/zwave-js/node-zwave-js/issues/5602
+				this.driver,
+				zwaveNode
+			),
+			numDailyRepeatingSlots:
+				ScheduleEntryLockCC.getNumDailyRepeatingSlotsCached(
+					// @ts-expect-error https://github.com/zwave-js/node-zwave-js/issues/5602
+					this.driver,
+					zwaveNode
+				),
+		}
 		const weeklySchedules: (ScheduleEntryLockWeekDaySchedule &
 			ScheduleEntryLockSlotId)[] = []
 		const yearlySchedules: (ScheduleEntryLockYearDaySchedule &
@@ -1117,6 +1136,31 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 			)
 		} else {
 			throw new Error('Invalid schedule type')
+		}
+
+		// means that is not using supervision, read slot and check if it matches
+		if (!result) {
+			const methods = {
+				daily: 'getDailyRepeatingSchedule',
+				weekly: 'getWeekDaySchedule',
+				yearly: 'getYearDaySchedule',
+			}
+			const res = await zwaveNode.commandClasses['Schedule Entry Lock'][
+				methods[type]
+			](slot)
+
+			if (
+				(isDelete && !res) ||
+				(!isDelete && res && utils.deepEqual(res, schedule))
+			) {
+				result = {
+					status: SupervisionStatus.Success,
+				}
+			} else {
+				result = {
+					status: SupervisionStatus.Fail,
+				}
+			}
 		}
 
 		if (result.status === SupervisionStatus.Success) {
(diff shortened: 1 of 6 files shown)


lib/ZwaveClient.ts
```diff
@@ -961,6 +980,173 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+			)
+		}
+
+		const userCodes = await zwaveNode.commandClasses[
+			'User Code'
+		].getUsersCount()
+
+		const numSlots = await zwaveNode.commandClasses[
+			'Schedule Entry Lock'
+		].getNumSlots()
```
AlCalzone, line 1001: This will request the slots from the lock again. This information **should** be queried during the interview, so you can read the cached information using
```ts
ScheduleEntryLockCC.getNumWeekDaySlotsCached(driver, zwaveNode)
```
and similar for the other types.

https://github.com/zwave-js/node-zwave-js/blob/4854b0cf2b2c9bad15ac1b42a2632776e28db4b7/packages/cc/src/cc/ScheduleEntryLockCC.ts#L492-L543

lib/ZwaveClient.ts
```diff
@@ -961,6 +980,173 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+
+		if (!zwaveNode?.commandClasses['Schedule Entry Lock'].isSupported()) {
+			throw new Error(
+				'Schedule Entry Lock CC not supported on node ' + nodeId
+			)
+		}
+
+		const userCodes = await zwaveNode.commandClasses[
+			'User Code'
+		].getUsersCount()
```
AlCalzone, line 997: This will request the number of users from the lock again. This information **should** be queried during the interview, so you can read the cached information using
```ts
UserCodeCC.getSupportedUsersCached(driver, zwaveNode)
```

https://github.com/zwave-js/node-zwave-js/blob/4854b0cf2b2c9bad15ac1b42a2632776e28db4b7/packages/cc/src/cc/UserCodeCC.ts#L1007-L1015

lib/ZwaveClient.ts
```diff
@@ -961,6 +980,173 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+				'Schedule Entry Lock'
+			].setYearDaySchedule(
+				slot,
+				schedule as ScheduleEntryLockYearDaySchedule
+			)
+		} else {
+			throw new Error('Invalid schedule type')
+		}
+
+		if (result.status === SupervisionStatus.Success) {
```
AlCalzone, line 1122: This will throw if the command was sent without Supervision. The methods return `SupervisionResult | undefined`.

I have a couple of utility functions exported from `@zwave-js/core` that you can use to decide what to do based on the result:
https://github.com/zwave-js/node-zwave-js/blob/4854b0cf2b2c9bad15ac1b42a2632776e28db4b7/packages/core/src/consts/Transmission.ts#L275-L308

If the command **was** unsupervised, you may want to verify the change by reading the slot again, in case it was not accepted by the lock.

AlCalzone (changes requested):

lib/ZwaveClient.ts
robertsLando: Argument of type 'Driver' is not assignable to parameter of type 'ZWaveApplicationHost'.
  Type 'Driver' is missing the following properties from type 'ZWaveApplicationHost': options, controllerLog, getValueDB, tryGetValueDB, and 10 more

lib/ZwaveClient.ts
robertsLando: In poor words if result is undefined I should read the slot again

jmgiaever: If I don't remember correctly @AlCalzone, the DanaLock might (will?) return false even though codes are set.

Not that I expect a workaround/hack for it, but it might help you to know during development and debigging, that you should pull and check what is set or not and not wasting time on buggy software. 

And there should possible be a refresh button in the UI? 😅

robertsLando: In latest commits I added a check and if result is undefined I read the slot back and compare it with the actual set, if they match I return success

jmgiaever: But I actually think the DanaLock returns `false`

AlCalzone: you mean a Supervision status of Fail?

jmgiaever: I quote you're email: 

```
Setting a schedule always results in a supervision status of Fail (0x02). Here's an example for a daily repeating schedule (Mon, Wed, Fri, from 00:00 to 08:30) for user 17, slot 1:

2022-09-01T07:59:44.109Z DRIVER » [Node 002] [REQ] [SendData]

                                  │ transmit options: 0x25

                                  │ callback id:      114

                                  └─[Security2CCMessageEncapsulation]

                                    │ sequence number: 224

                                    └─[SupervisionCCGet]

                                      │ session id:      52

                                      │ request updates: true

                                      └─[ScheduleEntryLockCCDailyRepeatingScheduleSet]

                                          payload: 0x0111012a0000081e

...

2022-09-01T07:59:46.695Z DRIVER « [Node 002] [REQ] [ApplicationCommand]

                                  └─[Security2CCMessageEncapsulation]

                                    │ sequence number: 236

                                    └─[SupervisionCCReport]

                                        session id:          52

                                        more updates follow: true

                                        status:              Fail

                                        duration:            0s

but reading the schedule back afterwards shows that the schedule was indeed accepted by the lock:

2022-09-01T08:00:02.736Z DRIVER » [Node 002] [REQ] [SendData]

                                  │ transmit options: 0x25

                                  │ callback id:      117

                                  └─[Security2CCMessageEncapsulation]

                                    │ sequence number: 227

                                    └─[ScheduleEntryLockCCDailyRepeatingScheduleGet]

                                        payload: 0x1101

...

2022-09-01T08:00:02.952Z DRIVER « [Node 002] [REQ] [ApplicationCommand]

                                  └─[Security2CCMessageEncapsulation]

                                    │ sequence number: 239

                                    └─[ScheduleEntryLockCCDailyRepeatingScheduleReport]

                                        payload: 0x11012a0000081e

 

 

In that case, I would have expected a supervision status of Success (0xff) instead, to signal to the controlling node that the schedule was accepted.
```

Issue #3077, opened 2023-05-05
Yale YRD430-ZW2 spamming network


murkumurdok:
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

NO, my problem is NOT within Home Assistant or the ZWave JS integration

### Is your problem within Z-Wave JS UI (formerly ZwaveJS2MQTT)?

NO, my problem is NOT within Z-Wave JS UI

### Checklist

- [X] I have checked the [troubleshooting section](https://zwave-js.github.io/node-zwave-js/#/troubleshooting/index) and my problem is **not** described there.

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

What causes the bug?

Adding a newly purchased YRD430-ZW2 causes the entire network to slow down and spams traffic

What do you observe?

All devices, including lock operation are significantly slowed down, debug logs are filled with entries from YRD430-ZW2.  Excluding the lock returns the network to normal.  Including it again reintroduces the problem.  Reinteviewing and healing the node do not help.

What did you expect to happen?

I also have an older YRD256 on another door with presumably the same zwave chipset, it performs normally and does not spam the network

Including the YRD430-ZW2 creates the issues.

### Device information

Manufacturer: Yale
Model name: YRD430-ZW2
Node ID in your network: 99

### How are you using `node-zwave-js`?

- [X] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker image (latest)
- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker image (dev)
- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker manually built _(please specify branches)_
- [ ] `ioBroker.zwave2` adapter _(please specify version)_
- [X] `HomeAssistant zwave_js` integration  _(please specify version)_
- [ ] `pkg`
- [ ] `node-red-contrib-zwave-js` _(please specify version, double click node to find out)_
- [ ] Manually built from GitHub _(please specify branch)_
- [ ] Other _(please describe)_

### Which branches or versions?

version:
`node-zwave-js` branch:
`zwave-js-ui` branch:
Homeassistant addon Z-Wave JS UI.
Addon version 1.12.0.
zwavejsui version v8.15.0

### Did you change anything?

yes (please describe)

### If yes, what did you change?

Included YRD430-ZW2

### Did this work before?

[...]

PR #3078, opened 2023-05-05
fix: improve schedule lock entry cc query to prevent spam

robertsLando:
Fixes: #3077

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -1005,7 +1005,12 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 	/**
 	 * If the node supports Schedule Lock CC parses all available schedules and cache them
 	 */
-	async getSchedules(nodeId: number, mode?: ZUIScheduleEntryLockMode) {
+	async getSchedules(
+		nodeId: number,
+		opts: { mode?: ZUIScheduleEntryLockMode; fromCache: boolean } = {
+			fromCache: true,
+		}
+	) {
 		const zwaveNode = this.getNode(nodeId)
 
 		if (!zwaveNode?.commandClasses['Schedule Entry Lock'].isSupported()) {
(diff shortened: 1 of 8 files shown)


lib/ZwaveClient.ts
```diff
@@ -1020,48 +1022,18 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
-				const schedule = await zwaveNode.commandClasses[
-					'Schedule Entry Lock'
-				].getDailyRepeatingSchedule(slot)
-				if (schedule) dailySchedules.push({ ...slot, ...schedule })
-			}
-		}
+		const isInited = !!node.schedule
 
-		const node = this._nodes.get(nodeId)
 		node.schedule = {
```
AlCalzone, line 1037: I don't understand this tbh. You check first if node.schedule is defined, then you override it.

AlCalzone: Also, are the schedules cached locally or do you re-query every time this is called?

robertsLando: The diff doens't help understanding this change, you should see the end result. BTW I firstly check if schedule is defined because I call `getSchedule` on node ready, the first time is called so it just create the `node.schedule` object with the cached info, on next calls it also query the slots. So that check is used to allow init node.schedule and then skip the query that is done in the `for` loop.

robertsLando: > Also, are the schedules cached locally or do you re-query every time this is called?

Them are cached on node, not on disk BTW

robertsLando:
TODOs:

- [ ] Filter years from now
- [ ] Fix typo in duration
- [ ] Allow to cancel query
- [ ] Lock get schedule call
- [ ] Only refresh a single mode
- [ ] Do not show not supported modes
- [ ] Use v-timepicker for duration
- [ ] cc @AlCalzone

lib/ZwaveClient.ts
```diff
@@ -996,95 +1013,218 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+						status === UserIDStatus.Available ||
+						status === UserIDStatus.StatusNotAvailable
+					) {
+						// skip query on not enabled userIds or empty codes
+						continue
+					}
+
+					node.userCodes.available.push(i)
+
+					const enalbedUserid =
```
AlCalzone, line 1108:
```suggestion
					const enabledUserid =
```

AlCalzone (changes requested):
small typo. let's just get it out and get some feedback

robertsLando:
I also added methods to get slots from cache

PR #3131, opened 2023-06-07
fix(ui): filter battery null values

kvaster:
For fibaro thermostat we may have endpoint 1 and endpoint 2 for batteries. In such case battery level for endpoint 0 will be also shown. And it will be shown as 'null' value.

### DIFF
--- a/src/components/nodes-table/SmartView.vue
+++ b/src/components/nodes-table/SmartView.vue
@@ -301,7 +301,7 @@ import {
 } from '@mdi/js'
 import { mapState } from 'pinia'
 import useBaseStore from '../../stores/base.js'
-import { jsonToList } from '../../lib/utils.js'
+import { getBatteryDescription, jsonToList } from '../../lib/utils.js'
 
 export default {
 	props: {
@@ -505,10 +505,7 @@ export default {
 				icon = mdiPowerPlug
 				description = 'Main power source'
 			} else {
-				description = Array.isArray(node.batteryLevels)
-					? 'All battery levels: ' +
-					  node.batteryLevels.map((v) => `${v}%`).join(',')
-					: 'Unknown battery level'
+				description = getBatteryDescription(node)
 				if (level <= 10) {
 					icon = mdiBatteryAlertVariantOutline
 					iconStyle = `color: ${colors.red.base}`
--- a/src/components/nodes-table/nodes-table.js
+++ b/src/components/nodes-table/nodes-table.js
@@ -26,6 +26,7 @@ import {
 	mdiSleep,
 } from '@mdi/js'
 import useBaseStore from '../../stores/base.js'
+import { getBatteryDescription } from '../../lib/utils.js'
 
 export default {
 	props: {
@@ -273,10 +274,7 @@ export default {
 				description = 'mains-powered'
 			} else {
 				label = `${level}%`
-				description = Array.isArray(node.batteryLevels)
-					? 'All battery levels: ' +
-					  node.batteryLevels.map((v) => `${v}%`).join(',')
-					: 'Unknown battery level'
+				description = getBatteryDescription(node)
 				if (level <= 10) {
 					icon = mdiBatteryAlertVariantOutline
 					iconStyle = `color: ${colors.red.base}`
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -166,3 +166,12 @@ export function arraysEqual(a, b) {
 
 	return true
 }
+
+export function getBatteryDescription(node) {
+	return typeof node.batteryLevels !== 'undefined'
+		? 'All battery levels: ' +
+				Object.values(node.batteryLevels)
+					.map((v) => `${v}%`)
+					.join(',')
+		: 'Unknown battery level'
+}
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -172,6 +172,6 @@ export function getBatteryDescription(node) {
 		? 'All battery levels: ' +
 				Object.values(node.batteryLevels)
 					.map((v) => `${v}%`)
-					.join(',')
+					.join(', ')
 		: 'Unknown battery level'
 }
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -227,11 +227,11 @@ const observedCCProps: {
 } = {
 	[CommandClasses.Battery]: {
 		level(node, value) {
-			const levels: number[] = node.batteryLevels || []
+			const levels: { [key: number]: number } = node.batteryLevels || {}
 
 			levels[value.endpoint] = value.value
 			node.batteryLevels = levels
-			node.minBatteryLevel = Math.min(...levels)
+			node.minBatteryLevel = Math.min(...Object.values(levels))
 
 			this.emitNodeUpdate(node, {
 				batteryLevels: levels,
@@ -470,7 +470,7 @@ export type ZUINode = {
 	inited: boolean
 	healProgress?: HealNodeStatus | undefined
 	minBatteryLevel?: number
-	batteryLevels?: number[]
+	batteryLevels?: { [key: number]: number }
 	firmwareUpdate?: FirmwareUpdateProgress
 	firmwareCapabilities?: FirmwareUpdateCapabilities
 	eventsQueue: NodeEvent[]


robertsLando (commented):
Changes looks good to me, anyway I would like to know why this could happen. 

cc @AlCalzone

AlCalzone:
Can I has logs?

robertsLando:
@kvaster 
Please make a [driver log](https://zwave-js.github.io/zwave-js-ui/#/troubleshooting/generating-logs?id=driver-logs), loglevel `debug` and attach it here as a file (drag & drop into the text field).

kvaster:
Attached logs to related issue: https://github.com/zwave-js/node-zwave-js/pull/5899

Thanks.

kvaster:
Is anything else needed from me to help analyze problem ?

robertsLando:
I'm waiting for @AlCalzone to check your logs

AlCalzone:
PR will be merged soon

AlCalzone (changes requested):
The config file PR will be merged soon - meaning this PR should be unnecessary then.
Also, given the upcoming changes to values in `v11`, this change would actually be wrong, as `null` will be an allowed and expected value then.

kvaster:
These are different problems. Without fix in node-zwave-js repo I can see three batteries - 0 and 1 endpoints duplicated. With the fix there is only endpoint 1 and 2, but zwavejsui uses array for storing batteries values and it will have still three batteris with `null` value on the endpoint 0.

AlCalzone:
Ok, but we should still make sure that the filter is filtering the correct values. In `node-zwave-js` version 11, `null` will mean "known to be unknown" and `undefined` will mean "missing information" or "not yet known".
Checking `... != null` will exclude both, which might not be correct.

kvaster:
And what will be the best - convert values to hash where will be no `0` key or just filter undefined ? I'm not sure here...

kvaster:
@AlCalzone , probably updated implementation is better....

src/components/nodes-table/SmartView.vue
```diff
@@ -505,10 +505,16 @@ export default {
...
 				description = 'Main power source'
 			} else {
-				description = Array.isArray(node.batteryLevels)
-					? 'All battery levels: ' +
-					  node.batteryLevels.map((v) => `${v}%`).join(',')
-					: 'Unknown battery level'
+				description =
+					node.batteryLevels !== null &&
+					typeof node.batteryLevels === 'object' &&
+					!Array.isArray(node.batteryLevels)
```
robertsLando, line 511: This check is useless because they could never be an array now

src/components/nodes-table/SmartView.vue
```diff
@@ -505,10 +505,16 @@ export default {
...
+						? 'All battery levels: ' +
+						  Object.entries(node.batteryLevels)
```
robertsLando, line 513: You sort them now but no info about the endpoint?

src/components/nodes-table/nodes-table.js
```diff
@@ -273,10 +273,16 @@ export default {
 				description = 'mains-powered'
 			} else {
 				label = `${level}%`
-				description = Array.isArray(node.batteryLevels)
-					? 'All battery levels: ' +
-					  node.batteryLevels.map((v) => `${v}%`).join(',')
-					: 'Unknown battery level'
+				description =
```
robertsLando, line 276: Same as above, also giving that this is the same function consider moving it to `lib/utils.js` and create a function `getBatteryDescription(node)`

robertsLando (changes requested):

src/components/nodes-table/SmartView.vue
kvaster: I've changed array to object and key order in object is not defined or predictable - I'm sorting entries by key and combining values as it was previously with array - result should be the same.

src/components/nodes-table/SmartView.vue
kvaster: Makes sense. Will update PR in a moment.

robertsLando (dismissed):
LGTM! Thanks @kvaster , @AlCalzone good to merge?

AlCalzone (changes requested):

AlCalzone:
Also regarding the other comment about object property order:
Keys that look like they are numeric (i.e. numeric strings) are iterated in ascending order, no matter the insertion order:
https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order/38218582#38218582

src/lib/utils.js
```diff
@@ -166,3 +166,13 @@ export function arraysEqual(a, b) {
 
 	return true
 }
+
+export function getBatteryDescription(node) {
+	return node.batteryLevels !== null
```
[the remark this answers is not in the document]
kvaster: Telling the truth, I'm a little bit confused how js handles this. In my case value for endpoint 0 is never assigned and here goes test:

```
Welcome to Node.js v20.3.0.
Type ".help" for more information.
> obj = []
[]
> obj[1] = 99
99
> obj[2] = 99
99
> obj.map((v) => `${v}%`).join(',')
',99%,99%'
> obj
[ <1 empty item>, 99, 99 ]
> Object.entries(obj)
[ [ '1', 99 ], [ '2', 99 ] ]
> Object.values(obj)
[ 99, 99 ]
> Math.min(...obj)
NaN
> Math.min(...Object.values(obj))
99
> obj[0] = null
null
> obj.map((v) => `${v}%`).join(',')
'null%,99%,99%'
> obj
[ null, 99, 99 ]
> Object.values(obj)
[ null, 99, 99 ]
> Math.min(...obj)
0
> Math.min(...Object.values(obj))
0
```

Without changing batteryLevels to object instead of Array minBatteryLevel is shown as `null` in my case and battery levels are shown similar to: `null%,60%,90%`

I've tried following patch and in this case minBatterLevel is shown OK: `60%`, but battery levels are still shown as `null%,60%,90%`:

```
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -231,7 +231,7 @@ const observedCCProps: {
 
                        levels[value.endpoint] = value.value
                        node.batteryLevels = levels
-                       node.minBatteryLevel = Math.min(...levels)
+                       node.minBatteryLevel = Math.min(...Object.values(levels))
 
                        this.emitNodeUpdate(node, {
                                batteryLevels: levels,
diff --git a/src/components/nodes-table/SmartView.vue b/src/components/nodes-table/SmartView.vue
index 795a13d7..b10866ff 100644
--- a/src/components/nodes-table/SmartView.vue
+++ b/src/components/nodes-table/SmartView.vue
@@ -507,7 +507,7 @@ export default {
                        } else {
                                description = Array.isArray(node.batteryLevels)
                                        ? 'All battery levels: ' +
-                                         node.batteryLevels.map((v) => `${v}%`).join(',')
+                                         Object.values(node.batteryLevels).map((v) => `${v}%`).join(',')
                                        : 'Unknown battery level'
                                if (level <= 10) {
                                        icon = mdiBatteryAlertVariantOutline
diff --git a/src/components/nodes-table/nodes-table.js b/src/components/nodes-table/nodes-table.js
index 8360d8dc..035b339b 100644
--- a/src/components/nodes-table/nodes-table.js
+++ b/src/components/nodes-table/nodes-table.js
@@ -275,7 +275,7 @@ export default {
                                label = `${level}%`
                                description = Array.isArray(node.batteryLevels)
                                        ? 'All battery levels: ' +
-                                         node.batteryLevels.map((v) => `${v}%`).join(',')
+                                         Object.values(node.batteryLevels).map((v) => `${v}%`).join(',')
                                        : 'Unknown battery level'
                                if (level <= 10) {
                                        icon = mdiBatteryAlertVariantOutline
```

Probably something happens during transfering data from server to client ?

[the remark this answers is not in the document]
kvaster: I've updated PR - removed sorting and useless checks.

src/lib/utils.js
```diff
@@ -166,3 +166,12 @@ export function arraysEqual(a, b) {
 
 	return true
 }
+
+export function getBatteryDescription(node) {
+	return typeof node.batteryLevels !== 'undefined'
+		? 'All battery levels: ' +
+				Object.values(node.batteryLevels)
+					.map((v) => `${v}%`)
+					.join(',')
```
AlCalzone, line 175:
```suggestion
					.join(', ')
```

AlCalzone (changes requested):
Just a minor thing

robertsLando (approved):

PR #3132, opened 2023-06-08
Add default WebSocket URL to Home Assistant section. Closes #3120.

### LINKED ISSUE #3120: Suggest adding the default URL of a0d7b954-zwavejs2mqtt to the docs instead of localhost
It took me forever to find this in the Home Assistant docs that the URL is ws://a0d7b954-zwavejs2mqtt:3000 instead of ws://localhost:3000. In fact, I don't see this URL anywhere in https://zwave-js.github.io/zwave-js-ui/#/usage/setup?id=home-assistant.

I'd be happy to make a pull request to https://github.com/zwave-js/zwave-js-ui/blob/master/docs/homeassistant/homeassistant-official.md if you agree but maybe there's a reason you're not putting this URL in?

Thanks for your consideration.

zachfeldman:
I'd wished this URL was in this section per my filed issue - adding it to the docs!


zachfeldman:
@kpine makes sense, sorry about that. I added it to that section instead, I also did look here for the URL in my troubleshooting path, so this would still be really useful I think for other folks getting set up. Thanks!

zachfeldman:
Ok @kpine , I'll close my pull request since you think others will find them confusing. Hopefully they'll at least find my issue and this PR when searching for solutions. Have a nice day!

PR #3139, opened 2023-06-11
feat: added state_class: measurement to appropriate sensors

btreecat:
There are several sensor types that do not currently report a property of "state_class": "measurement" which means home-assistant does not have the ability to show them in a statistics graph. This causes issues for certain types of UI cards as they are unable to display any data. 

Documentation for the values is here: 

https://developers.home-assistant.io/docs/core/entity/sensor/#available-state-classes

I have run `yarn test` and `yarn lint` successfully. I have also updated the matching test. 

Additionally when trying to run `yarn pkg` I ran into two issues.

1. `egrep` deprecation (migrated to `grep -E` per: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html)
1. `arch` command binary not available (migrated to `uname -m` per: https://linux.die.net/man/1/arch)

Testable container image is available here: `btreecat/zwave-js-ui-dev:feat-add-state-classes`

### DIFF
--- a/lib/Constants.ts
+++ b/lib/Constants.ts
@@ -17,6 +17,7 @@ interface ISensorType {
 }
 
 interface ISensorProps {
+	state_class?: string
 	device_class?: string
 	unit_of_measurement?: string
 	icon?: string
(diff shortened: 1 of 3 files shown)


robertsLando (approved):
Thanks for this PR @btreecat ! It looks good to me, do you need to make any other changes?

btreecat:
I was able to successfully test the built container on my home-lab k8s instance. Looks good from my side :smile: 

Thank you for this awesome project and making it so easy to contribute.

PR #3154, opened 2023-06-29
feat: custom return routes management

### LINKED ISSUE #3129: [Feature] Assigning custom Return Routes and Priority Return Routes
Routes computed by the zwave proprietary firmware are garbage with insane latencies and an useless number of hops. The application (us) knows better and we might either manually assign our own routes (computed by the best AI available: the zwave network owner) or devise an algorithm to brute force its way to best possible route.
We can already set Priority Routes but unfortunately that's not enough because[ this route is not set at the end node, only the controller will use this route. If the end node needs to send a message to the controller it will use this route if it is the LWR otherwise it will use one of its own assigned routes](https://drzwave.blog/2021/07/13/z-wave-mesh-priority-routes-explained/).
So we need to support setting `Priority Return Routes` along with the existing `Priority Routes`.
Even with the `Priority Return Routes` the node will fallback to the shitty firmware-made routes in case of failure, so it would be better to be able to also set (and possibly cache) `Return Routes` as well to avoid high latencies whenever a Priority Return Route fails. There should also be a way to restore the cached return routes because we cannot fetch them and there is no guarantee that they haven't been tampered with.
That would be the minimum to be able to start playing with better routes and it will be a good starting point to develop algorithms to brute force the best possible routes.

Being able to set all possible route types: https://drzwave.blog/2021/07/13/z-wave-mesh-priority-routes-explained/

robertsLando:
- [x] Refresh neighbors when opening mesh graph
- [x] Add support for `discoverNodeNeighbors`.
- [x] GET return routes:
	- [x]  `getPriorityReturnRoute`
	- [x] `getPrioritySUCReturnRoute`
	- [x] `getCustomReturnRoute`
	- [x] `getCustomSUCReturnRoute`
- [x] SET return routes:
	- [x] `assignReturnRoutes`
	- [x] `assignSUCReturnRoutes` 
	- [x] `assignPriorityReturnRoute`
	- [x] `assignPrioritySUCReturnRoute`
	- [x] `assignCustomReturnRoutes`
	- [x] `assignCustomSUCReturnRoutes`
- [x] DELETE return routes
	- [x] `deleteReturnRoutes`
	- [x]  `deleteSUCReturnRoutes` 
- [x] UI
	- [x] Show return routes on node panel with get/set/delete commands
	- [x] Show return routes on graph
- [x] Generate mqtt apis docs for new apis

Ref docs: https://zwave-js.github.io/node-zwave-js/#/api/controller?id=managing-routes

Fixes #3129


AlCalzone:
Further TODOs:

- [ ] In the return routes list, put the priority route always on top and do not make it draggable
- [ ] Table headers for the return routes (if it looks somewhat ok?)
- [ ] in the map, show a star on the priority return route too
- [ ] do not show "route failed between undefined --> undefined, undefined --> undefined" in the priority route editor
- [ ] add toggles to turn off rendering priority and return routes (can get chaotic)

PR #3165, opened 2023-07-12
fix: respect `event.deltaY` when scrolling in network map

AlCalzone:
as discussed in Discord, the current approach of `vis-network` in regards to scrolling results in terrible UX for trackpad users with pixel-wise scrolling.

This PR patches the `vis-network` bundle with essentially this change:
```diff
-          scale *= 1 + (event.deltaY < 0 ? 1 : -1) * (this.options.zoomSpeed * 0.1);
+          scale *= 1 - event.deltaY/120 * (this.options.zoomSpeed * 0.1);
```
120 seems to be a common scaling factor, so I used that and doubled the zoom speed because it felt slow.


robertsLando:
Considered also opening a PR on vis network package?

AlCalzone:
Not yet, wanted an easy way to test the change for someone with an affected track pad.

PR #3224, opened 2023-08-16
feat: add new controller rfregion and txpower settings

### LINKED ISSUE #3222: Add configuration settings for controller region and power levels
Z-Wave JS now supports persistent configuration of controller region and tx power levels via driver settings: https://github.com/zwave-js/node-zwave-js/pull/6159

This ensures the controller is always configured correctly, even if something like a firmware update reverts them to defaults.

Add matching Z-Wave configuration settings in ZUI. These should be optional, which would use the current behavior, that is the user sets them manually in the control panel.

None.

**Additional context**
An open question would be what happens if a user configures these, but then changes them manually in the control panel. Z-Wave JS would change them on next startup if still configured. Maybe that is OK, it could just be documented in the help text.

robertsLando:
Fixes #3222

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -577,7 +577,7 @@ export type ZwaveConfig = {
 	serverServiceDiscoveryDisabled?: boolean
 	maxNodeEventsQueueSize?: number
 	higherReportsTimeout?: boolean
-	rf: {
+	rf?: {
 		region?: RFRegion
 		txPower?: {
 			powerlevel: number
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -577,6 +577,13 @@ export type ZwaveConfig = {
 	serverServiceDiscoveryDisabled?: boolean
 	maxNodeEventsQueueSize?: number
 	higherReportsTimeout?: boolean
+	rf: {
+		region?: RFRegion
+		txPower?: {
+			powerlevel: number
+			measured0dBm: number
+		}
+	}
 }
 
 export type ZUIDriverInfo = {
@@ -2107,6 +2114,7 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 				userAgent: {
 					[pkgjson.name]: pkgjson.version,
 				},
+				rf: this.cfg.rf,
 			}
 
 			// ensure deviceConfigPriorityDir exists to prevent warnings #2374
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -2114,7 +2114,22 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 				userAgent: {
 					[pkgjson.name]: pkgjson.version,
 				},
-				rf: this.cfg.rf,
+			}
+
+			if (this.cfg.rf) {
+				const { region, txPower } = this.cfg.rf
+
+				if (region) {
+					zwaveOptions.rf.region = region
+				}
+
+				if (
+					txPower &&
+					typeof txPower.measured0dBm === 'number' &&
+					typeof txPower.powerlevel === 'number'
+				) {
+					zwaveOptions.rf.txPower = txPower
+				}
 			}
 
 			// ensure deviceConfigPriorityDir exists to prevent warnings #2374
--- a/src/components/nodes-table/NodeDetails.vue
+++ b/src/components/nodes-table/NodeDetails.vue
@@ -347,8 +347,8 @@
 import ValueID from '../ValueId'
 import { mapState, mapActions } from 'pinia'
 import { validTopic } from '../../lib/utils'
+import { rfRegions } from '../../lib/items'
 import { ConfigValueFormat } from '@zwave-js/core/safe'
-import { RFRegion } from 'zwave-js/safe'
 import useBaseStore from '../../stores/base.js'
 import InstancesMixin from '../../mixins/InstancesMixin.js'
 import { isUnsupervisedOrSucceeded } from '@zwave-js/core/safe'
@@ -385,12 +385,7 @@ export default {
 				parameter: 1,
 				valueFormat: ConfigValueFormat.UnsignedInteger,
 			},
-			rfRegions: Object.keys(RFRegion)
-				.filter((k) => isNaN(k))
-				.map((key) => ({
-					text: key,
-					value: RFRegion[key],
-				})),
+			rfRegions,
 		}
 	},
 	computed: {
--- a/src/lib/items.js
+++ b/src/lib/items.js
@@ -1,0 +1,8 @@
+import { RFRegion } from 'zwave-js/safe'
+
+export const rfRegions = Object.keys(RFRegion)
+	.filter((k) => isNaN(k))
+	.map((key) => ({
+		text: key,
+		value: RFRegion[key],
+	}))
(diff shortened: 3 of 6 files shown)


robertsLando:
@AlCalzone Should I set some hints in settings? Does passing `undefined` values create any issue?

AlCalzone:
`undefined` is fine, you just cannot set one of the `rf.txPower` settings. Need to be both or none.

robertsLando:
@AlCalzone Could it be I don't see any error when setting only one of them? I added a validation now anyway driver doesn't throw in that case

AlCalzone:
This check should catch it:
https://github.com/zwave-js/node-zwave-js/blob/9d9407fdddcfc1a81bfedd7b7d3ba9095da50794/packages/zwave-js/src/lib/driver/Driver.ts#L397-L412

robertsLando:
Thanks for the report @kpine. Fixed on master now

apella12:
Don't know if I should put up as a problem, new to zwave-js-ui, but with 700 controller was advised by Zooz.
```
Hi Robert,
Thank you for reaching out!
It should display as Tx Power Level Normal -8 dB and Tx Power Level Measured 0 dB.
Please let us know if you have any other questions
Sara
Zooz Support Team
```
I have two setups of zwave-js-ui Rpi3 on docker with
```
zwave-js-ui: 8.23.0.73606d4
zwave-js: 11.12.0
home id: 3306325200
home hex: 0xc51284d0
```
Could not change to -8 either under Control Panel (Controller) settings (reverts back to 0) or under the Setting Zwave.  Note the Setting (Zwave) did reflect the change.
pi3 setting zwave

but the Debug under the Control Panel (Controller) still shows the unchanged values.
```
"powerlevel": 0,
  "measured0dBm": 0,
  "RFRegion": 1,
```
and so does the Control Panel(Controller)
pi3controlpanelcontroller

and yes I did SAVE several times.

OTOH I have wave-js-ui on an RPI4 made from code using yarn and it changed fine.  Of course there was only one place on this earlier version before this PR to make the change.
```
zwave-js-ui: 8.22.3.8607bd8
zwave-js: 11.9.2
home id: 4056168225
home hex: 0xf1c43721
```
Could something in the PR be blocking the change?  Also as a side comment it is a little confusing to have two places to change these values.  Could copy this to a problem if you want, but it seems highly likely to be related to this PR.

apella12:
Did you click the Save button?
_-Yes as noted above when using the Settings Area.
-When in the Control Panel under the node I just used the arrow._

If yes, did you restart ZUI?
_-I'm new, but thought I read hitting save did restart ZUI? The controller stats were all reset and from my zsniffer all the nodes were sent a NOP.  Do I need to stop the docker container and restart it? or what?_

The difference is that the settings are applied on every startup.
- _So that is more permanent, yes?_

Zooz told me something different back in April. 
_-That needs to be sorted.  As a note, the Pi4 (that was changed) originally had those settings (6,0) as default (7.18.3).  What prompted me to ask in the first place was the Pi3 controller default (still 7.18.1) was (0,0) In my e-mail to zooz, I just said 700 controller.  Both were shipped with 7.18.3 earlier this year._

AlCalzone:
> I looks like the change failed twice once on startup and second on the Control panel

Agreed, it looks like the stick doesn't accept that powerlevel. Can you try a positive value just to check?

apella12:
Same story.  I tried the other Zooz settings. 

[zwave-js-ui-store (1).zip](https://github.com/zwave-js/zwave-js-ui/files/12456645/zwave-js-ui-store.1.zip)
 I want to upgrade to 7.18.3 to see if that is any different. Maybe it is the stick?

AlCalzone:
Before you upgrade, you can ask Zooz about it. Here's the corresponding part of the log, which shows that setting the powerlevel fails:
```
2023-08-28T16:31:51.183Z CNTRLR   Querying configured powerlevel...
2023-08-28T16:31:51.187Z SERIAL » 0x0104000b13e3                                                       (6 bytes)
2023-08-28T16:31:51.188Z DRIVER » [REQ] [SerialAPISetup]
                                    command: GetPowerlevel16Bit
                                    payload: 0x13
2023-08-28T16:31:51.193Z SERIAL « [ACK]                                                                   (0x06)
2023-08-28T16:31:51.195Z SERIAL « 0x0108010b1300000000ee                                              (10 bytes)
2023-08-28T16:31:51.196Z SERIAL » [ACK]                                                                   (0x06)
2023-08-28T16:31:51.198Z DRIVER « [RES] [SerialAPISetup]
                                    command:               GetPowerlevel16Bit
                                    normal powerlevel:     0.0 dBm
                                    output power at 0 dBm: 0.0 dBm
2023-08-28T16:31:51.201Z CNTRLR   Current powerlevel 0 dBm (0 dBm) differs from desired powerlevel 6 dBm (0 dBm)
                                  , configuring it...
2023-08-28T16:31:51.207Z SERIAL » 0x0108000b12003c0000d2                                              (10 bytes)
2023-08-28T16:31:51.208Z DRIVER » [REQ] [SerialAPISetup]
                                    command:               SetPowerlevel16Bit
                                    normal powerlevel:     6.0 dBm
                                    output power at 0 dBm: 0.0 dBm
2023-08-28T16:31:51.213Z SERIAL « [ACK]                                                                   (0x06)
2023-08-28T16:31:51.215Z SERIAL « 0x0105010b1200e2                                                     (7 bytes)
2023-08-28T16:31:51.216Z SERIAL » [ACK]                                                                   (0x06)
2023-08-28T16:31:51.217Z DRIVER « [RES] [SerialAPISetup]
                                    command: SetPowerlevel16Bit
                                    success: false
```

apella12:
I don't want to confuse zooz with data.  I upgraded to 7.18.3 and issue went away- No issue with this PR. case closed
I'm having trouble getting a straight answer on the power settings. Latest is this;
```
The SDK version does not matter. 
You can easily stick to the HA recommended setting or follow our settings which can be either 6 or 8 for for Power Level Normal and for Measured 0 dBm. Those are the default values for our ZST10-700 stick. 
Thanks
Bartek
```
I've asked for clarification since I was told -8, not plus 8.  I don't know if they are talking about the box with the "typical" 700 controller on the Zwave-js-ui or something else. It seems it doesn't matter much??

apella12:
Ok excerpts of my back and forth with zooz;
THEM
```
I am sorry for the confusion here. We double-checked with the factory once again and had confirmation about the recommended settings: 
1. Tx Power Level Normal -8 dB and Tx Power Level Measured 0 dB for the ZST39 USB stick
2. Tx Power Level Normal -5 dB and Tx Power Level Measured 0 dB for the ZAC93 GPIO module
Theoretically, the higher this value the better the signal. Based on FCC labs, to comply with the USA-related regulations, the suggested Tx Power Level Normal for the ZST39 module is -8dB. For the TX Power Level Measured, the lab does not test this. We suggest setting it as 0 as we have already done the calibration on the product in the lab.
```
ME
`Almost there.  Is the S2 ZST10 -700 the same as the ZST39?  That is what I have.`
THEM (answering your question, although I never sent that information (sent the (+6,0) from several posts above.
```
For ZST10-700 we recommend following the HA setting that you initially provided:
Tx Power Level Normal 0 dB and Tx Power Level Measured +3.3 dB for the ZST10-700 USB stick
```
What an ordeal

apella12:
The real question is what they haven't been set at ;-)
Based on the last email from zooz (finally answering the question for my devices) I set them both at (0,+3.3) this morning. Of note zwave-ui-js did not seem to change using the Node_1 off the control panel, but worked fine from settings/zwave/ and then save

apella12:
oh, sorry. I dont recall.  It may have been (6/0)

apella12:
I'm a minimalist regarding logging.  Sorry.  Also since I do not use HA I did the upgrade on Simplicity Studio.

PR #3244, opened 2023-08-27
fix: remove controller default powerlevel message

kpine:
The message regarding the typical controller powerlevel settings is not really accurate anymore. Both Zooz and Aeotec use different values. It probably only applies to the UZB7 developer reference board and the US region, but even SiLabs has never provided correct numbers. It's better to just remove these as it causes confusion when using other models of controllers.

### DIFF
--- a/src/components/nodes-table/NodeDetails.vue
+++ b/src/components/nodes-table/NodeDetails.vue
@@ -135,7 +135,7 @@
 				>
 					<small
 						>{{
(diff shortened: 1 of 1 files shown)


jfhautenauven (approved):

robertsLando (approved):

robertsLando:
Sorry was on vacation last week, thanks for the PR @kpine ! :)

PR #3263, opened 2023-09-05
chore: switch to vitejs

robertsLando:


robertsLando:
TODOs:

1. Proxy of websockets are not working: https://github.com/vitejs/vite/issues/4124
2. Check PWA is working
3. Try to lazy load components
4. Test everything works behind reverse proxy.

PR #3268, opened 2023-09-07
feat: bump zwave-js@12 and fix breaking changes

robertsLando:
Ref: https://zwave-js.github.io/node-zwave-js/#/getting-started/migrating-to-v12

### DIFF
--- a/lib/ZwaveClient.ts
+++ b/lib/ZwaveClient.ts
@@ -38,8 +38,8 @@ import {
 	FoundNode,
 	GetFirmwareUpdatesOptions,
 	guessFirmwareFileFormat,
-	HealNetworkOptions,
-	HealNodeStatus,
+	RebuildRoutesOptions,
+	RebuildRoutesStatus,
 	InclusionGrant,
 	InclusionOptions,
 	InclusionResult,
@@ -96,6 +96,7 @@ import {
 	ZWaveOptions,
 	ZWavePlusNodeType,
 	ZWavePlusRoleType,
+	FirmwareUpdateInfo,
 } from 'zwave-js'
 import { getEnumMemberName, parseQRCodeString } from 'zwave-js/Utils'
 import { logsDir, nvmBackupsDir, storeDir } from '../config/app'
@@ -175,7 +176,7 @@ export const allowedApis = validateMethods([
 	'replaceFailedNode',
 	'hardReset',
 	'softReset',
-	'healNode',
+	'rebuildNodeRoutes',
 	'getPriorityRoute',
 	'setPriorityRoute',
 	'assignReturnRoutes',
@@ -190,8 +191,8 @@ export const allowedApis = validateMethods([
 	'deleteReturnRoutes',
 	'deleteSUCReturnRoutes',
 	'removePriorityRoute',
-	'beginHealingNetwork',
-	'stopHealingNetwork',
+	'beginRebuildingRoutes',
+	'stopRebuildingRoutes',
 	'isFailedNode',
 	'removeFailedNode',
 	'refreshInfo',
@@ -525,7 +526,7 @@ export type ZUINode = {
 	interviewStage?: keyof typeof InterviewStage
 	status?: keyof typeof NodeStatus
 	inited: boolean
-	healProgress?: HealNodeStatus | undefined
+	rebuildRoutesProgress?: RebuildRoutesStatus | undefined
 	minBatteryLevel?: number
 	batteryLevels?: { [key: number]: number }
 	firmwareUpdate?: FirmwareUpdateProgress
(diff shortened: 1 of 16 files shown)


robertsLando:
cc @AlCalzone

src/views/ControlPanel.vue
```diff
@@ -532,10 +536,10 @@ export default {
 					const { includeSleeping } =
 						await this.$listeners.showConfirm(
 							'Info',
-							'Healing network causes a lot of traffic, can take minutes up to hours and users have to expect degraded performance while it is going on',
+							'Rebuilding nodes routes causes a lot of traffic, can take minutes up to hours and users have to expect degraded performance while it is going on',
```
AlCalzone, line 539:
```suggestion
							'Rebuilding routes causes a lot of traffic, can take minutes up to hours and users have to expect degraded performance while it is going on',
```

src/views/ControlPanel.vue
```diff
@@ -370,12 +370,13 @@ export default {
 							action: 'rebuildNodeRoutes',
 							args: {
 								confirm:
-									'Healing a node causes a lot of traffic, can take minutes up to hours and you can expect degraded performance while it is going on',
+									'Rebuilding node routes causes a lot of traffic, can take minutes up to hours and you can expect degraded performance while it is going on',
```
AlCalzone, line 373:
```suggestion
									'Rebuilding routes causes a lot of traffic, can take minutes up to hours and you can expect degraded performance while it is going on',
```

lib/ZwaveClient.ts
```diff
@@ -4551,14 +4551,14 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
-	 * Triggered on each progress of healing process
+	 * Triggered on each progress of rebuild routes process
 	 */
 	private _onRebuildRoutesProgress(
 		progress: ReadonlyMap<number, RebuildRoutesStatus>
 	) {
 		const toHeal = [...progress.values()]
 		const healedNodes = toHeal.filter((v) => v !== 'pending')
-		const message = `Healing process IN PROGRESS. Healed ${healedNodes.length} nodes`
+		const message = `Rebuild Routes process IN PROGRESS. Healed ${healedNodes.length} nodes`
```
AlCalzone, line 4561:
```suggestion
		const message = `Rebuild Routes process IN PROGRESS. ${healedNodes.length} nodes done.`
```

lib/ZwaveClient.ts
```diff
@@ -4602,7 +4602,7 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 	}
 
 	private _onRebuildRoutesDone(result) {
-		const message = `Healing process COMPLETED. Healed ${result.size} nodes`
+		const message = `Rebuild Routes process COMPLETED. Healed ${result.size} nodes`
```
AlCalzone, line 4605:
```suggestion
		const message = `Rebuild Routes process COMPLETED. ${result.size} nodes done.`
```

AlCalzone (changes requested):

lib/ZwaveClient.ts
```diff
@@ -4551,14 +4551,14 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
 
 	/**
-	 * Triggered on each progress of healing process
+	 * Triggered on each progress of rebuild routes process
 	 */
 	private _onRebuildRoutesProgress(
 		progress: ReadonlyMap<number, RebuildRoutesStatus>
 	) {
 		const toHeal = [...progress.values()]
 		const healedNodes = toHeal.filter((v) => v !== 'pending')
```
AlCalzone, line 4560: "heal" :)

package.json
```diff
@@ -162,7 +162,7 @@
     "vuedraggable": "^2.24.3",
     "vuetify": "^2.6.14",
     "winston": "^3.8.2",
-    "zwave-js": "12.0.0-beta.2"
+    "zwave-js": "^11.14.3"
```
AlCalzone, line 165: sure about that?

robertsLando: Ops 🙈

src/stores/base.js
```diff
@@ -21,6 +21,7 @@ const useBaseStore = defineStore('base', {
 			port: '/dev/zwave',
 			allowBootloaderOnly: false,
 			commandsTimeout: 30,
+			sendToSleepTimeout: 250,
```
AlCalzone, line 24: Do you need to set a default? This would override any changes I do in the driver, right?

robertsLando: Nope I don't need to as long as you accept undefined as value.

AlCalzone: yep:
grafik

PR #3299, opened 2023-09-24
fix(graph): match text color with icon color on ZWave Graph legend

PlasmaEye:
change the legend text color to match the legend icon's color.

Before (dark): 
image

Before (light):
image

After (dark):
image

After (light):
image


PlasmaEye:
I feel like I've met the requirements but the bot still isn't happy. What am I missing?

PR #3319, opened 2023-10-02
feat: show releases changelog on login

### LINKED ISSUE #3315: Show dialog with changelog on upgrades
Similar to statistics dialog when an upgrade is detected we should show the changelog in a popup when a user opens the UI, the changelog should be fetched from GH release body and shown only once.

May be useful to also include release notes of zwave-js as well when there is a driver update (this could be done by using a regex on changelog to get new driver version)

CC @AlCalzone

robertsLando:
Fixes #3315


cybergrimes:
This is fantastic. Thank you.

PR #3332, opened 2023-10-10
feat: add setting to enable new version notification

robertsLando:


robertsLando:
cc @frenck @AlCalzone this setting will be opt in

PR #3346, opened 2023-10-16
fix(ui): firmware update OTA result not shown

### LINKED ISSUE #3341: OTA update through update service: Error starting the update and update result isn't shown
I had an issue updating a device where the firmware update support information was incorrect, causing the firmware update to fail almost immediately. `controller.firmwareUpdateOTA` returned this result, but it was nowhere to be seen in the UI:
```ts
			const result: FirmwareUpdateResult = {
				success: false,
				status: FirmwareUpdateStatus.Error_TransmissionFailed,
				reInterview: false,
			};
```

After I managed to complete it, the result wasn't shown either, although this is in the logs:
```
2023-10-13 14:04:08.074 INFO Z-WAVE: [Node 005] Firmware update finished successfully.
   Status: OK_RestartPending.
   Wait before interacting: 40s.
   Result: [object Object].
```

After the interview, the already-installed firmware update is still shown too:
grafik

robertsLando:
Fixes #3341

### DIFF
--- a/src/views/ControlPanel.vue
+++ b/src/views/ControlPanel.vue
@@ -923,7 +923,8 @@ export default {
 								// handled in App.vue
 								break
 							}
-							case 'updateFirmware': {
+							case 'updateFirmware':
+							case 'firmwareUpdateOTA': {
 								const result = response.result
 
 								const title = `Firmware update ${


AlCalzone:
Doesn't seem to fix the issue I'm afraid

PR #3422, opened 2023-11-14
feat: add npm deploy and move backend source to `/api` folder

### LINKED ISSUE #3418: Publish package to npmjs registry
I want install `zwave-js-ui` on a LXC Alpine container, that it is running on a OpenWRT host.

It seems easier to install node.js from an APK and run `npm -g install zwave-js-ui` than download the packaged version, extract, etc. But above all it seems much easier to maintain and update it:
```
npm -g --depth=0 outdated
npm -g update zwave-js-ui
```
or maybe:
```
npm -g update
```
I don't know if I'm missing something, I don't usually use globally installed NPM packages. But I think it should work and it would be more comfortable.

https://docs.npmjs.com/updating-packages-downloaded-from-the-registry#updating-globally-installed-packages

robertsLando:
Fixes #3418


robertsLando:
TODOs:

- [ ] Docs
- [ ] Add npm publish to release workflow (requires token, see https://github.com/mqttjs/MQTT.js/blob/main/.github/workflows/release-it.yml#L42)

PR #3424, opened 2023-11-14
fix: user callbacks not working with MQTT

robertsLando:


robertsLando:
cc @sarakha63 give this a try

PR #3440, opened 2023-11-26
fix(docker): serialport prebuilds should be ok now, so can skip rebuilding it

chrisns:


robertsLando (changes requested):
Without that line and the changes I did related to node_modules, the install will be skipped and so the arm builds will ship with the amd64 binaries of serialport (that will not work at all on arms)

chrisns:
hold fire, need to get another PR fixed for serial-port

chrisns:
blocked by https://github.com/serialport/bindings-cpp/pull/163 or move to non-musl (alpine) based image

robertsLando:
#3443 fixes all the problems 🎉

Issue #3430, opened 2023-11-17
Delay between release and container publish


mmattel:
Q: when there is a new release like in the screenshot below, what is the expected update on hub.docker to make the release availabe?

**Repo**
![image](https://github.com/zwave-js/zwave-js-ui/assets/3321281/6de6c2d4-9d4c-45ce-a491-08ff025112b1)
**hub.docker**
![image](https://github.com/zwave-js/zwave-js-ui/assets/3321281/2de40767-8889-4afc-83f5-d5cef4b03b77)

myoung34:
This also bit me [here](https://github.com/myoung34/homelab/commit/1a7cb51e0e92068509ebfa0e75a775aa1e78793a)
It picks up releases but this project seems to consistently not publish

* https://github.com/zwave-js/zwave-js-ui/issues/3413
* https://github.com/zwave-js/zwave-js-ui/issues/3313

myoung34:
Seems to be related to this ![image](https://github.com/zwave-js/zwave-js-ui/assets/879922/58692caf-dac7-44bf-bdd0-cd300a73d078)

robertsLando:
I'm aware of this guys and I'm sorry for the inconvenience but I don't understand what's going on, it's an issue with zui but seems something related to github actions/docker or yarn....

myoung34:
The builds are timing out if you look at the error. Your best bet is to build them in parallel similar to how [I do](https://github.com/myoung34/docker-github-actions-runner/actions/runs/6898384474)

it's basically an update to your [yaml](https://github.com/myoung34/docker-github-actions-runner/blob/master/.github/workflows/deploy.yml)

robertsLando:
What I sincerly do not understand is why they always worked and then suddenly stopped working. It happened to me on another project with the same result, with armv6/v7 install simply hangs for no reason. Notice that the test workflows (each test runs on a specific arch) end successfully so the error happens only when building all archs together with buildx. This makes me think something may have changed on action runner env

myoung34:
It could be, but the blast radius would be smaller in parallel (they'd upload as they finish instead of failing at the very end for all if one does)

You could always try qemu/buildx instead of `uraimo/run-on-arch-action@v2` and see if that helps

robertsLando:
@myoung34 Wait, it's not `run-on-arch-action` that fails, it's docker build-push action

myoung34:
[...]

PR #3441, opened 2023-11-27
chore(docker): switch to non-alpine image and drop armv6

robertsLando:


robertsLando:
cc @chrisns . Temporary fix for #3430

PR #3447, opened 2023-11-28
feat: use `npm` instead of `yarn`

robertsLando:


AlCalzone (approved):

ahochsteger:
Just out of curiosity, why the switch from yarn to npm?

robertsLando:
See here: https://github.com/giaever-online-iot/zwave-js-ui/pull/5#issuecomment-1834125225

Issue #3473, opened 2023-12-07
OpenWrt (and other minimal embedded environment) support


dwmw2:
We package [Domoticz](https://domoticz.com/) for the OpenWrt distribution, and have historically used OpenZWave with it. Domoticz is now [switching](https://github.com/domoticz/domoticz/pull/5605) to ZWave-JS-UI so it would be great to package ZWave-JS-UI to work out of the box on OpenWrt too: https://github.com/nxhack/openwrt-node-packages/pull/1826

An always-on embedded router/AP is a perfect place to run home automation, but the environment can be a little constrained in CPU, memory and "disk" storage.  For the last few years mine has been running on a MIPS24k with 128MiB of RAM.

We have a proof of concept basically working _(not on that box, but on a quad Arm7 with 256MiB of RAM)_. I'd appreciate some help cleaning it up though, please. There are a few things to look at. Please let me know if I should file separate issues for each.

### 1. The package is huge — about 100MiB.
Are we doing it wrong? Are there any parts which can be left out? Here are some of the largest files....

```
644	./node_modules/vue/node_modules/@vue/compiler-sfc/dist/compiler-sfc.js
644	./node_modules/vue/packages/compiler-sfc/dist/compiler-sfc.js
676	./node_modules/vis-network/standalone/esm/vis-network.min.js
676	./node_modules/vis-network/standalone/umd/vis-network.min.js
704	./node_modules/vis-network/esnext/esm/vis-network.js
740	./node_modules/mqtt/dist/mqtt.js
748	./node_modules/vis-network/esnext/umd/vis-network.js
1136	./dist/assets/index-822e6be1.js
1192	./node_modules/vis-network/peer/esm/vis-network.js
1224	./node_modules/vis-network/peer/umd/vis-network.js
1364	./node_modules/vis-network/dist/vis-network.esm.js
1368	./node_modules/vuetify/dist/json/web-types.json
1400	./node_modules/vis-network/dist/vis-network.js
1552	./node_modules/vis-network/standalone/esm/vis-network.js
1592	./node_modules/vis-network/standalone/umd/vis-network.js
1628	./node_modules/vuetify/dist/vuetify.js
2748	./node_modules/@mdi/js/mdi.js
2928	./node_modules/@mdi/js/commonjs/mdi.js
```
[...]

PR #3480, opened 2023-12-11
feat: create esbuild bundle for embedded devices

robertsLando:
Ref #3473

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,5 @@ typings/
 !.yarn/versions
 
 fakeNodes.json
+
+/build/
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -4400,7 +4400,7 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 				controllerNode.statistics as ControllerStatistics
 			controllerNode.statistics = stats
 
-			if (stats.messagesRX > oldStatistics?.messagesRX ?? 0) {
+			if (stats.messagesRX > (oldStatistics?.messagesRX ?? 0)) {
 				// no need to emit `lastActive` event. That would cause useless traffic
 				controllerNode.lastActive = Date.now()
 			}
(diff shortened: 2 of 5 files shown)


dwmw2:
OK, I have a package building and working with these changes. Thanks! 

https://github.com/nxhack/openwrt-node-packages/commit/bcd6acf5b6cdc26bb301d3530cf110430d5e0f1a

Just a few nits...

To build against the release tarball, I needed to the patches to be against `server/api/{utils.js,ZWaveClient.js}` instead of the corresponding `.ts` files. But that'll just fall out when there's a release.

I also needed to change the `entryPoints` in `esbuild.js`  from `api/bin/www.ts` to `server/bin/www.js` though.

And when the package build does its `npm install`, I end up with the config in `node_modules/zwave-js/node_modules/@zwave-js/config` rather than `node_modules/@zwave-js/config` so for now I just made a symlink and then `npm run bundle` works.

robertsLando:
Humm insteresting... but what tarball are you donwloading? npm tarball?

robertsLando:
Would it help adding a flag like `--use-js-entrypoint` to handle that case on esbuild?

dwmw2:
Or could we make it use server/bin/www.js unconditionally? Don't we always have to build `api/bin/www.ts` → `server/bin/www.js` anyway? The original TypeScript file isn't even in the bundle, surely?

dwmw2:
https://janessagarrow.com/blog/typescript-and-esbuild/ seems to suggest that esbuild allows the `.ts` file to be specified because it might actually get built to multiple *different* `.js` files, for different environments. But that isn't a concern here as we're only building for node, right? So we can just unconditionally list `server/bin/www.js` as the entrypoint without needing an option?

robertsLando:
@dwmw2 reason is that compiling against ts with source maps set to true allows me to have a better stack errors as they will point to ts lines, otherwise it might be difficult to debug things

dwmw2:
Don't you get that from the source maps anyway?

robertsLando:
> Don't you get that from the source maps anyway?

Wrong, If I ship sourcemaps (with also the code inside `api` folder) I get full logs

PR #3510, opened 2024-01-08
feat(discovery): add support for availability topics

ccutrer:
Require the client to be connected (LWT will set to false when MQTT connection is lost because the process goes down), the driver to be ready, and the individual node to be online

### DIFF
--- a/api/lib/Gateway.ts
+++ b/api/lib/Gateway.ts
@@ -931,7 +931,7 @@ export default class Gateway {
 					// Set device information using node info
 					payload.device = this._deviceInfo(node, nodeName)
 
-					this._availabilityConfig(node, payload)
+					this.setDiscoveryAvailability(node, payload)
 
 					hassDevice.object_id = utils
 						.sanitizeTopic(hassDevice.object_id, true)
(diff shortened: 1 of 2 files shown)


robertsLando (changes requested):
Thanks for this PR! Did you tested it? Is it working well? I cannot test with HA or Domoticz as I don't use them. 

You may need to wrap this in a method on the client and add it also to the discoverDevice method

ccutrer:
I've tested it with the current releases of Home Assistant and openHAB (I primarily use the latter, and am the maintainer of its MQTT/Home Assistant compatibility layer). It's been working quite well - immediately showing the devices as offline if I kill zwave-js-ui, and then showing them as online when I restart it.

ccutrer:
> You may need to wrap this in a method on the client and add it also to the discoverDevice method

I've wrapped this in a method and tested the regular path, but I don't (currently) have any custom devices to test it with.

api/lib/Gateway.ts
```diff
@@ -2370,6 +2370,36 @@ export default class Gateway {
 		}
 	}
 
+	private _availabilityConfig(
```
robertsLando, line 2373: Only a naming preference: 
```suggestion
	private setDiscoveryAvailability(
```

robertsLando (changes requested):
All good for the rest!

api/lib/Gateway.ts
ccutrer: Done. Please keep me on my toes style-wise... I'm going to have a few more PRs, and TypeScript (and JavaScript) is definitely not my first language.

robertsLando (approved):
LGTM

PR #3525, opened 2024-01-17
feat(ui): custom browser TZ/LOCALE and UI persistent preferences

### LINKED ISSUE #2721: [feat]: Persist Rows per page on smart provisioning entities table
### Checklist

### Deploy method

Manually built (git clone - yarn build - yarn install

### Z-Wave JS UI version

8.2.0

### ZwaveJS version

10.3.0

### Describe the bug

On the /smart-start page, the "Rows per page" always reverts to '10' upon every refresh. This is unlike the persistence that is seen on other pages such as /control-panel where picking a value other than the initial one survives a page refresh.

### To Reproduce

1. Go to /smart-start page
2. Select a value other than '10' from "Rows per page"
3. Refresh page.
4. Value will be back to '10'

### Expected behavior

Chosen value for "Rows per page" should survive a page refresh.

### Additional context

_No response_

### LINKED ISSUE #3013: Configurable time and date format (12/24h & dd-mm-yyyy / mm-dd-yyyy / yyyy-mm-dd)
Even though I live in the USA I am European and 100% prefer the 24 hour system for time formatting. I could not find any setting to configure the time display to use the 24h system. Please add :)

image

Being able to configure the date format would be awesome as well. Please do not attach it to locale or some other default setting based on location as that forces formats that may not match what the user needs/wants.

Configuration for time and date formats in UI tab

None that I can think of

**Additional context**
None

Grazie Daniel!

### LINKED ISSUE #3524: Option to show node event timestamps in local timezone
The node events tab is a convenient place to quickly check if a node has reported some values, vs. going to the debug logs. However, the timestamps in the node events tab are always displayed in UTC time and if you aren't in that timezone, it takes extra effort to convert.

I'd like the option of showing the timestamps in the local timezone (e.g. the OS configured timezone) instead of UTC. Then I could just compare the event times to the "wall clock" and easily correlate them.

None

**Additional context**
See discussion https://github.com/zwave-js/zwave-js-ui/discussions/3517.

image

robertsLando:
Fixes #3524 
Fixes #3013
Fixes #2721

### DIFF
--- a/.env.app.example
+++ b/.env.app.example
@@ -1,17 +1,21 @@
 # The host address to bind to
-# HOST="::"
+HOST="::"
 # The port to listen to for incoming requests. Default is `8091`
-# PORT=8091
+PORT=''
 # The absolute path to the directory where all files will be stored. Default is `<path to your zui dir>/store`
-# STORE_DIR
+STORE_DIR=''
 # Used as secret for session. If not provided the default one is used
-# SESSION_SECRET
+SESSION_SECRET="zwavejsisawesome"
 
-
-# NETWORK_KEY
-# HTTPS=true
+# S0 Key
+NETWORK_KEY=''
+HTTPS=''
 # Set the cookie [secure](https://github.com/expressjs/session#cookiesecure) option.
-# USE_SECURE_COOKIE
+USE_SECURE_COOKIE=''
 # Mostly needed for docker users.
 # https://zwave-js.github.io/node-zwave-js/#/usage/external-config?id=specifying-an-external-config-db-location
-# ZWAVEJS_EXTERNAL_CONFIG
+ZWAVEJS_EXTERNAL_CONFIG=''
+
+# Browser preferred locale/tz to use for date/time formatting 
+TZ=''
+LOCALE=''
--- a/api/app.ts
+++ b/api/app.ts
@@ -1037,6 +1037,8 @@ app.get(
 			serial_ports: [],
 			scales: scales,
 			sslDisabled: sslDisabled(),
+			tz: process.env.TZ,
+			locale: process.env.LOCALE,
 			deprecationWarning: process.env.TAG_NAME === 'zwavejs2mqtt',
 		}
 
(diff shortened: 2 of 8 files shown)


robertsLando:
@kpine Could you tell me if this looks good to you? I have created a store function to format dates in locale date/time strings, the tz and locale can be set using env vars

robertsLando:
@kpine try now

robertsLando:
I think it makes more sense to have them in reverse order by default, let me change that

robertsLando:
Ok I have created a Pinia `preferences` prop where we can store the UI proferences to/from localstorage of browser. For now I added smart start table options (like sorting, items per page etc...) and the events list vars, could add more in the future

dwmw2:
Please don't use locales and preferences for dates. Just default to YYYY-MM-DD for dates everywhere.
http://david.woodhou.se/isodate.html explains why.

dwmw2:
> Hi @dwmw2, while I agree with you, I think many users will complain about that... I could make that the default BTW.

It's definitely a good default. Let users jump through hoops if they want to do something different. Perhaps a checkbox for _"Use obsolete parochial local format for dates instead of the global standard"_, which enables the selection of *which* locate :)

Although most users will realise that YYYY-MM-DD dates are fairly ubiquitous, once they start to notice them.

> Dunno if there is a locale that reflects that format.

If you're *only* using it for the date format, I think en_CA would work.

```
 $ date +%x
18/01/24
 $ LC_ALL=en_US date +%x
01/18/2024
 $ LC_ALL=en_CA date +%x
2024-01-18
```

robertsLando:
ATM the default is to show dates in browser locale, I would also be curious to know @kpine and @AlCalzone opinion about this

dwmw2:
Oh wow, that's awful. So HA doesn't even have the *option* of showing dates in the standard YYYY-MM-DD form? Let alone not doing it right by default?

Definitely not a role model.

robertsLando:
>  So HA doesn't even have the option of showing dates in the standard YYYY-MM-DD form

Based on his screenshot it does, see the last option in the select input opened

dwmw2:
That does 2024/1/18. It should be 2024-01-18.

Two digits for month and never 1. There's a different strftime format character for that 

And most importantly, `-` instead of `/`. Which makes it immediately visually distinct from the legacy formats. Whenever you see any of those you get a "pipeline stall" and have to stop and work out what order things are in (even if it ends up being the "correct" one for your locale). When you see YYYY-MM-DD you don't get that. It's completely unambiguous. The dashes are an important part of that.

PR #3539, opened 2024-01-19
fix(discovery): improve MQTT discovery for scene CC

ccutrer:
* don't expose writable values as read-only scenes
 * give HA a hint when values are numbers and not strings


robertsLando (approved):

robertsLando:
LGTM

PR #3548, opened 2024-01-23
fix(ui): power level and measured power at 0 dbm limits

### LINKED ISSUE #3547: Range of TX power and measured at 0 dBm settings should be changed
The serial API is a bit confusing, since there are two commands (SetPowerlevel and SetPowerlevel16Bit) for this with different limits. With the current range of -12.8...+12.7 dBm, we can't use the entire allowed range of ZWLR or Z-Wave Classic in EU. I recently learned that ZW in EU should use +13 dBm, not 0...

The ranges (in the Z-Wave settings, and in the controller node accordion) should be changed to:
TX Power: -10 dBm ... +20 dBm
Measured at 0: -10 dBm ... +10 dBm

Note that setting values above +12.7 dBm will fail on controllers that do not support the `SetPowerlevel16Bit` command, but I think we can live with that.

robertsLando:
Fixes #3547


src/views/Settings.vue
```diff
@@ -644,10 +644,10 @@
...
-											:min="-12.8"
-											:max="12.7"
+											:min="-10"
+											:max="20"
 											:step="0.1"
-											hint="Power level in dBm. Min -12.8, Max 12.7"
+											hint="Power level in dBm. Min -10, Max +20"
 											suffix="dBm"
 											type="number"
 											:rules="[validTxPower]"
```
AlCalzone, line 653: Can't see it but does this method need an update aswell?

robertsLando: yeah nice catch :) Fixed on master

PR #3569, opened 2024-01-29
feat(ui): allow to toggle persistent/discovery fields from HA discovery table

### LINKED ISSUE #3558: Blacklist some (but not all) HA config discoveries
I have highlighted in the OpenHAB (OH) community that ZUI can be used with OH in place of the OH Z-wave binding (and have implemented that myself).  I think this would provide the OH community with additional features and ZUI with additional customers.
However, I implemented ZUI without the HA discovery because during testing several HA configs do not work within OH (The details are TL;DR for this request).  Instead I used generic MQTT things (in OH lingo) directly from the zwave section of MQTT. This is okay for many OH users, but for less technical ones I had hoped for HA config discovery.

Ability in the Node HASS tab to disable some of the config IDs, but not all.  (This may be possible already, but all I have managed to do is disable all or none.)  I'd also add that recently added configuration CC configs (number and switch) be disabled by default.  Those are in the "static" category and IMO the ratio of clutter to usefulness is high. (I was changing two configurations (112 params) in a 35 node network and noticed within the HA config test environment have 600 new HA configs

Before the recent change (112 parameters) I registered and an issue with jinjava about using numbers as keys.  Jinjava is what OH uses to decode the HA configs.  There is also a PR to address, but it hasn't been merged for several months (don't know why, except the developers do not see as a problem). The other prior problem is with scenes (91) a null value creates a warning in a number channel, but works as a string item.  Since the recent change there are numerous WARN messages because the device config files are not always right for the version of the device. (again tl;dr regarding examples).  Rather than fix potentially dozens of device configs I'm suggesting the blacklist idea and that those be disabled by default.

**Additional context**
These probably won't mean much, but are from the OH logs.

Original issue.

`Executing the JINJA-transformation failed: An error occurred while transformation. InterpretException: Error resolving expression [{22: "Window/door is open",5632: "Window/door is open in regular position",5633: "Window/door is open in tilt position"}[value_json.value] | default(value_json.value)]: TemplateStateException: Dict key must be a string or identifier, was: 22`

Configuration issues (Configs 9 & 10) do not exist for my device (tried to change in ZUI, but returns "0"
```
Command '0' from channel 'mqtt:homeassistant_zwavejs2mqtt_5F0xf1c43721_5Fnode9:6035d9536c:zwavejs2mqtt_5F0xf1c43721_5Fnode9:zwavejs2mqtt_5F0xf1c43721_5F9_2D112_2D0_2D9#number' not supported by type 'NumberValue': 0 is out of range
Command '0' from channel 'mqtt:homeassistant_zwavejs2mqtt_5F0xf1c43721_5Fnode9:6035d9536c:zwavejs2mqtt_5F0xf1c43721_5Fnode9:zwavejs2mqtt_5F0xf1c43721_5F9_2D112_2D0_2D10#number' not supported by type 'NumberValue': 0 is out of range
```

robertsLando:
Fixes #3558

### DIFF
--- a/src/components/nodes-table/HomeAssistant.vue
+++ b/src/components/nodes-table/HomeAssistant.vue
@@ -100,10 +100,23 @@
 						{{ item.object_id }}
 					</template>
 					<template v-slot:[`item.persistent`]="{ item }">
-						{{ item.persistent ? 'Yes' : 'No' }}
+						<v-checkbox
+							v-model="item.persistent"
+							@click.stop
+							@change="updateDevice(item)"
+							hide-details
+							dense
+						></v-checkbox>
 					</template>
 					<template v-slot:[`item.ignoreDiscovery`]="{ item }">
-						{{ item.ignoreDiscovery ? 'Disabled' : 'Enabled' }}
+						<v-btn
+							@click.stop="toggleField(item, 'ignoreDiscovery')"
+							:color="item.ignoreDiscovery ? 'error' : 'success'"
+							rounded
+							x-small
+						>
+							{{ item.ignoreDiscovery ? 'Disabled' : 'Enabled' }}
+						</v-btn>
 					</template>
 				</v-data-table>
 			</v-col>
@@ -115,7 +128,7 @@
 							color="blue darken-1"
 							:disabled="errorDevice"
 							text
-							@click="addDevice"
+							@click="addDevice()"
 							>Add</v-btn
 						>
 					</template>
@@ -129,7 +142,7 @@
 							color="blue darken-1"
 							:disabled="errorDevice"
 							text
-							@click="updateDevice"
+							@click="updateDeviceJSON()"
 							>Update</v-btn
 						>
 					</template>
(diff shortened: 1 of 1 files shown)


apella12:
Awesome ! Tested a bit this am and it's GREAT. 

I need to familiarize myself with how this works to be more efficient. What I did was disable the HA ids that I did not want, hit store (makes them persistent on Startup?).  Go to MQTT Explorer to delete the HA topics.  Restart ZUI to make sure, then restart OH to clear the unwanted topics.

Was thinking the best way for a new installation was to set HA to Manual, edit the HA ids in the nodes, then switch back to Auto.  Does that make sense?

Anyway, thanks again.

robertsLando:
> Was thinking the best way for a new installation was to set HA to Manual, edit the HA ids in the nodes, then switch back to Auto. Does that make sense?

Exactly that's why the manual discovery exists :) Glad you like this addition, wondering why I didn't did it before now 😆

Issue #3570, opened 2024-01-31
Scene Activation CC (43) support with MQTT Discovery


barrydouu:
### Version

Build/Run method:

[ X] Docker
[ ] Git Clone + npm install + npm run build + npm start
[ ] Pkg

Z-Wave JS UI version: 9.8.2
Z-Wave JS version: 12.4.3

### Describe the bug

I have a ZME_RC2  remote controller. It was working well until I upgrade to the last version (I don't know which version I was using before, one from the end of 2023), 
After troubleshooting, I ve found that the change states are not published to "right" mqtt topic.
In home assistant section, i see hass device json with command topic like : 
"command_topic": "zwave/27/112/0/10/set",

But when I check with mqtt explorer, when the device change, the value is set in "zwave/27/43/0/"

Here is debug information when I press a key on remote controller
```
2024-01-31 20:56:38.707 DEBUG MQTT: Publishing to zwave/27/lastActive: { time: 1706734598707, value: 1706734598706 } with options { qos: 1, retain: false }
2024-01-31 20:56:38.722 DEBUG MQTT: Publishing to zwave/27/43/0/sceneId: {
  time: 1706734598721,
  value: 72,
  nodeName: 'Telecommande',
  nodeLocation: ''
} with options { qos: 1, retain: false }
2024-01-31 20:56:38.724 INFO Z-WAVE: [Node 027] Value notification: 43-0-sceneId 72
2024-01-31 20:56:38.729 DEBUG MQTT: Publishing to zwave/27/43/0/dimmingDuration: {
  time: 1706734598726,
  value: Duration { unit: 'default', _value: 0, [value]: [Getter/Setter] },
  nodeName: 'Telecommande',
  nodeLocation: ''
} with options { qos: 1, retain: false }
2024-01-31 20:56:38.731 INFO Z-WAVE: [Node 027] Value updated: 43-0-dimmingDuration default => default
2024-01-31 20:56:38.959 DEBUG MQTT: Publishing to zwave/27/lastActive: { time: 1706734598959, value: 1706734598706 } with options { qos: 1, retain: false }
2024-01-31 20:56:39.724 DEBUG MQTT: Publishing to zwave/27/43/0/sceneId: {
  time: 1706734599724,
  value: undefined,
  nodeName: 'Telecommande',
  nodeLocation: ''
} with options { qos: 1, retain: false }
```

Here is the full Hass Device Json
```
{
  "type": "number",
  "object_id": "config_number_16",
[...]

PR #3581, opened 2024-02-03
change default for central_scene value_template for MQTT AD

ccutrer:
to be a number, not a string, since it's a number data type now

refs #3570

tested against openHAB and Home Assistant


ccutrer:
ugh, this won't work. 0 is a valid scene id. I'll look into it more tomorrow

ccutrer:
See #3582

PR #3582, opened 2024-02-05
fix(mqtt-ad): revert setting state_class on central scene entities

ccutrer:
While it is a number, it also often has no value at all, and Home Assistant can't deal with that properly, so leaving it as a string that can default to an empty string when it has no value is a better solution.

Refs #3570


robertsLando (approved):

robertsLando:
@kpine Could you also give me your 👍🏼 on this?

ccutrer:
Closed in favor of a full revert: #3583

PR #3671, opened 2024-04-17
docs: add FAQ

macpoupou:
Adding some tips to fix issues with access to zwave usb devices.


robertsLando:
@macpoupou Did you tried using the `/dev/serial/by-id` path instead?

macpoupou:
Yes but in Debian this path is not available. But yes it's something that we can add also.

robertsLando:
I would suggest to create a udev rule that handles that instead, otherwise the soft reset could break that, see https://github.com/zwave-js/zwave-js-ui/discussions/3589

Also instead of adding that FAQ in docs I would prefer you to open a discuttion type Q&A with the quesrtion and the answe marked :)

PR #3684, opened 2024-04-23
feat(ui): smart start view improvements

### LINKED ISSUE #3674: Smart Start Screen Data Inconsistencies
### Checklist

### Deploy method

Docker

### Z-Wave JS UI version

9.11.1.9e7d24e

### ZwaveJS version

12.5.5

### Describe the bug

edit2: view the below comment from @robertsLando on changes: https://github.com/zwave-js/zwave-js-ui/issues/3674#issuecomment-2068960023

edit: After looking at the code, I think this is the intended behavior. [disable check](https://github.com/zwave-js/zwave-js-ui/blob/9e7d24eef6afc5889f3f3729670349b8f205c2b0/src/views/SmartStart.vue#L30)

After upgrade from `9.9.1` to `9.11.1`, the smart start screen has an issue loading all of the edit buttons, check boxes, protocol toggles.

image

### To Reproduce

Have smart start items.
Upgrade docker from `9.9.1` to `9.11.1`.
Unsure if it's some edge case from my personal Smart Start items.

### Expected behavior

Shows all edit buttons, able to check all boxes, able to toggle the protocols.

### Additional context

Looks like the items upgraded correctly with the recent updates
from
```json
{
"k": "controller.provisioningList",
    "v": [
        {
            //...
            "supportedProtocols": [
                0,
                1
            ],
            //...
        }
    ]
}
```
to
```json
{
"k": "controller.provisioningList",
    "v": [
        {
            //...
            "supportedProtocols": [
                "ZWave",
                "ZWaveLongRange"
            ],
            //...
        }
    ]
}

```

robertsLando:
- Add a way to easily switch to the node attached to a specific provisioning entry (either by opening a modal there or a link to control panel)
- Add an info panel to explain some useful info about smart start entries
- Fill node manufacturer/model/protool informations when missing from QR code but existing nodeId

Fixes #3674


AlCalzone:
Some feedback:
- when including a node via SmartStart while the SmartStart page is open, the "link" to the node does not get updated. We need to be careful though in case the user is editing something, so we don't reload or discard their changes.
- Should the info panel be there already? I don't see it.
- The hover effect on the node ID link is too subtle IMO
- maybe add a tooltip on the grayed out fields of an included node: "The node is already included, so editing its SmartStart settings is not possible."

src/components/custom/MissingKeysAlert.vue
```diff
@@ -0,0 +1,43 @@
...
+	<v-alert
+		max-width="600px"
+		dense
+		border="left"
+		type="warning"
+		v-if="missingKeys.length > 0"
+	>
+		Some security keys are missing:
+		<strong>{{ missingKeys.join(', ') }}</strong
+		>. Please check your zwave settings.
```
AlCalzone, line 11:
```suggestion
		>. Please check your Z-Wave settings.
```

AlCalzone (changes requested):
Also, I'd only show the hint about not being able to edit nodes as a tooltip when hovering the disabled fields.

PR #3692, opened 2024-04-26
fix(ui): use vuetify's material design palette colors for protocol color

acshef:
This PR is intended to update the colors of the protocol icons to those of the [Material design palette offered by Vuetify](https://vuetifyjs.com/en/styles/colors/#material-colors). This matches the color palette of the other columns in the node table.

Considerations:
* In the spirit of an easily digestible PR, I am not addressing non-Material colors elsewhere in the ZWJS UI
* I tried using `colors.indigo.base` for `Protocols.ZWave`, but the contrast between ZWave and ZWaveLongRange wasn't great. There isn't a good color in the material design palette that's close to the outdated `rgb(0, 0, 255)` color -- perhaps one of the indigo/blue accent colors would better match the original author's intention.
* There's no requirement to adhere to the original author's intention -- it's not like we're using colors chosen by the Z-Wave Alliance to represent protocol icons, so we could pick something other than blue if it feels like those are too close to the app's primary color.

Before:
image

After:
image
image

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -1,4 +1,5 @@
 import { isValidDSK, Protocols } from '@zwave-js/core/safe'
+import colors from 'vuetify/lib/util/colors'
 
 export function copy(o) {
 	return JSON.parse(JSON.stringify(o))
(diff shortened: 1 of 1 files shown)


robertsLando (approved):

robertsLando:
Thanks @acshef !

PR #3706, opened 2024-05-08
feat: zniffer

robertsLando:


AlCalzone:
I'll just put my thoughts here...

The settings for Zniffer should **probably** not offer the serial port that's already in use by the controller:
grafik


I don't think generating keys there makes sense, as you want to spy on an existing network with existing keys:
grafik
We may also have to consider setting them on the fly to retroactively decode some commands, but that will require support in the driver first.

AlCalzone:
Filtering nodes makes sense in the table view but not here IMO:
grafik

AlCalzone:
Saving settings for the Zniffer should probably **not** restart the controller.

AlCalzone:
grafik

CAPTURE should be renamed so it's clear that it saves to a file.

robertsLando:
> The settings for Zniffer should probably not offer the serial port that's already in use by the controller:

Yeah but what if the driver is disabled? (still have to add this option)

> I don't think generating keys there makes sense, as you want to spy on an existing network with existing keys:

Should I drop keys from there and use the one set on driver? Or add a button to clone the ones on driver? 

> We may also have to consider setting them on the fly to retroactively decode some commands, but that will require support in the driver first.

👍🏼 

> Saving settings for the Zniffer should probably not restart the driver for the controller.

Yeah agree

> There's an error in the devtools and nothing is being shown in the table:

That error happens because I'm using ZWaveFrameType that seems not exported in `safe`: https://github.com/zwave-js/zwave-js-ui/pull/3706/files#diff-71a3cb674c70b3fa9b3f52e9afc7d873f7d0eeefcbccb7175f1448a816811a99

AlCalzone:
> Should I drop keys from there and use the one set on driver? Or add a button to clone the ones on driver?

Cloning might make sense to debug the current network.

Other than that, some observations (I know the layout is not optimized yet):

- [ ] Starting the Zniffer is somewhat unreliable. I have to sometimes start/stop twice. With [my test script](https://github.com/zwave-js/node-zwave-js/blob/master/test/run_zniffer.ts) it works fine.

- [ ] The "menu" bar should be sticky IMO:
grafik

- [ ] The region column is not necessary in the table. We can only zniff one region at once and that is a configuration setting. It should probably be in the Zniffer UI though like in the Windows app.

- [ ] Home ID should be formatted as HEX

- [ ] Type is meant for the application to decide how to display a frame, not for the end user.

- [ ] Sequence No, TX Power should be in the frame details (which we don't have yet), as it only exists for some of them.

- [ ] Source, destination and repeaters should probably be merged into a single column, that also highlights which hop the frame was sent through, similar to the log output `[070 » 069 › 001]`

- [ ] Protocol Data Rate takes up a lot of room. I'd probably use the icons we also use in the node list, and only display the speed as text.

AlCalzone:
Additional findings in the latest version:

- [ ] Timestamp should have fractions of seconds `HH:mm:ss.fff`. Not sure about the date - it seems that it takes away unnecessary room in the table.
- [ ] Wrong enum for protocol data rate, should be `ZnifferProtocolDataRate`
- [ ] When enabling "convert RSSI" in the settings, the RSSI column should show `${rssi} dBm`, not `${rssiRaw}`
- [ ] Delta[ms] makes more sense directly after the timestamp column
- [ ] I'd put Home ID before the source/dest columns
- [ ] Regarding the details pane:
    grafik
- [ ] Nested CCs aren't shown - this should contain an IndicatorSet: 
grafik
- [ ] Can't reproduce it now, but some `toLogEntry` results are multiline, the details view currently ignores `\n`

AlCalzone:
More things:

- [ ] Routed frames seem to be shown as buffers, not CCs
- [ ] The selected row should be highlighted
- [ ] When selecting a row, pause autoscroll. Re-enable when scrolling all the way down (or using a button in the UI)

AlCalzone:
- [ ] Dragging results in weird layout jumps:
[Screencast from 14.05.2024 12:12:06.webm](https://github.com/zwave-js/zwave-js-ui/assets/17641229/0cb6ba2e-dfc8-485f-aa48-659f37f436d9)

AlCalzone:
- [ ] Allow negative search filters: grafik

AlCalzone:
- [ ] Save Capture should download the file, like NVM backup does

robertsLando:
Thanks @kpine ! About the lost frames, I dunno if there is an api actually to fetch in memory frames from zniffer, @AlCalzone ?

I could store them on my side but I think the zniffer instance already have them somewhere in order to create the capture.

Also it's a nice idea to be able loading a capture file from UI, didn't thought about that. Cc @AlCalzone 

About the rest I can take a look tomorrow 🙏🏻

AlCalzone:
- [ ] This tooltip doesn't make sense. You're now showing the current frequency, right?
    grafik
    I'd just remove the text below the dropdown. 

- [ ] The RSSI unit is duplicated when having "convert RSSI" enabled:
    grafik

- [ ] Timestamps should be zero-padded to 3 digits:
    grafik

AlCalzone:
The layout bothers me a bit:
grafik



Also, is there any way to give the columns a fixed/min width to avoid this?
grafik
Maybe even disable the sorting altogether?

AlCalzone:
Yep I should really just expose the captured frames so Z-UI can restore the state from those, and add a hook to actually clear them.

AlCalzone:
I feel like this is going into nitpick territory, but number columns should be right-aligned:
grafik

AlCalzone:
Regarding formatting of the route - I'd also put that in the overview table, so you can see the route there immediately.
Routed frames have a `hop` property that tells us where along the route we are. The `direction` property tells us in which direction along the route we are going and does not have to be listed separately.
grafik

For reference, this is how this frame's route is formatted in the logs:
grafik

AlCalzone:
- [ ] Wound be nice to see if the Zniffer is running without being on the tab, for example:
    grafik

robertsLando:
TODOs:

- [ ] https://github.com/zwave-js/zwave-js-ui/pull/3706#issuecomment-2111844512
- [ ] Save capture should download the file not storing it on store (or both?)
- [ ] Protocol Data Rate takes up a lot of room. I'd probably use the icons we also use in the node list, and only display the speed as text. cc @AlCalzone ?
- [ ] Understand if removing security keys from settings
- [ ] Don't allow driver and zniffer to use the same port (when both are enabled)

AlCalzone:
- [ ] Raw payload in details view is taking up too much room. One or two lines should be enough. If the payload is very long, it can expand automatically.
- [ ] Raw payload should have a label (e.g. `Raw:`) in front of it
- [ ] Probably move the raw payload below the parsed payload?
- [ ] Probably omit the `0x` from buffers
- [ ] All buffers should be displayed in a monospace font (raw payload, CC payload)
- [ ] The resize bar should have a slightly gray background IMO
- [ ] Use the new parameter of `znifferProtocolDataRateToString` to omit the protocol name, use icons to distinguish Z-Wave and LR
- [ ] `trimStart` on log entry details:
    grafik
- [ ] Add line number as first column
- [ ] Show somewhere how many frames have been captured
- [ ] Indicate "speed modified" in the Protocol Data Rate column, e.g. with a 🐌 icon to the right of the speed:
    grafik
- [ ] Choose a default height for the details pane that is high enough for the entire table to be visible:
    grafik
- [ ] Add a button to resume auto-scrolling

AlCalzone:
> - [x] Add a button to resume auto-scrolling

Am I blind? :)

robertsLando:
@AlCalzone About the button I forgot to ask you what you mean because actually auto scroll is always enabled except if you have a frame opened in details, do you want to be able to re-enable it when you have a frame opened so?

PR #3710, opened 2024-05-12
feat: bump @zwave-js/server to 1.35.0

raman325:
No real changes from the beta but just to keep up with the latest :)


robertsLando (approved):
Thanks @raman325 ! :)

PR #3738, opened 2024-06-02
fix(ui): nodes table header from breaking line

zackbcom:
## Changes
Adjusts the headers in the nodes-table to not break lines.
Adjust minor css of headers to make it pleasant

## Testing Artifacts
- Before:
  - image
- After:
  - image


robertsLando (approved):
Well done with this! I always hated that line break on headers.

LGTM :)

PR #3740, opened 2024-06-02
feat(ui): updates to dark mode logic

zackbcom:
## Changes
- Updates to dark mode logic to have 3 options, System defined, Dark and, Light.
- Reused bool in the settings to use `null` as system defined.
- Created listener for changes to system dark mode. Works on mac when selecting system theme.
- Added button toggle to support the 3 cases.

## TODO
- [ ] Need to figure out what to do with the login checkbox. 
  - [ ] Might be able to look into a indeterminate state for system.
  - [ ] Replace with button stack
  - [ ] Remove and use system default, update in settings.

## Testing Artifacts
- After
  - image


src/App.vue
```diff
@@ -666,6 +669,8 @@ export default {
 			hideTopbar: false,
 			title: '',
 			messages: [],
+			systemDark: false,
+			mq_system_dark: null,
```
robertsLando, line 673: why don't use a watch for this instead?

src/App.vue
```diff
@@ -1612,6 +1629,14 @@ export default {
...
 		})
 	},
+	beforeUnmount() {
+		if (this.mq_system_dark !== null) {
+			this.mq_system_dark.removeEventListener(
+				'change',
+				this.updateSystemDark,
+			)
+		}
+	},
```
robertsLando, line 1639: this is a bit useless as the App will only be unmounted when the page is destroyed

src/App.vue
robertsLando: Also it's not clear to me what `mq_system_dark` is used for, you never re-assign it, I would instead use a let in beforeMount instead

robertsLando (changes requested):

src/App.vue
zackbcom: First time using Vue. Will re look at the logic here. Thanks for the notes.

robertsLando: It's ok no worries I appreciate a lot your help 🙏🏼

zackbcom:
I don't have time right now to work on this.

robertsLando:
@zackbcom No worries, keep it open for now will see if I can find some time to look at it :)

latyii (approved):

robertsLando:
Supersided by #4266

PR #3745, opened 2024-06-03
fix: set up events before calling `Zniffer.init()`

AlCalzone:
fixes: #3744

### DIFF
--- a/api/lib/ZnifferManager.ts
+++ b/api/lib/ZnifferManager.ts
@@ -105,6 +105,25 @@ export default class ZnifferManager extends TypedEventEmitter<ZnifferManagerEven
 		parseSecurityKeys(config, znifferOptions)
 
 		this.zniffer = new Zniffer(config.port, znifferOptions)
+		this.zniffer.on('error', (error) => {
+			this.onError(error)
+		})
+
+		this.zniffer.on('frame', (frame, rawData) => {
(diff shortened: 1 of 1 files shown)


api/lib/ZnifferManager.ts
```diff
@@ -113,33 +132,8 @@ export default class ZnifferManager extends TypedEventEmitter<ZnifferManagerEven
...
-			this.zniffer.on('error', (error) => {
-				this.onError(error)
-			})
-
-			this.zniffer.on('ready', () => {
-				logger.info('Zniffer ready')
-			})
 		} catch (error) {
-			this.onError(error)
-
```
AlCalzone, line 138: This is being called anyways in the `error` event, so this would just log twice

api/lib/ZnifferManager.ts
```diff
@@ -113,33 +132,8 @@ export default class ZnifferManager extends TypedEventEmitter<ZnifferManagerEven
...
 			logger.info('Retrying in 5s...')
-
-			await this.close()
```
AlCalzone, line 141: This prevents reusing the Zniffer instance. Alternatively, constructing the Zniffer instance would have to be done in `this.init()` too.

zackbcom (approved):

robertsLando (approved):

PR #3785, opened 2024-06-24
feat: show supported controller rf regions

robertsLando:
Ref: https://github.com/zwave-js/node-zwave-js/pull/6957

### DIFF
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -6032,12 +6032,11 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		node.hasDeviceConfigChanged = zwaveNode.hasDeviceConfigChanged()
 
 		if(node.isControllerNode) {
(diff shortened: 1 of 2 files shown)


api/lib/ZwaveClient.ts
```diff
@@ -6029,6 +6030,15 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 
 		node.deviceId = this._getDeviceID(node)
 		node.hasDeviceConfigChanged = zwaveNode.hasDeviceConfigChanged()
+
+		if(node.isControllerNode) {
+			node.rfRegions = this.driver.controller.getSupportedRFRegions()?.toSorted().map(
```
AlCalzone, line 6035: Z-Wave JS sorts by region number already. I think this sorts by the string representation of the numeric value?

AlCalzone (dismissed):

api/lib/ZwaveClient.ts
robertsLando: @AlCalzone LOL nope, I wanted to do that but I have to do the sort later. Let me fix it

PR #3799, opened 2024-07-08
feat: support for Z-Wave JS v13

AlCalzone:
TODO:

- [x] Upgrade to an official version of zwave-js-server
- [x] Wait for the official release of zwave-js
- [x] Test a bit more
- [x] Show the reason why adding an association failed in the UI, not only "error while adding"


api/lib/ZwaveClient.ts
```diff
@@ -1802,7 +1798,8 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 						this.logNode(
 							zwaveNode,
 							'warn',
-							`Unable to add Node ${a.nodeId} to Group ${groupId} of ${sourceMsg}, association not allowed`,
+							// FIXME: We should explain the reason why it failed in human-readable form, see doc comments of AssociationCheckResult
```
AlCalzone, line 1801: still TODO

api/lib/ZwaveClient.ts
```diff
@@ -4602,9 +4599,11 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		}
 	}
 
-	private _onInclusionStarted(secure: boolean) {
+	private _onInclusionStarted(strategy: InclusionStrategy) {
+		const secure = strategy !== InclusionStrategy.Insecure
 		const message = `${secure ? 'Secure' : 'Non-secure'} inclusion started`
 		this._updateControllerStatus(message)
+		// FIXME: Should the frontend also accept the strategy instead of the boolean?
 		this.emit('event', EventSource.CONTROLLER, 'inclusion started', secure)
```
AlCalzone, line 4607: See FIXME comment

api/lib/ZwaveClient.ts
```diff
@@ -127,11 +128,6 @@ export const configManager = new ConfigManager({
 	deviceConfigPriorityDir,
 })
 
-export async function loadManager() {
-	await configManager.loadNamedScales()
-	await configManager.loadSensorTypes()
-}
```
robertsLando, line 133: why this isn't needed anymore?

api/lib/ZwaveClient.ts
robertsLando: do you mean on frontend side?

api/lib/ZwaveClient.ts
robertsLando: Frontend uses `startInclusion` and passes the mode:

https://github.com/zwave-js/zwave-js-ui/blob/80c1cd86edf77c6b375ae2046e7d506d52f483d3/src/components/dialogs/DialogNodesManager.vue#L995-L998


https://github.com/zwave-js/zwave-js-ui/blob/80c1cd86edf77c6b375ae2046e7d506d52f483d3/src/components/dialogs/DialogNodesManager.vue#L1149-L1156

api/lib/ZwaveClient.ts
robertsLando: Ok got it:

```ts
import { getAllNamedScaleGroups, getAllSensors } from '@zwave-js/core'
```

robertsLando:
@AlCalzone Only missing piece on my side seems a missing safe export of `AssociationCheckResult`

api/lib/ZwaveClient.ts
AlCalzone: Both logs and frontend IMO

api/lib/ZwaveClient.ts
AlCalzone: I meant the event that is emitted below the comment. Right now it gets passed `secure`, I wonder if it should now include the strategy instead.

api/lib/ZwaveClient.ts
robertsLando: Check what i did and let me know if it's ok for you

AlCalzone: Although, if we do https://github.com/zwave-js/zwave-js-ui/issues/3801, we don't have to show the error reason in the logs I think, because there shouldn't be one.

api/lib/ZwaveClient.ts
robertsLando: that event actually is only emittet to MQTT when `Emit events` flag is enabled in MQTT settings, nothing used by UI btw. I can fix that and use the inclusion strategy instead

api/lib/ZwaveClient.ts
robertsLando: there shouldn't doesn't mean there will not, also consider that api can be also called from driver function and MQTT so I would leave it there

PR #3802, opened 2024-07-09
feat(ui): show association error in association dialog

### LINKED ISSUE #3801: Show in the association dialog if and why an association is now allowed
There are a lot of reasons why an association may not be allowed. Z-Wave JS v13 lets applications distinguish between them. We should expose the information in a way that's understandable for end users.

Something like this, but with different messages for each reason:
grafik

Also, the ADD button should be disabled when it's not allowed to add an association.

Filter out associations that may not be allowed, but this probably leads to more confusion

**Additional context**
none

robertsLando:
Fixes #3801

### DIFF
--- a/src/components/dialogs/DialogAssociation.vue
+++ b/src/components/dialogs/DialogAssociation.vue
@@ -73,7 +73,7 @@
 										v-model="group.target"
 										:items="filteredNodes"
 										return-object
-										:rules="[required]"
+										:rules="[required, allowedAssociation]"
 										hint="Node to add to the association group"
 										persistent-hint
 										item-text="_name"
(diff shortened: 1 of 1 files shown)


src/components/dialogs/DialogAssociation.vue
```diff
@@ -190,6 +190,20 @@ export default {
...
+			if (!mainNode || !targetNode) {
+				return true
+			}
+
+			if (mainNode.security !== targetNode.security) {
+				return `Association not allowed, "${targetNode._name}" security is different from "${mainNode._name}"`
+			}
+
+			return true
+		},
```
AlCalzone, line 206: No, use `Controller.checkAssociation` to figure out if the association is allowed and if not, why not. Then display a different message for each reason.

AlCalzone: The screenshot in the issue was just **one** example.

AlCalzone (changes requested):

src/components/dialogs/DialogAssociation.vue
robertsLando: Is that actually exported from safe somewhere?

AlCalzone: The idea is to make a backend api call, the backend calls driver.controller.checkAssociation and passes the result back.

This requires v13 because the method and AssociationCheckResult was only added there.

robertsLando:
Supersided by #3804

PR #3812, opened 2024-07-15
chore: do not watch local zwave-js repo

AlCalzone:
Since the switch back to `npm` this no longer works


robertsLando (approved):
Thanks! LGTM

PR #3813, opened 2024-07-15
chore: add `esModuleInterop` flag

robertsLando:


robertsLando:
cc @jmgiaever

jmgiaever:
Removes the particular issue, but seems there's conflicts with TS versions.

PR #3814, opened 2024-07-16
feat: link reliability checks

robertsLando:


robertsLando:
cc @AlCalzone

AlCalzone:
this is pending a node-zwave-js release

robertsLando:
Bumped

src/components/dialogs/DialogLinkReliability.vue
```diff
@@ -283,13 +292,10 @@ export default {
 
 			if (response.success) {
 				this.showSnackbar('Link statistics aborted', 'success')
+				this.running = false
 			}
-
-			this.running = false
 		},
 		async checkLinkReliability() {
-			this.running = true
```
AlCalzone: This...

src/components/dialogs/DialogLinkReliability.vue
```diff
@@ -306,10 +312,9 @@ export default {
 				},
 			)
 
-			this.running = false
-
 			if (response.success) {
 				this.statistics = response.result
+				this.running = true
```
AlCalzone: ...and this doesn't make sense. Awaiting `checkLinkReliability` is the entire process, so after it returns, the check is done and is no longer running.

AlCalzone (changes requested):

AlCalzone:
I swear I won't bother you with lots of polishing for now, but a progress bar (indeterminate state for infinite rounds) would be nice:
grafik

src/components/dialogs/DialogLinkReliability.vue
robertsLando: Oh didn't know that I thought the api was only used to start the process!

PR #3823, opened 2024-07-22
fix(ui): use inverted checkbox to clarify settings

robertsLando:


robertsLando:
cc @AlCalzone

AlCalzone:
There are a few more:
grafik
(also why is this a checkbox when all others are a toggle?)

grafik

grafik

robertsLando:
So should I only change the dns discovery one?

AlCalzone:
I guess, yeah

Issue #3427, opened 2023-11-16
ZWaveJS fails to load when running with external auth due to service worker


ajacques:
### Checklist

- [X]  I am not using Home Assistant. Or: a developer has told me to come here.
- [X] I have checked the troubleshooting section and my problem is not described there.
- [X] I have read the changelog and my problem is not mentioned there.

### Deploy method

Docker

### Z-Wave JS UI version

9.3.2

### ZwaveJS version

12.3.0

### Describe the bug

I'm using an external auth provider (nginx + [vouch proxy](https://github.com/vouch/vouch-proxy)) to protect all of my services with one SSO solution. I choose to do this instead of having separate username and passwords for different services. Effectively how it works is it checks for a cookie on all requests to ZWaveJS and redirects the browser to the login page. 

The problem is that ZWaveJS now appears to use a service worker and works as a PWA (I suspect this came in with this commit d409051b9e96939fa0510399e45693b331a34cf1) which intercepts the call to `/` and causes this to fail. When the authorization expires, I have to hard refresh the page to load it. 

What happens is the `/` page loads from the Service worker, then the `GET /api/auth-enabled` call gets a 302 Found to redirect to the login page, the app doesn't know what to do and fails to load the app. The `/` request does not hit the backend app (I've checked my logs), so it can't know that `/` is failing. The only way to bypass this is to hard refresh the page, but that's not easy on mobile phones.

### To Reproduce

1. Install ZWaveJS behind some kind of reverse proxy such as nginx
2. On the reverse proxy, enable an auth proxy that intercepts requests, such as [vouch proxy](https://github.com/vouch/vouch-proxy)
3. Try to load the ZWaveJS and see that it gives a "Network Error. Retry" and the page fails
4. Hard refresh the page (ctrl-shift-r) and notice the app loads because it bypasses the service worker for `GET /` and the app will work until the auth token expires.

### Expected behavior

I expect the ZWaveJS application to be able to handle when there's a separate auth provider and redirect to the login page. ZWaveJS does not have to specifically have to support these providers, just handle the redirects.

I see two options here:

1. When the service worker receives the `GET /` request, it revalidates with the service to see if the response includes a redirect or not. This can also be used to refresh the cache in the service worker. (basically cache-bust the `/` call because the app can't work without internet anyway)
[...]

PR #3825, opened 2024-07-22
fix(ui): trigger page reload on 401 (fix basic auth)

oandrew:
# Changes
Throw exception when `401` returned from `/auth-enabled` to trigger page reload. This is needed for  http basic auth to work properly.

# Details
I run zwave-js-ui behind basic http auth and get  "Unexpected end of JSON input" error after browser restart:
Screenshot 2024-07-22 at 2 33 55 PM

This happens when `401` status is returned since Chrome keeps auth only until browser restart.
Typically Chrome would display the auth popup in this case  but turns out it doesn't work requests made by service workers:  
 https://issues.chromium.org/issues/40474438

This is essentially the same issue as  #3427 

FWIW Without this PR, it can be "fixed" with the following Caddy config (send `302` when `Authorization` header is missing)
```
(auth) {
        basicauth {
                ...
        }
}

zwave.mydomain.com {
        @auth_check {
                path /api/auth-enabled
                header !Authorization
        }
        # trigger redirect when auth header not set
        handle @auth_check {
               redir zwave.mydomain.com 302
        }
        handle {
                import auth
                reverse_proxy 127.0.0.1:8091
        }
}
```

### DIFF
--- a/src/apis/ConfigApis.js
+++ b/src/apis/ConfigApis.js
@@ -66,7 +66,7 @@ export default {
 				Accept: 'application/json',
 			},
 		})
-		if (response.type === 'opaqueredirect') {
+		if (response.type === 'opaqueredirect' || response.status === 401) {
 			throw new axios.AxiosError(
 				'Caught redirect for auth-enabled, rethrowing',
 				response.status,


robertsLando (approved):
LGTM

PR #3892, opened 2024-09-15
feat: add INGRESS_TOKEN env variable to set the ingress session cookie

darkbasic:
`SERVER_URL=https://192.168.1.x:8123/api/hassio_ingress/abcdefg INGRESS_TOKEN=123456789  npm run dev`

image

I needed a way to connect my local development frontend to the remote zwave-js-ui backend. Since the HASSIO docker container doesn't expose the port to the host the easiest way around it is via [Hass.io Ingress](https://www.home-assistant.io/blog/2019/04/15/hassio-ingress/). Unfortunately that needs cookie authentication, which can be easily injected via Vite's built-in proxy. This PR adds an `INGRESS_TOKEN` env variable where you can specify the token and it will take care of injecting the cookie for each API request. Use Chrome/Firefox DevTools to figure out what's the correct Ingress `SERVER_URL` and `INGRESS_TOKEN`.


robertsLando (approved):
LGTM

PR #3893, opened 2024-09-16
fix(ui): routed ack frames row style

### LINKED ISSUE #3775: Zniffer: Routed acks / errors should be shown in the frame list, not only the details
These are frames with `routedError: true`:
grafik

We should show that those are routed errors and color them red in this view.

Likewise those are frames with `routedAck: true`:
grafik

We should highlight them like ACKs and show in the payload that they are routed acks.

robertsLando:
Fixes #3775


robertsLando:
cc @AlCalzone

PR #3902, opened 2024-09-18
feat: make default credentials for auth customizable

karmingc:
The current implementation defaults to username: `admin` and password: `zwave` when enabling auth. However, it can also be beneficial to make this customizable so that it can avoid some additional steps in updating these values.This is completely backwards compatible as well.

@robertsLando I didn't see any contributing guidelines, but if there's anything missing, please let me know.


robertsLando (approved):
Thanks for the PR @karmingc! LGTM

PR #3905, opened 2024-09-19
fix(discovery): define units for scene and configuration values

ccutrer:
And for scene values, use the correct template because the value is a JSON object that also includes the units, but nests the actual value one level deeper.

### DIFF
--- a/api/lib/Gateway.ts
+++ b/api/lib/Gateway.ts
@@ -1336,6 +1336,10 @@ export default class Gateway {
 						valueId.property,
 						valueId.propertyKey,
 					)
+					if (valueId.value?.unit) {
+						cfg.discovery_payload.value_template =
+							"{{ value_json.value.value | default('') }}"
+					}
 					break
 				case CommandClasses['Binary Sensor']: {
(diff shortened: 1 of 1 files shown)


api/lib/Gateway.ts
```diff
@@ -1336,6 +1336,10 @@ export default class Gateway {
 						valueId.property,
 						valueId.propertyKey,
 					)
+					if (valueId.value?.unit) {
+						cfg.discovery_payload.value_template =
+							"{{ value_json.value.value | default('') }}"
```
robertsLando, line 1341: I think you have a typo here? 

```suggestion
							"{{ value_json.value | default('') }}"
```

ccutrer: No, this is intentional, and the critical part of this chain. Example data from MQTT:

`zwave/Garage/Loft_Stairwell_Lights/38/0/duration`
```json
{"time":1726797698474,"value":{"value":0,"unit":"seconds"}}
```

Notice value is nested within value.

robertsLando: Oh didn't know about that! Thanks for clarification

robertsLando (approved):

PR #3946, opened 2024-10-12
fix(ui): scanning of small QR codes

AlCalzone:
https://github.com/nimiq/qr-scanner uses the native [BarcodeDetector](https://developer.chrome.com/docs/capabilities/shape-detection#barcodedetector) if available in browsers (essentially only Chrome), otherwise defaults to a JS port of ZXing. That port seems to be suboptimal, since it really struggles with the small Z-Wave QR codes.

https://www.npmjs.com/package/barcode-detector is a polyfill based on the WASM version of ZXing, which works much, much better. I was able to scan a tiny, blurry, distorted QR code in a second, without having to select "small QR code", or aiming a lot.

We need to decide how to include the wasm module though, since we don't want to download it at runtime, see https://github.com/Sec-ant/barcode-detector?tab=readme-ov-file#setzxingmoduleoverrides


robertsLando:
@AlCalzone You mean this?

```
// src/index.ts
import wasmFile from "../node_modules/zxing-wasm/dist/reader/zxing_reader.wasm?url";

import {
  setZXingModuleOverrides,
  BarcodeDetector,
} from "barcode-detector/pure";

setZXingModuleOverrides({
  locateFile: (path, prefix) => {
    if (path.endsWith(".wasm")) {
      return wasmFile;
    }
    return prefix + path;
  },
});

const barcodeDetector = new BarcodeDetector();

// detect barcodes
// ...
```

AlCalzone:
Yes, minus the `const barcodeDetector = ...`

src/components/custom/QrReader.vue
```diff
@@ -133,12 +133,29 @@
 	</div>
 </template>
 <script>
-import QrScanner from 'qr-scanner'
+// The BarcodeDetector Web API is not yet supported in all browsers,
+// and "qr-scanner" defaults to a suboptimal implementation if it is not available.
+// The following import makes a better implementation available that is based on a
+// WebAssembly port of ZXing:
+import { setZXingModuleOverrides } from 'barcode-detector/pure'
+import 'barcode-detector/side-effects'
```
AlCalzone, line 141:
```suggestion
import { setZXingModuleOverrides } from 'barcode-detector'
```
should work as well

robertsLando (approved):

src/components/custom/QrReader.vue
```diff
@@ -133,12 +133,29 @@
...
+// The following import makes a better implementation available that is based on a
+// WebAssembly port of ZXing:
+import { setZXingModuleOverrides } from 'barcode-detector'
 import { wait } from '../../lib/utils.js'
 import logger from '../../lib/logger'
 
 const log = logger.get('QrReader')
 
+import QrScanner from 'qr-scanner'
+
```
AlCalzone, line 147: Double check why this import is now here. It might not be needed.

robertsLando: It is actually needed: https://github.com/zwave-js/zwave-js-ui/blob/f909f84185cdad7bcf6e8b7d37c65da1b781cad2/src/components/custom/QrReader.vue#L290

AlCalzone: Yeah but why was it not there before the PR?

robertsLando: it was on the first line, check the diff :)

AlCalzone: Ohh, yeah. I was blind.

PR #3960, opened 2024-10-22
feat: zwave-js@14.3.0 and @zwave-js/server@1.40.0

AlCalzone:
I'll collect the necessary changes here for when v14 gets released.


package.json
robertsLando, line 1: what changes did you do on this file except the dep bump?

AlCalzone: AFAIK nothing. Seems to be a whitespace change.

AlCalzone: I'll try and clean that up before the final merge.

AlCalzone:
Bumped the dependencies, but we're still waiting for a server release.

PR #3981, opened 2024-11-05
feat(ui): re-order Z-Wave settings, prevent setting invalid regions

AlCalzone:
This PR groups the Z-Wave settings panel a bit more logically, adds subheadings and explains some settings a bit better. Also it is no longer possible to choose "Unknown" and "Default (EU)" as default regions, since they are not a valid setting.


robertsLando (approved):
LGTM

PR #3985, opened 2024-11-06
feat: allow to upload firmware in zip files

robertsLando:
Fixes https://github.com/zwave-js/zwave-js-ui/issues/3983

### DIFF
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -18,6 +18,7 @@ import {
 	Protocols,
 	createDefaultTransportFormat,
 	FirmwareFileFormat,
+	tryUnzipFirmwareFile,
 } from '@zwave-js/core'
 import { JSONTransport } from '@zwave-js/log-transport-json'
 import { isDocker } from '@zwave-js/shared'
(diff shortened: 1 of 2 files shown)


robertsLando:
@AlCalzone About `tryUnzipFirmwareFile` should I check the file extension first and then use it before guess ?

AlCalzone:
Almost correct. If ".zip" use `tryUnzipFirmwareFile`, otherwise use `guessFirmwareFileFormat`.
`tryUnzipFirmwareFile` does the guessing aswell, otherwise the zip file would have to be parsed twice.

robertsLando:
Ok so the code I wrote should be correct, right?

AlCalzone (approved):

AlCalzone:
Looks good. Just need to wait for 14.1.0 to finish building.

api/lib/ZwaveClient.ts
```diff
@@ -3892,13 +3894,28 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 			const firmwares: Firmware[] = []
 
 			for (const f of files) {
-				const { data, name, target } = f
+				let { data, name } = f
 				if (data instanceof Buffer) {
```
robertsLando, line 3898: @AlCalzone What's the correct way to check buffer here? Should I check if it's an Uint8Array

AlCalzone: To test whether something is a `Uint8Array`, use the `isUint8Array` function exported from `node:util/types` (not portable) or `@zwave-js/shared` (portable).

Issue #4053, opened 2024-12-05
9.27.8 / 14.3.7 - crashed after working for some hours after update


kpanchen:
### Checklist

- [X]  I am not using Home Assistant. Or: a developer has told me to come here.
- [X] I have checked the troubleshooting section and my problem is not described there.
- [X] I have read the changelog and my problem is not mentioned there.

### Deploy method

PKG executable

### Z-Wave JS UI version

9.27.8

### ZwaveJS version

14.3.7

### Describe the bug

After update from 9.27.7 worked for some hours and stopped responding. UI is not responding, MQTT seems publishing but not responding to updates. Restart won't fix and issue. Reverting back to 9.27.7 works fine.

In the log just see the line:
`2024-12-05T15:40:29.423Z DRIVER   Dropping message because the driver is not ready to handle it yet.`

Line repeats from time to time.

Logs attached. Note: that log includes both failure and working normally after reverting back.
Let me know if more information required.

### To Reproduce

Update to a new version.

### Expected behavior

Operate normally

[zwavejs_current.zip](https://github.com/user-attachments/files/18029933/zwavejs_current.zip)
[z-ui_current.zip](https://github.com/user-attachments/files/18029935/z-ui_current.zip)

### Additional context

Not sure about this error:
`2024-12-05T15:40:36.525Z CONFIG   Writing the device index to disk failed: ENOENT: no such file or directory, op
                                  en '/snapshot/build/node_modules/@zwave-js/config/config/devices/index.json'`

robertsLando:
You should set the env var `ZWAVEJS_EXTERNAL_CONFIG=./store/.config-db` fix this command in order to customize the destination folder

robertsLando:
Starting from next minor version setting that env var will not be needed anymore: https://github.com/zwave-js/zwave-js-ui/pull/4055

kpanchen:
Thanks!! Had to do a bit of googling but understand the issue now, guess I'll wait for a fix on your side, prev version just works fine! Thanks a lot for the answer!!!

Issue #4003, opened 2024-11-13
On update to 9.27.1, all devices show unknown manufacturer, product, product code


luckyjay:
### Checklist

- [X]  I am not using Home Assistant. Or: a developer has told me to come here.
- [X] I have checked the troubleshooting section and my problem is not described there.
- [X] I have read the changelog and my problem is not mentioned there.

### Deploy method

PKG executable

### Z-Wave JS UI version

9.27.1

### ZwaveJS version

14.3.2

### Describe the bug

I hope I didn't do something stupid, but after upgrading from 9.26.0 to 9.27.1, all my zwave devices are listed in the UI with unknown manufacturer, product, product code
![image](https://github.com/user-attachments/assets/a19bf6a0-eb3b-4666-998f-c368b02c3b33)
 

### To Reproduce

Update from 9.26.0 to 9.27.1.

### Expected behavior

No unknowns in manufacturer, product, product code columns

### Additional context

Falling back to 9.26.0 no longer shows unknowns.

Re-interviewing on 9.27.1 will complete, but does not correct the unknowns even after restarting the service.

Environment is Ubuntu 24 LTS x64 VM, fully patched/updated running on Proxmox.

luckyjay:
Note: I was running the 9.27.1 executable from https://github.com/zwave-js/zwave-js-ui/releases/download/v9.27.1/zwave-js-ui-v9.27.1-linux.zip

zmcnaney:
Same issue here - simply updated it in place using:

`RELEASE=$(curl -s https://api.github.com/repos/zwave-js/zwave-js-ui/releases/latest | grep "tag_name" | awk '{print substr($2, 2, length($2)-3) }')
  systemctl stop zwave-js-ui.service
  wget https://github.com/zwave-js/zwave-js-ui/releases/download/${RELEASE}/zwave-js-ui-${RELEASE}-linux.zip &>/dev/null
  unzip zwave-js-ui-${RELEASE}-linux.zip &>/dev/null
  \cp -R zwave-js-ui-linux /opt/zwave-js-ui
  service_path="/etc/systemd/system/zwave-js-ui.service"
  echo "[Unit]
  Description=zwave-js-ui
  Wants=network-online.target
  After=network-online.target
  [Service]
  User=root
  WorkingDirectory=/opt/zwave-js-ui
  ExecStart=/opt/zwave-js-ui/zwave-js-ui-linux
  [Install]
[...]

PR #4055, opened 2024-12-09
feat: add default value for external Z-Wave JS config database path

robertsLando:
This will prevent issues when not setting this env var causing `Writing the device index to disk failed: ENOENT: no such file or directory` on pkg

Ref #4053 #4003


robertsLando:
cc @kpine @AlCalzone

AlCalzone (approved):

PR #4104, opened 2025-01-27
fix: store nodes.json using homeHex

### LINKED ISSUE #3817: After switching controllers, the node names and locations are attributed to the wrong devices
grafik

I previously had another stick connected with 3 nodes, using different IDs 2, 3 and 4, which had these names and locations set. After switching sticks, the names and locations persist in the UI, but they belong to the wrong devices.

robertsLando:
Fixes #3817

### DIFF
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -812,23 +812,6 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		this.scenes = jsonStore.get(store.scenes)
 
 		this._nodes = new Map()
-		this.storeNodes = jsonStore.get(store.nodes)
-
-		// convert store nodes from array to object
-		if (Array.isArray(this.storeNodes)) {
-			const storeNodes = {}
-
-			for (let i = 0; i < this.storeNodes.length; i++) {
-				if (this.storeNodes[i]) {
-					storeNodes[i] = this.storeNodes[i]
-				}
-			}
-
-			this.storeNodes = storeNodes
-
-			// eslint-disable-next-line @typescript-eslint/no-floating-promises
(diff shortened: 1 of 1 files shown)


AlCalzone (commented):
Does this gracefully handle existing `nodes.json`? It's not entirely clear to me how much magic is behind the scenes.

I wouldn't want to delete all names and locations for users who have them set.

api/lib/ZwaveClient.ts
```diff
@@ -2483,10 +2466,54 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+
+		const keys = Object.keys(nodes)
+
+		// ensure store nodes are stored using homeHex
+		if (keys.length > 0 && !keys[0].startsWith('0x')) {
+			this.storeNodes = nodes
+			await jsonStore.put(store.nodes, {
+				[this.homeHex]: nodes,
+			})
+		} else {
```
robertsLando, line 2499: this is the back compatibility trick I added. Basically I just check if the keys in nodes.json are strings starting with 0x (homeHex), if not it means them are not converted to the new format yet so I store the keys in the new format

robertsLando: To keep this simply and less triky I decided to change the format of nodes.json instead of changing the filename itself as that would have required a rewrite of the jsonstore that I prefer to avoid

AlCalzone:
Ok, let me do a quick test

AlCalzone:
We should probably also clean up these entries when migrating:
grafik

robertsLando:
@AlCalzone done

AlCalzone:
Also, should we delete entries when hard-resetting the controller? Afterwards the current home ID is "lost", unless someone restores an NVM backup. Not entirely sure.

PR #4139, opened 2025-03-10
feat: support changing Zniffer LR channel config, dynamically populate frequency list

AlCalzone:
This PR adds support for changing the Zniffer LR channel configuration on supporting Zniffers, when the current region is LR capable:
grafik

In addition, the region list is now populated dynamically, so now it uses the region names exposed by Z-Wave JS for the actual stick - here for a 500 series Zniffer:

grafik

The PR is based on the v15 branch and the unmerged PR https://github.com/zwave-js/zwave-js/pull/7665, so this will have to wait a bit and needs to be rebased before being merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [20.x, 22.x]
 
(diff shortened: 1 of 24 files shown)


robertsLando (dismissed):
Looks good! :) Thanks 🙏🏼  Will wait for the official release then before merging :)

PR #4159, opened 2025-03-20
fix: config sync inside `pkg` binaries

AlCalzone:
fixes: https://github.com/zwave-js/zwave-js-ui/issues/4156

### DIFF
--- a/api/lib/PkgFsBindings.ts
+++ b/api/lib/PkgFsBindings.ts
@@ -1,0 +1,86 @@
+import { FileHandle, FSStats, type FileSystem } from '@zwave-js/shared/bindings'
+import { fs as nodeFs } from '@zwave-js/core/bindings/fs/node'
+import path from 'node:path'
+
+// Ensures that the Z-Wave JS driver is looking for the right files in the right place
+// when running inside a `pkg` bundle. In this case, it will resolve its embedded
+// configuration dir to the path "/config", but the files reside in "node_modules/@zwave-js/config/config" instead.
+
+const CONFIG_PATH = '/config'
+const CONFIG_PATH_IN_PKG = path.join(
+	__dirname,
+	`node_modules/@zwave-js/config/config`,
+)
+
+export class PkgFsBindings implements FileSystem {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -223,5 +223,6 @@
         "infile": "CHANGELOG.md"
       }
     }
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
robertsLando, line 227: missed this I think :)

api/lib/PkgFsBindings.ts
```diff
@@ -0,0 +1,86 @@
...
+
+// Ensures that the Z-Wave JS driver is looking for the right files in the right place
+// when running inside a `pkg` bundle. In this case, it will resolve its embedded
+// configuration dir to the path "/config", but the files reside in "node_modules/@zwave-js/config/config" instead.
+
+const CONFIG_PATH = '/config'
+const CONFIG_PATH_IN_PKG = path.join(
+	__dirname,
+	`node_modules/@zwave-js/config/config`,
+)
```
robertsLando, line 13: I'm wondering if this could be fixed in esbuild bundle by simply monkey patching the file and replacing the place where you set this path?

AlCalzone: I'm afraid that will just lead to us chasing our tail again when I find the need to change the specific code again. Right now it's relatively specific to handle 3 cases: file executed as JS, file executed as TS, run in a context where there is no "main" file:
```ts
/** The absolute path of the embedded configuration directory */
export const configDir = import.meta.url.startsWith("file:")
	? path.join(
		path.dirname(fileURLToPath(import.meta.url)),
		import.meta.url.endsWith("src/utils.ts")
			? ".."
			: "../..",
		"config",
	)
	: import.meta.resolve("/config");
```

robertsLando (approved):
LGTM, thanks 🙏🏼

PR #4196, opened 2025-04-10
docs: add info about mqtt-name

spacemanspiff2007:


docs/usage/setup.md
```diff
@@ -130,6 +130,8 @@ Enable this to use Z-Wave JS UI as only a Control Panel
      - `property`: the value [property](https://zwave-js.github.io/node-zwave-js/#/api/valueid)
      - `propertyKey`: the value [propertyKey](https://zwave-js.github.io/node-zwave-js/#/api/valueid)
 
+     Note that `<mqtt_name>` will also be the numerical id of if the controller node
```
robertsLando, line 133: This is not correct, mqtt name will be the `name` setting set in MQTT settings

robertsLando: @spacemanspiff2007 Are you sure you don't have a `MQTT_NAME` env var set to `1`?

The code is pretty clear: 

https://github.com/zwave-js/zwave-js-ui/blob/0e6251a06dc9ff077c1849dacd0d5ba87852abc1/api/lib/MqttClient.ts#L346-L347

Where prefix is:

https://github.com/zwave-js/zwave-js-ui/blob/0e6251a06dc9ff077c1849dacd0d5ba87852abc1/api/lib/MqttClient.ts#L78-L79

spacemanspiff2007: I use a docker compose file and I am 100% sure that ``MQTT_NAME`` is not set there.
Maybe  ``MQTT_NAME`` is set to `1` by default?

I run an older version (9.24.0), maybe something changed?
I'll try updating to see if the issue still persists ...

robertsLando: > No the less I think the docs would benefit if they show where <mqtt_name> comes from.

I agree with that but what it's wrote there is wrong at least all my instances are not working like that so I'm wondering why you have that `1` there...

spacemanspiff2007: Hmm - as you can see in the above screenshot it indeed sends a success message so something is happening.

grafik

spacemanspiff2007: > so I'm wondering why you have that `1` there

That's the topic I publish to?

robertsLando: > Note that I copy pasted the wrong url to the docs above. The correct one would be [the zwave-apis](https://zwave-js.github.io/zwave-js-ui//#/guide/mqtt?id=z-wave-apis).

It doesn't matter, the mqtt-name is always that. Anyway could you try to connect a shell to container and run `echo $MQTT_NAME` ?

spacemanspiff2007: grafik

robertsLando: Ok so the publish now is ok, about the response to that topic I may know what's the reason, I have to make the check on name more strict (I only check the prefix but since I changed the way I do subscribes it may be that I also get others)

robertsLando: See https://github.com/zwave-js/zwave-js-ui/commit/0e08e3a79e02656f47dc5b058ec08c02bbf6d90f

docs/guide/mqtt.md
```diff
@@ -2,6 +2,8 @@
 
 You have access to almost all of the [Z-Wave JS APIs](https://zwave-js.github.io/node-zwave-js/#/README) (and more) via MQTT.
 
+The placeholder `<mqtt_name>` in the mqtt topic is the name configured in the MQTT settings.
```
spacemanspiff2007, line 5: I'm not sure if here is a good place or under Setup -> MQTT where the other wildcards are described

robertsLando, line 5:
```suggestion
The placeholder `<mqtt_name>` in the mqtt topic is the [sanitized](https://github.com/zwave-js/zwave-js-ui/blob/0e08e3a79e02656f47dc5b058ec08c02bbf6d90f/api/lib/utils.ts#L193)  `name` configured in the MQTT settings.
```

robertsLando (changes requested):

docs/guide/mqtt.md
robertsLando: Maybe it makes more sense to have it in setup mqtt

robertsLando:
It makes no sense there as it will not be present on those topics. I think it could be added on line 104 https://github.com/zwave-js/zwave-js-ui/pull/4196/files#diff-8d25956a6f64976ef0b21e91fa4acee5fcb5582bf778082dbf01d1c7c57c527dR104 right after the name description in mqtt section to explain that it will be used as client id later and will be referenced as `mqtt_name` in topics docs

robertsLando (approved):

PR #4229, opened 2025-05-08
fix: otw update and node firmware update progress hangs on finish

### LINKED ISSUE #3945: finished the install and upgraded the firmware of the controller but it would not go to 100% and not let me get out of the gui, i had to force-refresh the browser
### Checklist

### Deploy method

Home Assistant Add-on

### Z-Wave JS UI version

9.22.0

### ZwaveJS version

13.8.0

### Describe the bug

it finished the install and upgraded the firmware of the controller but it would not go to 100% in the gui and not let me get out of the gui, i had to force-refresh the browser
controller-otw-upgrade-gui-bug

### To Reproduce

upgrade a network based controller OTW (such as tube zwave kit + zac93) and it succeeds in the background but the gui stays at 99.91% and you cannot exit out of it without refreshing the browser

### Expected behavior

it to reach 100% and also let me escape the popup in cases of a failure like this

### Additional context

talked to al calzone in discord and confirmed this was a zui gui bug

robertsLando:
Fixes #3945

### DIFF
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -954,6 +954,16 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		}
 	}
 
+	private clearThrottle(key: string) {
+		const entry = this.throttledFunctions.get(key)
+		if (entry) {
+			if (entry.timeout) {
+				clearTimeout(entry.timeout)
+			}
+			this.throttledFunctions.delete(key)
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


AlCalzone (changes requested):
The status is not reset if I click away the dialog and immediately update again:

grafik

AlCalzone (changes requested):
Also, this needs to include this release: https://github.com/zwave-js/zwave-js/releases/tag/v15.4.2

PR #4243, opened 2025-05-19
feat: support OTW updates via update service

AlCalzone:
This PR adds a "firmware updates" tab to the controller node to perform automatic OTW updates through the firmware update service, like we have for end devices.


src/components/nodes-table/OTWUpdates.vue
```diff
@@ -0,0 +1,252 @@
...
+						v-model="rfRegion"
+					>
+						<template v-slot:prepend>
+							<v-icon>signal_cellular_alt</v-icon>
+						</template>
+					</v-select>
+				</v-row>
+			</v-col>
+
+			<template v-if="filteredUpdates.length > 0 && !node.firmwareUpdate">
```
AlCalzone: Not sure what the `&& !node.firmwareUpdate` does here?

src/components/nodes-table/OTWUpdates.vue
```diff
@@ -0,0 +1,236 @@
...
+						v-model="rfRegion"
+					>
+						<template v-slot:prepend>
+							<v-icon>signal_cellular_alt</v-icon>
+						</template>
+					</v-select>
+				</v-row>
+			</v-col>
+
+			<template v-if="this.fwUpdates.length > 0 && !node.firmwareUpdate">
```
AlCalzone, line 41: Not sure if `node.firmwareUpdate` always has the correct state for the controller node?

api/lib/ZwaveClient.ts
```diff
@@ -3929,6 +3930,24 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		throw new DriverNotReadyError()
 	}
 
+	/**
+	 * Used to trigger an update of controller FW via the Z-Wave JS firmware update service
+	 */
+	async firmwareAutoUpdateOTW(
```
AlCalzone, line 3936: I don't really like this name. Suggestions?

AlCalzone (commented):
I've got a few open questions. Also, how can we redo the check for firmware updates after the OTW update was completed? Right now the view isn't refreshed and still shows the old update.

AlCalzone:
Oh, yeah and this needs a new release of zwave-js

src/components/nodes-table/OTWUpdates.vue
robertsLando: firmwareUpdate has been removed from controller in favor of dedicated events

api/lib/ZwaveClient.ts
robertsLando: controllerFwUpdateOTW ?

src/components/nodes-table/OTWUpdates.vue
robertsLando, line 1: Is this basically a copy paste of the OTAUpdates view? If so I would use that and add a prop instead to handle the different API to call

api/lib/ZwaveClient.ts
AlCalzone: The other method (below) is called `firmwareUpdateOTW`, which also updates the controller.
Maybe they should just be merged and accept both a `FwFile` and a `FirmwareUpdateInfo`? Then decide based on the argument what to do.

src/components/nodes-table/OTWUpdates.vue
AlCalzone: A few changes actually
- Slightly different text
- removed the `target` property from display, since there is no such thing for OTW
- no downgrades
- Also slightly different confirmation text in `updateFirmware`

api/lib/ZwaveClient.ts
robertsLando: yep I think that would be better

src/components/nodes-table/OTWUpdates.vue
robertsLando: but as staited below this is only valid for nodes, not controller

AlCalzone:
Let me test later today, okay?

AlCalzone:
There is no `target` for OTW updates, we should not show that:
grafik

Otherwise this is good.

robertsLando:
@AlCalzone code is like:

```
<v-list-item-title
										v-if="f.target !== undefined"
										>Target:
										{{ f.target }}</v-list-item-title
									>
```

I'm wondering why the target property is set?

AlCalzone:
Hmm I'll have to check. Probably because I'm not distinguishing between controller and end device updates. Can we just hide it anyways?

robertsLando:
@zwave-js-bot fix lint

PR #4260, opened 2025-06-16
feat: load and display saved Zniffer captures

AlCalzone:
This is the initial work for being able to load Zniffer traces and showing them in the UI.

Depends on https://github.com/zwave-js/zwave-js/pull/7889, which is currently not finished.

### DIFF
--- a/src/views/Zniffer.vue
+++ b/src/views/Zniffer.vue
@@ -425,7 +425,7 @@ export default {
 					color: 'success',
 					tooltip: 'Start Zniffer',
 					action: this.startZniffer,
-					disabled: this.znifferState?.started,
+					disabled: this.viewState === 'recording',
 				},
 				{
 					id: 'stop',
@@ -433,7 +433,7 @@ export default {
 					color: 'error',
 					tooltip: 'Stop Zniffer',
 					action: this.stopZniffer,
-					disabled: !this.znifferState?.started,
+					disabled: this.viewState !== 'recording',
 				},
 				{
 					id: 'clear',
@@ -441,7 +441,8 @@ export default {
 					color: 'warning',
 					tooltip: 'Clear Zniffer',
 					action: this.clearFrames,
-					disabled: !this.frames.length,
+					disabled:
+						this.frames.length === 0 || this.viewState === 'loaded',
 				},
 				{
 					id: 'load',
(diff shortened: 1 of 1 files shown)


api/app.ts
```diff
@@ -758,6 +758,11 @@ function setupSocket(server: HttpServer) {
 					case 'saveCaptureToFile':
 						res = await zniffer.saveCaptureToFile()
 						break
+					case 'loadCaptureFromBuffer': {
+						const buffer = Buffer.from(data.buffer)
```
AlCalzone, line 762: Does this make sense?

src/views/Zniffer.vue
```diff
@@ -1025,6 +1042,68 @@ export default {
...
+				const reader = new FileReader()
+
+				reader.onload = async (e) => {
+					try {
+						const byteArray = new Uint8Array(e.target.result)
+
+						const response = await this.sendAction(
+							{
+								apiName: 'loadCaptureFromBuffer',
+								buffer: Array.from(byteArray),
```
AlCalzone, line 1076: Likewise, is this the way to upload a binary file?

src/views/Zniffer.vue
```diff
@@ -1025,6 +1042,68 @@ export default {
 				)
 			}
 		},
+		loadCapture() {
```
robertsLando, line 1045: Instead of using an hidden input I have an util method on App.vue that you can call using this.app.importFile, check this has instancemixins (so this.app is defined)

src/views/Zniffer.vue
robertsLando: See comment above

AlCalzone:
This is how it looks now:
grafik


I think we need to select a different color for the load/save icons, as they are too similar.

Also some of the state transitions, e.g. capturing -> load file from disk, need to be reworked. Right now, they do nothing.

AlCalzone:
eh I was hoping to be able to test with the alpha release, but npm shits on the parade again.

robertsLando:
> Also some of the state transitions, e.g. capturing -> load file from disk, need to be reworked. Right now, they do nothing.

What do you mean?

AlCalzone:
For example loading a trace should stop the existing one and starting a new one should clear the loaded one.

Haven't done that yet.

robertsLando (approved):

PR #4263, opened 2025-06-20
fix(ui): typo in restart prompt

The00Dustin:
While not necessarily incorrect when expanded (the zwave-js user interface), "the ZUI" sounds poorly translated, and if ZUI is considered as a proper noun (actually expanding to Z-Wave JS UI), then the existing wording is grammatically incorrect (think "Are you sure you want to restart the Windows?").


robertsLando (approved):
Thanks!

PR #4266, opened 2025-06-28
feat(ui): support system preference color scheme

joneshf:
## Context

We migrate from only supporting an explicit dark or light color scheme to also supporting whatever the current system preference is for the color scheme.

This allows folks to change their color scheme at the operating system level (whether they do it automatically or manually) and have that setting be reflected in Z-Wave JS UI. This also means that different device preferences should follow what the device has configured (e.g. a desktop might always be in dark mode, while a mobile device changes based on the time of day).

This is my first time contributing to `zwave-js-ui`, and I have limited experience with Vue. So apologies if I've done something wrong, or a bit off. Let me know if you have any suggestions/alternatives to the implementation.

Fixes https://github.com/zwave-js/zwave-js-ui/issues/3607

## Screenshots

I've attached some screenshots. I'm not very attached to the copy or design here. So I'm down to make whatever changes you'd prefer (so long as I still understand it 😅).

### Login

<details>
<summary>Login - dark color scheme</summary>
z-wave-js-ui-login-dark
</details>

<details>
<summary>Login - light color scheme</summary>
z-wave-js-ui-login-light
</details>

<details>
<summary>Login - system color scheme</summary>
z-wave-js-ui-login-system
</details>

<details>
<summary>Login - system color scheme - dark preference</summary>
z-wave-js-ui-login-system-dark
</details>

<details>
<summary>Login - system color scheme - light preference</summary>
z-wave-js-ui-login-system-light
</details>

### Settings

<details>
<summary>Settings - dark color scheme</summary>
z-wave-js-ui-settings-dark
</details>

<details>
<summary>Settings - light color scheme</summary>
z-wave-js-ui-settings-light
</details>

<details>
<summary>Settings - system color scheme</summary>
z-wave-js-ui-settings-system
</details>

<details>
<summary>Settings - system color scheme - dark preference</summary>
z-wave-js-ui-settings-system-dark
</details>

<details>
<summary>Settings - system color scheme - light preference</summary>
z-wave-js-ui-settings-system-light
</details>

### DIFF
--- a/src/lib/colorScheme.js
+++ b/src/lib/colorScheme.js
@@ -1,0 +1,70 @@
+/**
+ * @typedef {('dark' | 'light' | 'system')} ColorScheme
+ */
+
+/**
+ * @typedef {boolean} DarkMode
+ */
+
+/**
+ * The valid {@link ColorScheme}s that can be used.
+ */
+export const colorSchemes = ['dark', 'light', 'system']
+
+/**
+ * Maps a {@link ColorScheme} to its corresponding {@link DarkMode} value.
+ * Will lookup the preferred value when {@link colorScheme} is `'system'`
+ *
+ * @param {ColorScheme} colorScheme
+ * @returns {DarkMode}
+ */
+export const colorSchemeToDarkMode = (colorScheme) => {
+	switch (colorScheme) {
+		case 'dark':
+			return true
+
+		case 'light':
+			return false
+
+		case 'system':
+			return prefersColorSchemeDark.matches
+	}
+}
+
+/**
+ * Loads the {@link ColorScheme} from the {@link settings}.
+ *
+ * If {@link ColorScheme} is not in the {@link settings},
+ * this will fallback to migrating from {@link DarkMode}.
+ *
+ * @param {Settings} settings
+ * @returns {ColorScheme}
+ */
+export const loadColorScheme = (settings) => {
+	const darkMode = settings.load('darkMode', undefined)
+	let defaultColorScheme
+	switch (darkMode) {
+		case false:
+			defaultColorScheme = 'light'
+			break
+
+		case true:
+			defaultColorScheme = 'dark'
+			break
+
+		case undefined:
+			defaultColorScheme = 'system'
+			break
+	}
+
+	return settings.load('colorScheme', defaultColorScheme)
+}
+
+/**
+ * Tracks whether the current color scheme should be dark.
+ *
+ * @type {MediaQueryList}
+ */
+export const prefersColorSchemeDark = window.matchMedia(
+	'(prefers-color-scheme: dark)',
+)
--- a/src/lib/colorScheme.js
+++ b/src/lib/colorScheme.js
@@ -41,14 +41,14 @@ export const colorSchemeToDarkMode = (colorScheme) => {
  * @returns {ColorScheme}
  */
 export const loadColorScheme = (settings) => {
-	const darkMode = settings.load('darkMode', undefined)
+	const darkMode = settings.load('dark', undefined)
 	let defaultColorScheme
 	switch (darkMode) {
-		case false:
+		case 'false':
 			defaultColorScheme = 'light'
 			break
 
-		case true:
+		case 'true':
 			defaultColorScheme = 'dark'
 			break
 
--- a/src/views/Login.vue
+++ b/src/views/Login.vue
@@ -59,7 +59,6 @@
 						<color-scheme
 							prepend-icon="null"
 							hide-details
-							v-model="internalColorScheme"
 						></color-scheme>
 					</v-form>
 				</v-card-text>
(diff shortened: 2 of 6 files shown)


joneshf:
Sounds good, I'll give that a whirl when I get a minute. I'm not quite sure I see where to split the code at the moment (it'll probably become clearer once I start touching the code), but I'll ping if I get stuck on anything.

robertsLando:
@joneshf yess no worries, thanks 🙏🏼

joneshf:
So I made a component, but I'm not sure how we'd move the logic from `src/lib/colorScheme.js` to `src/components/custom/ColorScheme.vue`. For example, because the `ColorScheme` component is only used on two views (Login and Settings) I'm not sure we can move the event listener into the `ColorScheme` component, as the listener wouldn't be setup if someone was on a different view, so changes to the system color scheme (like from the os automatically doing it or a person manually doing it) wouldn't be picked up in Z-Wave JS UI.

Was this what you were looking for, or was there more we should be able to do? If I've set things up not the right way, lemme know and I can swap them around to align with what you're thinking.

src/views/Login.vue
```diff
@@ -56,13 +56,11 @@
...
-						<v-checkbox
+						<color-scheme
 							prepend-icon="null"
 							hide-details
-							persistent-hint
-							label="Dark mode"
-							v-model="internalDarkMode"
-						></v-checkbox>
+							v-model="internalColorScheme"
+						></color-scheme>
```
robertsLando, line 63: there is not need to specify v-model here, the logic in this case will be handled inside the component as it works with values from the store :)

src/views/Login.vue
```diff
@@ -123,14 +124,14 @@ export default {
...
-		internalDarkMode: {
+		internalColorScheme: {
 			get() {
-				return this.darkMode
+				return this.colorScheme
 			},
 			set(value) {
-				this.setDarkMode(value)
+				this.setColorScheme(value)
 			},
```
robertsLando, line 135: so this will not be needed then :)

robertsLando (changes requested):

robertsLando:
@joneshf What you did is correct except for one little thing I mentioned in my review above,

>  but I'm not sure how we'd move the logic from src/lib/colorScheme.js to src/components/custom/ColorScheme.vue

Of course that logic should stay where it is now, I only meant to create the component for the select to reduce repeated code :)

joneshf:
Ah! Okay. I think I get what you're suggesting. The `ColorScheme` component should use the `baseStore` directly, instead of relying on its parent component doing the book keeping. I can make that update!

src/lib/colorScheme.js
```diff
@@ -41,14 +41,14 @@ export const colorSchemeToDarkMode = (colorScheme) => {
...
+	const darkMode = settings.load('dark', undefined)
 	let defaultColorScheme
 	switch (darkMode) {
-		case false:
+		case 'false':
 			defaultColorScheme = 'light'
 			break
 
-		case true:
+		case 'true':
```
joneshf: These two bugs kind of caught me off-guard. I tried to write some tests for this stuff (so it doesn't catch the next person), but I couldn't really find a good way to do it:
- If I add a `src/lib/colorScheme.test.js` and update the [`test:ui` script](https://github.com/zwave-js/zwave-js-ui/blob/86fd3f7e0018e90d6c967d3f968ea868ec7126d6/package.json#L43) to include `src/lib`, some of the other files in `src/lib` cause failures (`src/lib/axios-progress-bar.js`, I believe). That seemed like too big of a thing to try and fix in this PR.
- I could move `src/lib/colorScheme.js` to `src/modules/ColorScheme.js` and write a test in `src/modules/ColorScheme.test.js`, but I'm not actually sure if that aligns with what a "module" is in this repository. The other things in `src/modules` are all classes that have non-trivial behavior, whereas this would be a couple of functions. Turning it into a class for the sake of looking like a "module" seemed not quite right.

    If it's fine for a "module" to be some helper functions, then I could move this stuff to `src/modules`.
- I thought about making a different PR to get `src/lib` able to have tests, but would rather talk about that before making that change. It might not align with what you want. Or maybe you're fine with this stuff having less coverage because you get confidence in some other way.

@robertsLando What are your thoughts on testing this new stuff?

joneshf (commented):
Updated it to use the `baseStore` properly. Mind giving it another look whenever you get a minute?

robertsLando (approved):
Looks good!

src/lib/colorScheme.js
robertsLando: > @robertsLando What are your thoughts on testing this new stuff?
This could be done in another PR if you want. Just change the test:UI command in `mocha -r @babel/register src/**/*.test.js` so it will only look for files ending with test.js.

**UPDATE**: I already pushed this change on master so feel free to open a new PR with the test (no need to move it to modules)

joneshf: Ah, yeah that makes sense. Thanks for that!

joneshf:
Thanks for the review/merge!

Just tried out the 10.8.0 release. Works for me. Thanks again!

robertsLando:
@joneshf Thanks to you for the contribution 🙏🏼

PR #4277, opened 2025-07-09
feat: make RF region setting mandatory, enable auto-powerlevel by default

AlCalzone:
This is the first step to help users with configuring controllers for their region a bit easier than right now.

The RF region setting is now mandatory, so there are no surprises when switching controllers or setting up a new one that defaults to a different region.
Auto-powerlevel is now enabled by default, because nobody should have to deal with powerlevels anyways.

### DIFF
--- a/src/stores/base.js
+++ b/src/stores/base.js
@@ -37,9 +37,9 @@ const useBaseStore = defineStore('base', {
 			logLevel: 'debug',
 			rf: {
 				region: undefined,
-				maxLongRangePowerlevel: undefined,
+				maxLongRangePowerlevel: 'auto',
 				txPower: {
-					powerlevel: undefined,
+					powerlevel: 'auto',
 					measured0dBm: undefined,
 				},
 			},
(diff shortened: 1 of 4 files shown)


src/stores/base.js
```diff
@@ -37,9 +37,9 @@ const useBaseStore = defineStore('base', {
...
 			rf: {
 				region: undefined,
-				maxLongRangePowerlevel: undefined,
+				maxLongRangePowerlevel: 'auto',
 				txPower: {
-					powerlevel: undefined,
+					powerlevel: 'auto',
 					measured0dBm: undefined,
 				},
 			},
```
AlCalzone, line 45: @robertsLando is this enough to ensure users who haven't set this before now get auto-powerlevel enabled?

robertsLando: Unfortunately nope: https://github.com/zwave-js/zwave-js-ui/blob/996e81a12666a7df45034009fafec029a065ad9e/src/stores/base.js#L508

robertsLando: You could set it to default in `initSettings` by checking if it is undefined?

robertsLando: The alternative is to use a merge module instead of object assign for this

PR #4278, opened 2025-07-09
feat: use default RF region from settings for OTA update check

AlCalzone:
This is a followup to #4277 and replaces the region selector on the OTA page with a hint to configure the region in the settings.

When the current controller region is not known, the region from the settings is now used instead.

Unfortunately, I can't test that as I don't have a controller that does not support changing the region.


robertsLando (approved):

robertsLando:
@AlCalzone can I merge this?

PR #4281, opened 2025-07-10
feat: add auto power levels option for RF configuration in settings

robertsLando:
This would make it impossible for users to unset this

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,7 +60,7 @@
         "vuetify": "^2.7.2",
         "winston": "^3.13.0",
         "winston-daily-rotate-file": "^5.0.0",
-        "zwave-js": "^15.8.0",
+        "zwave-js": "^15.9.0",
         "zxing-wasm": "^1.2.15"
       },
       "bin": {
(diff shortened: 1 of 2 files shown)


AlCalzone (dismissed):
Haven't tested, but looks good to me

robertsLando:
@AlCalzone Should we force autoPowerLevel setting to true if undefined? And if yes should we force it on frontend (requires then user to press save) or on backend (automatically picked up) ?

AlCalzone:
Hmm let's do it on the backend and see if someone complains.

robertsLando:
@AlCalzone ok but this will set both to auto even on users that specified some custom ones, a safer approach could be to consider the new option to true only if there is no custom powerlevel option set...

AlCalzone:
Right, that would make more sense.

AlCalzone:
By the way, when I check out this PR, I get two options:
grafik

AlCalzone:
> But this will set both to auto even on users that specified some custom ones, a safer approach could be to consider the new option to true only if there is no custom powerlevel option set...

Can you do that? I fixed the other issue I noticed.

PR #4290, opened 2025-07-16
feat: move to Vue/Vuetify 3

robertsLando:


robertsLando:
TODOs:

- [x] Nodes Manager dialog is broken
- [x] Tests are broken
- [x] Column filters are not working (search)
- [x] Checkboxes on z-wave graph not working
- [x] z-wave graph not filling available space
- [x] Filters not persisted on table
- [x] Refactor v-tooltips using `v-tooltip` directive where possible
- [x] Refactor v-list-item using :title and :subtitle where possible
- [x] Fix node details view in zwave graph (nothing is shown when clicking on a node)
- [x] Verify node healtcheck and link reliability dialogs
- [x] User codes table is broken
- [x] Replace `var(--` in css with `v-bind`
- [ ] Refactor color scheme to use new vuetify 3 `system` theme

PR #4375, opened 2025-09-18
chore: fix constant page reloads during development

AlCalzone:


robertsLando (approved):

robertsLando:
Another possible fix is to load components a-la-cart instead of loading them all btw let's go for this one

PR #4391, opened 2025-09-25
fix: refresh normal powerlevel after region change in auto mode

AlCalzone:
Fixes https://github.com/zwave-js/zwave-js-ui/issues/4377


robertsLando:
@AlCalzone you can use this to share code between api/ui now: https://github.com/zwave-js/zwave-js-ui/pull/4399. 

I suggest to create a shared.ts file in API where we can put untils functions like that

PR #4434, opened 2025-10-26
feat(ui): allow removing multiple failed nodes

maxberger:
This patch allows to call removeFailedNode on multiple selected nodes. Replace is intentionally not added, as it would not work well with multiple nodes.


robertsLando (dismissed):

robertsLando:
@AlCalzone opinions on this?

robertsLando:
@maxberger please could you run `npm run lint-fix` ?

maxberger:
> @maxberger please could you run `npm run lint-fix` ?

done.

robertsLando:
@AlCalzone this is a separeted list of actions that can be called only when there are multiple nodes selected. Of course you can also call it for a single node as before

PR #4463, opened 2025-11-18
fix: improve automatic firmware update checks

robertsLando:
When automatic fw updates is not disabled, updates are automatically checked when:

- All nodes are ready
- Daily on a random time
- After a node interview

### DIFF
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -6244,20 +6244,6 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 				}.\n   Result: ${JSON.stringify(result)}.`,
 			)
 
-			if (result.reInterview) {
-				this.logNode(zwaveNode, 'info', 'Will be re-interviewed')
-			} else {
-				// Query for new firmware updates immediately if not re-interviewing
-				// (unless automatic firmware update checks are disabled)
-				this._checkNodeFirmwareUpdates(zwaveNode.id).catch((error) => {
-					this.logNode(
-						zwaveNode,
-						'error',
-						`Failed to check firmware updates after update: ${error.message}`,
-					)
-				})
-			}
-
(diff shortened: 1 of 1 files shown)


robertsLando:
> After an fw update

@AlCalzone maybe I can remove this? A node is re-interviewed automatically after an update? This is also used to clean up available updates on node instance

api/lib/ZwaveClient.ts
```diff
@@ -6257,25 +6252,16 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
-								zwaveNode,
-								'error',
-								`Failed to check firmware updates after update: ${error.message}`,
-							)
-						},
+				this._checkNodeFirmwareUpdates(zwaveNode.id).catch((error) => {
+					this.logNode(
+						zwaveNode,
+						'error',
+						`Failed to check firmware updates after update: ${error.message}`,
```
robertsLando: here I immediately check for node fw updates when result doesn't have `reInterview` set

api/lib/ZwaveClient.ts
```diff
@@ -5781,20 +5786,13 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
-						zwaveNode,
-						'error',
-						`Failed to check firmware updates after interview: ${error.message}`,
-					)
-				},
+		this._checkNodeFirmwareUpdates(zwaveNode.id).catch((error) => {
+			this.logNode(
+				zwaveNode,
+				'error',
+				`Failed to check firmware updates after interview: ${error.message}`,
```
robertsLando: on interview complete

AlCalzone:
> @AlCalzone maybe I can remove this? A node is re-interviewed automatically after an update? This is also used to clean up available updates on node instance

Yeah. Immediately after a firmware update does not make sense, since at that point the information is outdated.

PR #4488, opened 2025-11-30
feat(ui): make header/footer fixed on control panel table

### LINKED ISSUE #3603: Add pagination buttons to top of control panel screen
I have 80 nodes in zwave-js-ui. 75% of the time I need to hit page forward at least twice to get to the node I want to look at.

Problem is when the pagination buttons are only at the bottom, the placement jumps all over the place vertically.  Let's say I want to get to node 60. Scroll down, hit page forward. Next page has different vertical spacing (different number of lines for descriptions, locations, etc.) so now I have to play where's waldo for the forward button.  OK, found it, scroll, hit next page.  Still not there, where's waldo now?  Found it, scroll, page forward.  Oh JFC where's waldo NOW? 

At the top, it'd be click-click-click hey there's my node!

Please add pagination buttons to the top as well.  Thanks!

robertsLando:
Fixes #3603

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -9,7 +9,7 @@
     "requireBranch": "master"
   },
   "hooks": {
-    "before:init": [
+    "after:bump": [
       "npm run build"
     ]
   },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,42 @@
 # Changelog
 
+## [11.9.1](https://github.com/zwave-js/zwave-js-ui/compare/v11.9.0...v11.9.1) (2025-12-20)
+
+### ✨ Features
+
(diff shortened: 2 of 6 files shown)


robertsLando:
@AlCalzone Would like your opinion on this, not sure if the view looks too "trimmed" now

robertsLando:
cc @AlCalzone

zackbcom:
I tried it on my macbook and it's a bit too squished when opening the device page
Maybe it could be an option to enable, or only enabled with showing on an. XXL screen.
Maybe locking the table header to the top of the screen and adding pagination to the top might help?

robertsLando:
@zackbcom You mean fixing both table header and pagination to the top then? 

P.S: thanks for the Christmas present, received it few days ago 🙏🏼❤️

PR #4536, opened 2026-02-07
feat: add support for the renewed plugin of prometheus exporter

varet80:
I have fixed my plugin and added new functionality. Please check and accept my new plugin into the tool.

It exports to Prometheus, using a separate httpserver (on port 9001) as typically is served by the metrics exporters/plugins

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8790,6 +8790,27 @@
         "node": ">=0.10.0"
       }
     },
+    "node_modules/cli-truncate/node_modules/ansi-regex": {
+      "version": "2.1.1",
+      "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
+      "integrity": "sha512-TIGnTpdo+E3+pCyAluZvtED5p5wCqLdezCyhPZzKPcxvFplEt4i+W7OONCKgeZFT3+y5NZZfOOS/Bdcanm1MYA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=0.10.0"
+      }
+    },
+    "node_modules/cli-truncate/node_modules/is-fullwidth-code-point": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/is-fullwidth-code-point/-/is-fullwidth-code-point-1.0.0.tgz",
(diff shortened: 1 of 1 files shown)


varet80:
It does not allow me to change the PR name. it should be `feat:`

varet80:
@robertsLando Yes I renamed it and moved it to new org. I have no access to that old package anymore. and it is outdated.

The new replaces the old.

robertsLando:
Thanks for the contribution! However, I'd like to avoid adding plugin packages as direct dependencies of zwave-js-ui to keep the dependency tree minimal. Plugins should be installed and managed independently by users.

The plugin system already supports loading plugins from **any path** — not just npm packages in `node_modules`. Here's a step-by-step guide for Docker users who want to use a custom plugin like yours:

### Using a custom plugin with Docker

**1. Create a plugins directory on your host**

```bash
mkdir -p /opt/zwave-plugins
```

**2. Install the plugin inside that directory**

```bash
cd /opt/zwave-plugins
npm install @ongit/zwavejsui-prom-exporter
```

This will create a `node_modules` folder inside `/opt/zwave-plugins` with the plugin and its dependencies.

**3. Mount the directory as a volume in your Docker container**

Add a volume mapping to your `docker-compose.yml`:

```yaml
services:
  zwave-js-ui:
    # ... your existing config ...
    volumes:
      - zwave-store:/usr/src/app/store
      - /opt/zwave-plugins:/usr/src/app/plugins  # <-- add this
```

Or if using `docker run`:

```bash
docker run ... -v /opt/zwave-plugins:/usr/src/app/plugins ...
```

**4. Configure the plugin path in Z-Wave JS UI**

Go to **Settings** → **General** → **Plugins** and type the full path to the plugin inside the container:

```
/usr/src/app/plugins/node_modules/@ongit/zwavejsui-prom-exporter
```

Press **Enter** to add it, then click **SAVE**.

**5. Restart the gateway**

The plugin will be loaded on the next gateway start. Check the logs to confirm it loaded successfully.

---

This approach keeps the plugin outside of the zwave-js-ui image, making it easy to update independently and avoiding dependency bloat in the main project. Could you update the plugin's README with these instructions instead?

varet80:
That means all other plugins will soon be removed? 

Like the other plugin https://github.com/zwave-js/zwave-js-ui/blob/de3eeaaade20d1df46c2ef18f79d18b90eea4860/package.json#L81

PR #4555, opened 2026-02-20
feat: split socket communication into channels

AlCalzone:
**Currently**, the server uses `io.emit()` / `SocketServer.emit()` to broadcast **every event to every connected client**. Clients that register a `socket.on()` handler only filter locally — the message has already been serialized, transmitted over the wire, and deserialized before being silently dropped. For busy networks with frequent `VALUE_UPDATED`, `STATISTICS`, `DEBUG`, or `ZNIFFER_FRAME` events, this wastes significant bandwidth and CPU on both server and client.

**With this PR**, Socket.IO events are routed through room-based channels. The server emits to rooms (e.g. `io.to('values').emit(...)`) so messages are only sent to clients that have explicitly subscribed. Clients subscribe to channels they need via a `SUBSCRIBE` event, and the UI dynamically manages subscriptions based on which views are open.

This lets external clients (scripts, integrations) avoid high-traffic streams by only subscribing to what they care about, or issue API commands without receiving any unwanted messages.

### Available channels

| Channel | Events |
| --- | --- |
| `controller` | `CONTROLLER_CMD`, `CONNECTED`, `INFO` |
| `nodes` | `NODE_FOUND`, `NODE_ADDED`, `NODE_REMOVED`, `NODE_UPDATED`, ... |
| `values` | `VALUE_UPDATED`, `VALUE_REMOVED`, `METADATA_UPDATED` |
| `statistics` | `STATISTICS` |
| `firmware` | `OTW_FIRMWARE_UPDATE` |
| `debug` | `DEBUG` |
| `zniffer` | `ZNIFFER_FRAME`, `ZNIFFER_STATE` |
| `rebuild` | `REBUILD_ROUTES_PROGRESS` |
| `diagnostics` | `HEALTH_CHECK_PROGRESS`, `LINK_RELIABILITY` |

> **UNTESTED, still working on this...**

### DIFF
--- a/src/views/Debug.vue
+++ b/src/views/Debug.vue
@@ -53,8 +53,6 @@
 	</v-container>
 </template>
 <script>
-import { socketEvents } from '@server/lib/SocketEvents'
-
 import { AnsiUp } from 'ansi_up'
 import { mapActions } from 'pinia'
 import useBaseStore from '../stores/base.js'
@@ -189,7 +187,7 @@ export default {
 
 		this.subscribeChannels(['debug'])
 
-		this.socket.on(socketEvents.debug, (data) => {
+		this.bindEvent('debug', (data) => {
 			if (this.debugActive) {
 				data = ansiUp.ansi_to_html(data)
 				data = data.replace(/\n/g, '</br>')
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -1320,6 +1320,17 @@ export default {
...
+
+				this.socket.emit('SUBSCRIBE', {
+					channels: [
+						'controller',
+						'nodes',
+						'values',
+						'statistics',
+						'firmware',
+					],
+				})
```
robertsLando: same

robertsLando (changes requested):

robertsLando (dismissed):

PR #4584, opened 2026-03-07
docs: add ioBroker link in README.md

mcm1957:
This PR adds ioBroker to the list of integrations


mcm1957:
Change betrifft DOKUMENTATION auch wenn das der title nicht mag ...

mcm1957:
> 💡 The pull request title does not match our guidelines 💡

Please explain whats the problem dear bot:

Do you like
feat: Add ioBroker link in README.md
better ?

> We follow conventional commits, check out the [description and examples]
(https://www.conventionalcommits.org/en/v1.0.0/). This means that the PR title must...

> start with the type of change, e.g. feat:, or fix:

The title start fith feat:

> optionally specify a scope, for example we use config for changes to config files: feat(config):

You did not like this - do you remember?

> contain a description that explains what is being changed, e.g. add a parameter, correct a typo, ...

This is present:

> and ideally explains what the change affects, e.g. the device (manufacturer, device label, firmware, ...)

If you need an explanation I will aks copilot to generate some text for this change

> The description should use uppercase and lowercase normally, except for the first character which must be lowercase.

I'll let copilot know

mcm1957:
Heureka - the bot is satisfies
lets go an get a BEER

mcm1957:
Link added. Sorry did not notice that there are two locations.

PR #4591, opened 2026-03-13
build: switch to Vite 8

AlCalzone:
There are currently two vite plugins that are not fully compatible yet:
- https://github.com/davidmyersdev/vite-plugin-node-polyfills/issues/142
- https://github.com/vite-pwa/vite-plugin-pwa/issues/912

At first glance, I do not see any issues at runtime though.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -92,7 +92,7 @@
         "@types/sinon-chai": "^3.2.12",
         "@typescript-eslint/eslint-plugin": "^8.45.0",
         "@typescript-eslint/parser": "^8.45.0",
-        "@vitejs/plugin-vue": "^6.0.2",
+        "@vitejs/plugin-vue": "^6.0.5",
         "@yao-pkg/pkg": "^6.10.1",
         "c8": "^10.1.2",
         "chai": "^4.4.1",
@@ -100,7 +100,7 @@
         "conventional-changelog-cli": "^5.0.0",
         "docsify": "^4.13.1",
         "docsify-cli": "^4.4.4",
-        "esbuild": "^0.25.1",
+        "esbuild": "^0.27.0",
         "eslint": "^9.37.0",
         "eslint-config-prettier": "^10.1.8",
         "eslint-plugin-babel": "^5.3.1",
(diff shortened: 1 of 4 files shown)


robertsLando:
The only super weird thing I noticed is that the built version shows weird icons, seems like them are bigger and bolder ??

[Kooha-2026-03-17-08-26-08.webm](https://github.com/user-attachments/assets/fd725e4e-abbc-47e2-ab3e-f9f36d493903)

robertsLando:
The icon font rendering issue (icons appearing bigger/bolder in production builds) was caused by Vite 8's Rolldown bundler reordering CSS chunks, reversing the cascade between dev and production modes.

Reported upstream: https://github.com/vitejs/vite/issues/21903

Workaround applied in b0607f84: wrap the `material-design-icons-iconfont` CSS import in a CSS `@layer` so Vuetify's component styles always take precedence regardless of source order.

robertsLando:
Follow-up on the Vite patch releases:

I bumped Vite to `8.0.16` in `7110b2f9` and retested the material icons production CSS ordering.

Result: latest Vite 8 does **not** remove the need for the `@layer` workaround in `src/assets/css/material-icons.css`.

Checks:

- Vite `8.0.0`, direct `material-design-icons-iconfont` import: production build links `VIcon-*.css` before `index-*.css`. `VIcon-*.css` contains `.v-icon` / `.v-icon--size-default`; `index-*.css` contains `.material-icons`, so `.material-icons` still wins later in the cascade.
- Vite `8.0.16`, direct import: same ordering remained. `.material-icons` still lands in `index-*.css` after `VIcon-*.css`.
- The conflicting declarations are still the original issue: Vuetify has `display: inline-flex` and icon size rules, while `.material-icons` later sets `display: inline-block` and `font-size: 24px`.
- Vite `8.0.16` with the current `@layer(icon-fonts)` wrapper builds successfully, and `.material-icons` remains layered, so unlayered Vuetify styles win regardless of stylesheet link order.

Builds run:

- `npm run build:server`
- `npm run build:ui` with Vite `8.0.0` and no layer
- `npm run build:ui` with Vite `8.0.16` and no layer
- `npm run build:ui` with Vite `8.0.16` plus the current layer workaround

Conclusion: keep the CSS layer patch.

PR #4597, opened 2026-03-22
fix(docs): clarify default value of BACKUPS_DIR

emlove:
See https://github.com/zwave-js/zwave-js-ui/blob/61b6b4e6083030d7883e330133035206beda0d51/api/config/app.ts#L14-L15


robertsLando:
Thanks @emlove

PR #4611, opened 2026-04-09
feat(gateway): discover stateless notification events

hlehoux2021:
Summary

add stateless Notification CC discovery branch in Gateway discover logic
keep Basic and Notification handled together
gate stateless discovery with cmdClass === CommandClasses.Notification and valueId.stateless
use mdi:alarm-light for stateless notification sensors
Why

ensure stateless notification events are discovered as Home Assistant entities without broadening Basic event behavior
Testing

manually reviewed discovery logic in Gateway
no full test run included in this change

### DIFF
--- a/api/lib/Gateway.ts
+++ b/api/lib/Gateway.ts
@@ -1515,8 +1515,7 @@ export default class Gateway {
 								break
 							case 'Water Alarm':
 								cfg = this._getBinarySensorConfig(
-									Constants.deviceClass.sensor_binary
-										.MOISTURE,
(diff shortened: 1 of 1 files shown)


hlehoux2021:
This PR solves a problem with mqtt discovery for devices that send event notifications like 
the Zipato Keypad.
the device sends e.g.

prefix/node_id/113/0/Access_Control/Keypad_unlock_operation
prefix/node_id/113/0/Access_Control/Keypad_lock_operation

but nothing is published for discovery in the homeassistant topic.

i added the case for stateless event notifications in gateway.ts discoverValue()

PR #4623, opened 2026-04-28
build: refactor TypeScript runtime syntax

AlCalzone:
[Node.js 26 is removing the `--experimental-transform-types` option](https://github.com/nodejs/node/pull/62526), effectively killing support for runtime TypeScript syntax like `enum`s. We use them in a couple of places - this PR rewrites them to a combination of objects and types with effectively the same behavior.


api/lib/Gateway.ts
```diff
@@ -142,17 +142,11 @@ export function closeWatchers() {
...
-}
+export const GatewayType = GATEWAY_TYPE
+export type GatewayType = (typeof GATEWAY_TYPE)[keyof typeof GATEWAY_TYPE]
 
-export enum PayloadType {
-	JSON_TIME_VALUE,
-	VALUEID,
-	RAW,
-}
+export const PayloadType = PAYLOAD_TYPE
```
robertsLando, line 148: Heads-up: this is a silent rename of the public-ish surface. The old enum exposed `PayloadType.JSON_TIME_VALUE` (= `0`); the alias now exposes only `PayloadType.TIME_VALUE` because `PAYLOAD_TYPE` uses the shorter key. I grepped the repo and there are no consumers of `JSON_TIME_VALUE` today, so nothing breaks — but anyone reading external docs/snippets referencing the old member name will hit `undefined` with no compile error (numeric union type swallows it).

Two options if you want to avoid that footgun:
1. Rename `PAYLOAD_TYPE.TIME_VALUE` → `JSON_TIME_VALUE` so the exported member name stays stable, or
2. Add a one-line comment here noting the rename for future readers.

Non-blocking either way.

robertsLando: Went with option 1 — renamed `PAYLOAD_TYPE.TIME_VALUE` → `JSON_TIME_VALUE` so the exported member name stays stable. Verified there are no internal call sites for `TIME_VALUE`, and the UI/store bind `payloadType` to numeric values (not key names), so no user-facing impact.

robertsLando (approved):

PR #4630, opened 2026-05-04
chore(deps): move frontend-only deps out of runtime deps

PolarBearEs:
# Summary

This PR moves frontend-only packages from `dependencies` to `devDependencies` so they are no longer kept in the runtime Docker image.

The current image runs the server and serves the built `dist/` frontend, so browser-only libraries used exclusively during frontend build and in-browser execution do not need to remain in production runtime dependencies.

# What Changed

- Moved frontend-only packages from `dependencies` to `devDependencies`
- Updated `package-lock.json` to reflect the new dependency classification

Packages moved include UI and browser-side libraries such as:

- `vuetify`
- `vue`
- `vue-router`
- `pinia`
- `vis-network`
- `vis-data`
- `prismjs`
- `axios`
- `socket.io-client`
- and related frontend-only packages

# Why

While reviewing the Docker image contents, `vuetify` stood out as a large dependency, around `60 MB` installed, even though it is only used by the built frontend and is not required by the runtime server.

From there, the same pattern was identified across a broader set of frontend-only libraries: they were useful for building and running the UI in the browser, but not needed in the production container that serves the already-built assets.

# Result

With the current Docker build path, the image size changed from:

- `507 MB` before
- `292 MB` after

That is a reduction of about `215 MiB` (`42%`).


robertsLando (approved):
Thanks for this PR @PolarBearEs! Dunno why I never catched that

PolarBearEs:
@robertsLando Honestly, it’s pretty easy to miss unless you’re actively looking for it or watching disk usage.

PR #4638, opened 2026-05-15
feat: implement credential management UI

### LINKED ISSUE #175: Get full Z-Wave credential management support via Z-Wave JS UI
### Problem statement

The User Credential Command Class (UCCC) is a direct certification requirement for Z-Wave controllers working with modern locks. It was introduced in the 2024A spec and refined in 2024B, which also added it to the official Compliance Test Tool (CTT v3.9.1). The Z-Wave Certification Portal released an Intermediate Update (2024B-1) to include support for the User Credential Command Class, and the End Device Emulator was updated to support UCCC alongside ZWLR and SmartStart.

The UCCC defines how user credentials — such as PIN codes, passwords, and biometric data — are managed and communicated between devices in a Z-Wave network, covering creation, modification, deletion, validity periods, and associated permissions. Without fully implementing it, Home Assistant cannot be certified as a compliant Z-Wave controller under current specs. This means credential management is not just a UX improvement — it's a prerequisite for OHF's Z-Wave certification path.

Based on the recent introduction of #70 we aim to extend to a full Z-Wave credential management support exposed via Z-Wave JS UI.

#70 aimed to establish a solid foundation to further expand and iterate upon and most importantly surface available Z-Wave credential management options in a more centralized and streamlined approach (also see #69) 

To cover the full Z-Wave credential management scope for future certification we aim to expose them through Z-Wave JS UI.

### Community signals

From a general scoping perspective community signals are mentioned in #70 

#### In scope
- Z-Wave User Credential Command Class (UCCC)
- [Step 1: Credential management UI](https://github.com/OpenHomeFoundation/roadmap/issues/196)

#### Not in scope
- [Step 2: Device management UI](https://github.com/OpenHomeFoundation/roadmap/issues/197)
- Any other command classes
- General feature improvements

### Foreseen solution

- https://github.com/zwave-js/zwave-js-ui/pull/4638

### Risks & open questions

Open questions

- Maybe we're able to (directly) add more functionality to #70 as well

### Appetite

Medium - 4 weeks

### Execution issues

_No response_

### Decision log

| Date | Decision | Outcome |
|------|----------|---------|
|      |          |         |

AlCalzone:
This PR adds a dedicated credential management UI, powered by Z-Wave JS's new Access Control API.
It includes everything that belongs to it: managing credentials, moving them between users, fingerprint and other credential learning, bulk deletion, logging, etc.

This also removes the `userCode` and `userIdStatus` values from the User Code CC dropdown, since those are no longer needed to manage codes.

image

image

image

https://github.com/user-attachments/assets/9a846252-3c4d-4990-a3a9-62eacd301ae7

image


robertsLando:
@AlCalzone do you have support in mock server for it?

AlCalzone:
> @AlCalzone do you have support in mock server for it?

Yeah, both CCs are extensively tested through mocks internally.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+								UserCredentialCCAdminPinCodeReport
+							) {
+								restore()
+								resolve(cmd)
+							}
+							return result
+						}
+					},
+				)
+			: Promise.resolve(null)
```
robertsLando: looks a bit hacky, maybe this requires a fix on driver?

src/components/nodes-table/AccessControl.vue
```diff
@@ -0,0 +1,1923 @@
...
+		},
+		handleLearnProgress(args) {
+			if (!this.learnDialog.show) return
+			if (
+				args.userId !== this.learnDialog.userId ||
+				args.credentialType !== this.learnDialog.type ||
+				args.credentialSlot !== this.learnDialog.slot
+			) {
+				return
+			}
```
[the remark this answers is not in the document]
AlCalzone: we can assume that zwave-js sends correct events.

src/components/nodes-table/AccessControl.vue
```diff
@@ -0,0 +1,1923 @@
...
+		supportsAdminCodeDeactivation() {
+			return !!this.currentEndpoint?.capabilities
+				.supportsAdminCodeDeactivation
+		},
+		adminCodeState() {
+			if (!this.supportsAdminCode) return 'unsupported'
+			const code = this.currentEndpoint?.adminCode
+			if (typeof code === 'string' && code.length > 0) return 'set'
+			return 'unset'
+		},
```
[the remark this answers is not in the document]
AlCalzone: verify against zwave-js sources what the correct behavior is

api/lib/ZwaveClient.ts
AlCalzone: uh what the fuck -> figure out what this is for and if we actually need it. I would guess no.

AlCalzone (commented):
## Deep review — `feat: implement credential management UI`

**Verdict:** Needs work — the approach is sound and composes well with the codebase, but a credential-leakage Blocker and a silent-data-loss Major should be fixed before merge.

**Top 3 risks**
1. Cleartext PINs/passwords broadcast to every connected client and retained in the in-memory event buffer.
2. Adding a user to a full lock silently overwrites user 1 instead of erroring.
3. The admin-code path monkeypatches a zwave-js-private method with a non-reentrant wrapper and reports unconfirmed changes as success.

**Strengths**
- Composes with house style: canonical `validateMethods`/`allowedApis` → `callApi` dispatch, the `emitNodeUpdate(node,{accessControl})` snapshot pattern shared with `userCodes`/`schedule`, and standard async-component tab registration.
- `accessControl.ts` is clean, well-factored pure logic; the result-code → user-message mappers are a nice UX touch.
- Comprehensive: learning/enrollment, bulk delete, move-between-users, admin code — all wired end to end.

**Themes**
1. **Secret handling** — credentials leak through the event stream, the state snapshot, and the mask function.
2. **The admin-code `handleCommand` monkeypatch** — one construct drew correctness, design, readability, and operability findings at once.
3. **Frontend/backend contract duplication** — enums and types hand-mirrored instead of imported.

**FYI (not line-anchored)**
- Backcompat is fine: hiding `userCode`/`userIdStatus` (NodeDetails.vue ~557) only filters the value grid; values stay in `node.values` and on MQTT, so automations are unaffected. But the legacy "Users" (`UserCodeTable`) tab in ExpandedNode.vue still renders alongside the new "Access Control" tab — two overlapping UIs for the same locks.
- `_onNodeEvent` (ZwaveClient.ts:7871) was broadened to `zwaveNode?.nodeId ?? zwaveNode?.id`; equal for root nodes, so no behavior change, but a shared dispatch helper was changed for one feature.
- Confirmed *not* problems: `formatHex` doesn't duplicate the backend `buffer2hex`/`num2hex`; no i18n infra exists so the hardcoded label maps bypass nothing; the two new dialogs are presentational and bypass no dialog base.

**Coverage:** All 7 specialists ran (Correctness, Security, DRY, Design/API, Operability, Tests, Readability). Performance not run — no DB/query/hot-loop surface in its lane. Reviewed against a worktree at PR head `6a0ab63`.

api/lib/ZwaveClient.ts
```diff
@@ -7996,11 +8723,35 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 			'notification',
 			'firmware update progress',
 			'firmware update finished',
+			'user added',
+			'user modified',
+			'user deleted',
+			'credential added',
```
AlCalzone, line 8626: **[Blocker] · Security + Operability**

Binding `credential added`/`credential modified` to `_onNodeEvent` forwards zwave-js's `CredentialChangedArgs`, whose `data` field is the cleartext PIN/password (or binary bytes). `_onNodeEvent` then pushes the raw args into `node.eventsQueue` (~100 events retained inside the node object) and `sendToSocket`s them to every client on the node channel (see the enqueue near :7883).

**Why:** door-lock secrets are broadcast to all connected sessions and retained in an in-memory buffer, though the activity feed only needs `userId`/`type`/`slot`/`status`.

**Fix:** redact/strip `data` (and the raw endpoint object) before enqueue/emit. Flagged independently by two reviewers.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+				} catch (e) {
+					logger.warn(
+						`UC setUser follow-up failed for node ${nodeId}, slot ${userId}: ${(e as Error).message}`,
+					)
+				}
+			}
+			return { userResult, credentialResult }
+		}
+
+		const userResult = await accessControl.setUser(userId, options)
```
AlCalzone: **[Major] · Correctness**

`accessControlAddUser` uses `setUser` (create-*or-update*) rather than `accessControl.addUser` (explicit `operationType: Add`), and `nextFreeUserSlot`/`nextFreeCredentialSlot` (accessControl.ts:320-343) return `1` when every slot is full.

**Why:** adding a user to a full lock silently overwrites user 1's fields and its slot-1 credential instead of surfacing `Error_AddRejectedLocationOccupied` — silent data loss.

**Fix:** return a sentinel when full and guard "Add"; prefer `accessControl.addUser`, which rejects occupied slots (it exists — the "not implemented yet" comment at :1750 is stale, and it also enforces the credential requirement and `credential.slot === userId`).

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+				maxCredentialLength: c.maxCredentialLength,
+				maxCredentialHashLength: c.maxCredentialHashLength,
+				supportsCredentialLearn: c.supportsCredentialLearn,
+				credentialLearnRecommendedTimeout:
+					c.credentialLearnRecommendedTimeout,
+				credentialLearnNumberOfSteps: c.credentialLearnNumberOfSteps,
+			})
+		}
+
+		const requiresCredentialAtUserCreation = !endpoint.supportsCC(
```
AlCalzone: **[Major] · Correctness**

`requiresCredentialAtUserCreation` is derived from `!endpoint.supportsCC('User Credential')`, ignoring `userCaps.supportsUsersWithoutCredentials` (read but never used).

**Why:** a U3C lock that forbids credential-less users is shown as "credential optional," so the UI lets you create a user with no credential and the backend writes a phantom/rejected user.

**Fix:** compute from `supportsUsersWithoutCredentials` (optionally OR'd with the legacy-UC check).

src/lib/accessControl.ts
```diff
@@ -0,0 +1,397 @@
...
+			return 'Step needs to be retried.'
+		case UserCredentialLearnStatus.InvalidAddOperationType:
+		case UserCredentialLearnStatus.InvalidModifyOperationType:
+			return 'The lock refused the operation.'
+		default:
+			return 'Enrollment in progress…'
+	}
+}
+
+export function maskCredential(c: { data?: string; isBinary?: boolean }) {
```
AlCalzone: **[Major] · Security + Tests**

`maskCredential` returns `formatHex(c.data)` for binary types — the full credential value unmasked — and the table never gates binary types behind the reveal toggle.

**Why:** RFID/NFC/biometric/DESFire values render in cleartext by default (screenshots, shoulder-surf). A function whose job is "never leak secrets" has an in-the-clear path, and no test pins the intended behavior.

**Fix:** mask binary values too; gate any reveal behind the explicit toggle, and add a test asserting the intended `isBinary` behavior.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+		const code = await accessControl.getAdminCode()
+		const node = this._nodes.get(nodeId)
+		if (node?.accessControl) {
+			const epEntry = node.accessControl.endpoints.find(
+				(e) =>
+					e.endpointIndex ===
+					(endpointIndex ?? node.accessControl.primaryEndpoint),
+			)
+			if (epEntry) {
+				epEntry.adminCode = code === undefined ? null : code
```
AlCalzone: **[Major] · Security**

The `node.accessControl` snapshot carries cleartext PINs/passwords/admin codes to the frontend, re-broadcasts them via `emitNodeUpdate` to all sessions, and holds them in Vuex for the node's lifetime.

**Why:** every connected client (and anyone inspecting socket frames) obtains all lock secrets, not just the operator who set them.

**Fix:** a threat-model decision on whether read-back is warranted; at minimum don't retain after the dialog closes and don't broadcast to unrelated sessions. Confirm whether zwave-js actually populates `data` on read-back vs only echoing this-session input — it changes the severity.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+				this.emitNodeUpdate(node, { accessControl: node.accessControl })
+			}
+		}
+		return { code: code === undefined ? null : code }
+	}
+
+	async accessControlSetAdminCode(
```
AlCalzone, line 1944: **[Major] · Design + Correctness + Readability**

`accessControlSetAdminCode` swaps the node's private `handleCommand` (four `as any` casts) to sniff an `AdminPinCodeReport`.

**Why:** it's not reentrant — a second concurrent call (or any command in the ~2s window) captures the first call's wrapper as its "original," so `restore()` can reinstall a stale handler or orphan one permanently; and it depends on an undocumented internal that a zwave-js rename breaks silently. It also interleaves three concerns (supervision call, report interception, cache update) in one 85-line method.

**Fix:** single-flight per node, or subscribe via a public CC-event accessor; push the missing `operationResult` gap upstream to zwave-js rather than patching internals.

src/lib/accessControl.ts
```diff
@@ -0,0 +1,397 @@
+/**
+ * Frontend constants + helpers for Z-Wave User Credential / User Code CC.
+ * Mirrors enums from `@zwave-js/cc` so the frontend doesn't depend on
+ * the backend package directly.
+ */
+
+export enum UserCredentialType {
```
AlCalzone: **[Major] · DRY + Design**

Hand-mirrors ~8 zwave-js protocol enums with hardcoded wire values, justified by "so the frontend doesn't depend on the backend package directly."

**Why:** that rationale is false for this repo — the frontend already imports these enums directly (`App.vue`, `NodeScheduler.vue`, `AssociationGroups.vue`, `DialogHealthCheck.vue` from `zwave-js`/`@zwave-js/cc`), so the bundle already includes them and a hand-copied numeric enum silently corrupts CC payloads if it drifts from the source of truth.

**Fix:** import the real enums from `zwave-js`/`@zwave-js/cc` as siblings do; keep only the label/icon/message maps local.

src/components/nodes-table/AccessControl.vue
```diff
@@ -0,0 +1,1923 @@
...
+				this.currentEndpoint.credentials
+					.filter((c) => c.type === type)
+					.map((c) => c.slot),
+			)
+			for (let i = 1; i <= (cap?.numberOfCredentialSlots ?? 1); i++) {
+				if (!used.has(i)) return i
+			}
+			return 1
+		},
+		async startLearn({ user, type, slot, timeout, totalSteps }) {
```
AlCalzone, line 1344: **[Major] · Operability + Correctness**

The `persistent` enroll dialog has no client-side watchdog and waits solely on socket events; a mid-learn disconnect strands the operator on a frozen spinner, and `cancelLearn`'s `apiRequest` won't resolve while disconnected (App.vue queues it). Separately, the queue watcher (:1768) keys on `eventsQueue.length` and reads only the tail element, so two events in one flush (e.g. `learn completed` + `credential added`) drop the completed handler.

**Why:** a 3am disconnect during enrollment leaves an unclosable modal.

**Fix:** arm a local timeout (from `learnDialog.timeout` + grace) that transitions to a failure state and fail/allow-close on disconnect; diff the queue delta instead of reading the tail.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+		const { accessControl } = this._accessControlEndpoint(
+			zwaveNode,
+			endpointIndex,
+		)
+		const credentials = await accessControl.getAllCredentials()
+		this._refreshAccessControlState(zwaveNode)
+		return credentials.map((c) => this._mapCredentialData(c))
+	}
+
+	async accessControlSetUser(
```
AlCalzone, line 1720: **[Major] · Operability**

None of the mutating methods (`setUser`, `setCredential`, `delete*`, `assignCredential`, learn start/cancel, `setAdminCode`) call `this.logNode`, though the PR description says it adds "logging" and sibling features log at info.

**Why:** no server-side audit trail of who changed lock state or the returned result codes.

**Fix:** `logNode(info)` per op with node/endpoint/user/slot/type/result — never the value.

src/lib/accessControl.ts
```diff
@@ -0,0 +1,397 @@
...
+	if (!hex) return ''
+	return (
+		hex
+			.toUpperCase()
+			.match(/.{1,2}/g)
+			?.join(' ') ?? hex
+	)
+}
+
+export function nextFreeUserSlot(
```
AlCalzone, line 278: **[Major] · Tests**

Zero tests for `nextFreeUserSlot`/`nextFreeCredentialSlot` (the fallback-to-1 path), `maskCredential`, `formatHex`, `isDirectEntry`/`isBinary`, and the result-code mappers, though the repo unit-tests exactly this shape of pure logic (`src/modules/ColumnFilterHelper.test.ts`, vitest).

**Why:** security-critical, zero-dependency logic gating a lock UI, trivially testable, and the convention already exists — it just wasn't applied here.

**Fix:** add `src/lib/accessControl.test.ts` covering slot allocation at 0/1/N-1/N occupancy (incl. the fallback-to-1 return), `maskCredential` null/empty/short/binary, `formatHex` odd-length/empty, and each mapper's default branch.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+			: Promise.resolve(null)
+		try {
+			await accessControl.setAdminCode(code)
+		} catch (e) {
+			restore()
+			throw e
+		}
+		const report = await reportPromise
+		const result: UserCredentialAdminCodeOperationResult =
+			report?.operationResult ??
```
AlCalzone: **[Minor] · Correctness + Operability**

On admin-code timeout (no report within `REPORT_WAIT_MS`, and every non-U3C call which resolves `null`), `result` defaults to `Modified` (success) and the UI stores the code.

**Why:** a silently-failing or slow lock is reported to the user as "Admin code updated."

**Fix:** treat "no confirmation" as unconfirmed (distinct from success), or re-`Get` to confirm.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+			UserCredentialType.UWB,
+			UserCredentialType.EyeBiometric,
+			UserCredentialType.FaceBiometric,
+			UserCredentialType.FingerBiometric,
+			UserCredentialType.HandBiometric,
+			UserCredentialType.UnspecifiedBiometric,
+			UserCredentialType.DESFire,
+		]
+		if (binaryTypes.includes(type)) {
+			return new Uint8Array(Buffer.from(data, 'hex'))
```
AlCalzone: **[Minor] · Security**

`Buffer.from(data,'hex')` silently drops non-hex nibbles and there's no server-side length/charset check; the client-side `canSave` guard is bypassable by a direct socket call.

**Why:** malformed/oversized credential data reaches `setCredential`, and silent hex truncation writes a different credential than intended.

**Fix:** validate hex and length at the API boundary.

src/lib/accessControl.ts
```diff
@@ -0,0 +1,397 @@
...
+			return 'The lock refused the operation.'
+		default:
+			return 'Enrollment in progress…'
+	}
+}
+
+export function maskCredential(c: { data?: string; isBinary?: boolean }) {
+	if (c.data == null) return '—'
+	if (c.isBinary) return formatHex(c.data)
+	return '•'.repeat(Math.max(c.data.length, 4))
```
AlCalzone: **[Minor] · Security**

`'•'.repeat(Math.max(c.data.length, 4))` reveals the exact credential length (for length ≥ 4) in the DOM.

**Fix:** use a fixed-width mask.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+				(options.userType != null && options.userType !== 0) ||
+				options.userName != null ||
+				options.credentialRule != null ||
+				options.expiringTimeoutMinutes != null
+			if (hasNonDefaultFields) {
+				try {
+					userResult = await accessControl.setUser(userId, options)
+				} catch (e) {
+					logger.warn(
+						`UC setUser follow-up failed for node ${nodeId}, slot ${userId}: ${(e as Error).message}`,
```
AlCalzone: **[Minor] · Operability**

In the legacy User-Code add path, a thrown `setUser` follow-up is caught to `logger.warn` but `userResult` stays `OK`.

**Why:** user name/type failures silently show as full success in the UI.

**Fix:** propagate the partial failure to the result.

src/components/nodes-table/AccessControl.vue
```diff
@@ -0,0 +1,1923 @@
...
+			)
+			if (!res.success) {
+				this.learnDialog.state = 'refused'
+				this.learnDialog.message = res.message
+			}
+			// Stay in 'starting' (indeterminate spinner) until the first
+			// learn-progress report arrives — handleLearnProgress flips
+			// state to 'waiting' once the lock acknowledges with steps.
+		},
+		async cancelLearn() {
```
AlCalzone, line 1380: **[Minor] · Operability**

`cancelLearn` runs with `errorSnack:false` and closes regardless of outcome.

**Why:** a failed cancel (lock still in learn mode) gives the operator no signal.

**Fix:** surface cancel failure.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+		return accessControl.setUser(userId, options)
+	}
+
+	/**
+	 * Add a user (with optional initial credential). Spec calls for
+	 * `accessControl.addUser` but it isn't implemented in zwave-js yet —
+	 * we use setUser (which auto-detects Add) and follow up with
+	 * setCredential when one was supplied.
+	 */
+	async accessControlAddUser(
```
AlCalzone, line 1734: **[Minor] · Design**

Inner result shapes are inconsistent across the family: `setUser`/`setCredential`/`deleteUser` return a raw enum, `addUser` returns `{userResult, credentialResult}`, `getAdminCode` returns `{code}`, `setAdminCode` returns `{result}`.

**Why:** every caller special-cases the unwrap (`res.result` vs `res.result.userResult` vs `res.result.result`); siblings like `setSchedule` return the raw zwave-js result.

**Fix:** pick one convention (raw result, or a consistently-keyed object) across the family.

src/lib/accessControl.ts
```diff
@@ -0,0 +1,397 @@
...
+	const used = new Set(
+		credentials.filter((c) => c.type === type).map((c) => c.slot),
+	)
+	for (let i = 1; i <= max; i++) {
+		if (!used.has(i)) return i
+	}
+	return 1
+}
+
+export type AccessControlUser = {
```
AlCalzone: **[Minor] · Readability + Design + DRY**

The exported `AccessControl*` types are never imported (the `.vue` files consume `node.accessControl` untyped) and duplicate the backend `ZUIAccessControl*` types field-for-field.

**Why:** no compile-time contract across the socket boundary despite the appearance of one, and the two copies drift as the payload evolves.

**Fix:** either type the payload with them (via the existing `@server` alias) or delete them.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+		data?: string
+		isBinary?: boolean
+	} {
+		if (data === undefined) return {}
+		if (typeof data === 'string') return { data, isBinary: false }
+		const hex = Buffer.from(data).toString('hex')
+		return { data: hex, isBinary: true }
+	}
+
+	private _decodeCredentialData(
```
AlCalzone: **[Minor] · DRY**

The binary-credential-type list in `_decodeCredentialData` duplicates the frontend `binaryTypes` set (accessControl.ts:200-211) — two hand-maintained lists of the same 10 types that must agree for encode/decode to round-trip.

**Why:** divergence mis-encodes a credential with no error.

**Fix:** derive binary-ness from zwave-js (the CC already classifies credential data) instead of two literal lists.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+	): Promise<{ code: string | null }> {
+		const zwaveNode = this._requireNode(nodeId)
+		const { accessControl } = this._accessControlEndpoint(
+			zwaveNode,
+			endpointIndex,
+		)
+		const code = await accessControl.getAdminCode()
+		const node = this._nodes.get(nodeId)
+		if (node?.accessControl) {
+			const epEntry = node.accessControl.endpoints.find(
```
AlCalzone: **[Minor] · Readability**

The `endpoints.find(e => e.endpointIndex === (endpointIndex ?? ...primaryEndpoint))` expression is duplicated verbatim here and at :2082-2086.

**Why:** a bugfix to one copy can miss the other.

**Fix:** extract `_findAccessControlEndpointEntry(node, endpointIndex)`.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		}
 	}
 
+	// ---------------- Access Control (User Credential CC / User Code CC) ----------------
+
+	private _accessControlEndpoints(zwaveNode: ZWaveNode): number[] {
```
AlCalzone: **[Minor] · Readability**

`_accessControlEndpoints` (returns `number[]`) and `_accessControlEndpoint` (:1500, returns `{endpoint, accessControl, index}`) differ only by a trailing "s" but have unrelated return shapes.

**Why:** easy to call the wrong one at a glance.

**Fix:** rename, e.g. `_supportedAccessControlIndices` / `_resolveAccessControlEndpoint`.

src/components/nodes-table/AccessControl.vue
```diff
@@ -0,0 +1,1923 @@
...
+									'The lock refused the operation.'
+								}}
+							</template>
+						</div>
+					</div>
+				</v-card-text>
+				<v-card-actions>
+					<v-spacer />
+					<v-btn
+						v-if="
```
AlCalzone: **[Minor] · Readability**

The enroll dialog's 7-state machine is implicit — action buttons are gated by inline arrays re-tested against `learnDialog.state` (`['waiting','starting'].includes(...)`, a 5-item Close list, etc.) scattered through the template.

**Why:** adding/renaming a state means hunting three separate array literals.

**Fix:** add `canCancelLearn`/`canRetryLearn`/`canCloseLearn` computeds next to the existing `learnDialogResult`.

src/components/nodes-table/AccessControl.vue
```diff
@@ -0,0 +1,1923 @@
...
+			return 'primary'
+		},
+		activityIcon(event) {
+			if (event.includes('user')) return 'person'
+			if (event.includes('credential learn')) return 'fingerprint'
+			if (event.includes('credential')) return 'key'
+			return 'info'
+		},
+		activitySummary(evt) {
+			const arg = evt.args?.[1] ?? evt.args?.[0]
```
AlCalzone: **[Minor] · Readability**

`evt.args?.[1] ?? evt.args?.[0]` is repeated three times (here and :1772, :1774) with no explanation; the reason (endpoint CC events emit `(endpoint, payload)` not `(node, payload)`) is documented only in ZwaveClient.ts:759, invisible from this file.

**Fix:** factor into an `eventPayload(evt)` helper carrying the one-line why.

src/components/nodes-table/AccessControl.vue
```diff
@@ -0,0 +1,1923 @@
...
+				slot: this.learnDialog.slot,
+				timeout: this.learnDialog.timeout,
+				totalSteps: this.learnDialog.totalSteps,
+			})
+		},
+		closeLearn() {
+			this._stopRetryFlash()
+			this.learnDialog.show = false
+		},
+		_triggerRetryFlash() {
```
AlCalzone, line 1411: **[Minor] · Readability**

`_triggerRetryFlash` hardcodes `step = 90` and loops `i < 3` with no named constant.

**Fix:** hoist to `RETRY_FLASH_STEP_MS` / `RETRY_FLASH_COUNT`.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+				this.emitNodeUpdate(node, { accessControl: undefined })
+			}
+			return
+		}
+		node.accessControl = ac
+		this.emitNodeUpdate(node, { accessControl: ac })
+	}
+
+	/** Clear the UI snapshot after a wildcard User Credential CC deletion */
+	private _clearAccessControlEndpointState(
```
AlCalzone: **[Nit] · Readability**

`_clearAccessControlEndpointState` is never called — both `accessControlDeleteAllUsers`/`accessControlDeleteCredentials` use `_refreshAccessControlState` instead.

**Fix:** remove it or wire it in.

api/lib/ZwaveClient.ts
AlCalzone: refactor this to addUser

api/lib/ZwaveClient.ts
AlCalzone: we should not leak the admin code, only transport whether it is set or not

api/lib/ZwaveClient.ts
AlCalzone: We really should not need this. What's the purpose?

src/components/nodes-table/AccessControl.vue
AlCalzone: If this concerns a disconnect between the UI and the backend, I do not care. The frontend can just be reloaded.

api/lib/ZwaveClient.ts
AlCalzone: The "logging" in the PR title refers to the access control log tab, not logging as text.

api/lib/ZwaveClient.ts
AlCalzone: calls to `setAdminCode` should inspect the supervision result, which already contains enough info to know if the change succeeded. We don't need the whole hacky dance we're doing here.

api/lib/ZwaveClient.ts
AlCalzone: The way I see it, this method only gets called from the frontend dialogs. These should only allow the user to set PIN/Password, and not any binary credentials. --> Make sure the frontend/API only allows manually setting those two types, then we can avoid all of this.

api/lib/ZwaveClient.ts
AlCalzone: we MUST use `addUser` to add users. setUser can only be used to modify them.

api/lib/ZwaveClient.ts
AlCalzone: > addUser returns {userResult, credentialResult}, getAdminCode returns {code}, setAdminCode returns {result}.

These all seem reasonable to me.

src/lib/accessControl.ts
AlCalzone: Me like strong types.

api/lib/ZwaveClient.ts
AlCalzone: see comment above regarding this.

src/components/nodes-table/AccessControl.vue
AlCalzone: Yes, I also found this confusing when reading. lets simplify this.

api/lib/ZwaveClient.ts
AlCalzone: remove I would say.

src/components/nodes-table/AccessControl.vue
```diff
@@ -0,0 +1,1923 @@
...
+				this.credentialsFor({ userId: cred.userId }).length === 1
+			const cascadeMsg = cascade
+				? `<br><b>This is the user's last credential.</b> The user will also be deleted.`
+				: ''
+			const ok = await this.app.confirm(
+				`Delete ${credentialTypeLabel(cred.type)} (slot ${cred.slot})?`,
+				`Owner: ${user?.userName || `User ${cred.userId}`}${cascadeMsg}`,
+				'warning',
+				{ confirmText: 'Delete', cancelText: 'Cancel' },
+			)
```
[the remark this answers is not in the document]
AlCalzone: Fixed in d4eae670 — the owner name is HTML-escaped before interpolation.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — the binary decoder and duplicated type list were removed.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 and d4eae670 — creation uses addUser and full slot sets no longer fall back to slot 1.

src/lib/accessControl.ts
AlCalzone: Fixed in d4eae670 — helper tests now cover slot exhaustion, masks, formatting, classifiers, and mapper fallbacks.

src/components/nodes-table/AccessControl.vue
```diff
@@ -0,0 +1,1923 @@
...
+		},
+		handleLearnCompleted(args) {
+			if (!this.learnDialog.show) return
+			if (
+				args.userId !== this.learnDialog.userId ||
+				args.credentialType !== this.learnDialog.type ||
+				args.credentialSlot !== this.learnDialog.slot
+			) {
+				return
+			}
```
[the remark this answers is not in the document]
AlCalzone: No change needed — zwave-js guarantees CredentialLearnCompletedArgs for this event.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — admin state changes only after an accepted supervision result.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — the requirement now comes from supportsUsersWithoutCredentials.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — endpoint snapshot lookup now uses one helper.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — the API now accepts manual PIN and password credential types only.

src/lib/accessControl.ts
AlCalzone: Fixed in d4eae670 — every credential value uses the same fixed mask, with coverage for binary values.

src/components/nodes-table/AccessControl.vue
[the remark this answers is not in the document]
AlCalzone: Verified against zwave-js: an empty code deactivates it. c8385c19 now exposes a separate adminCodeSet boolean instead.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — the helpers are now named by supported-index lookup and endpoint resolution.

src/components/dialogs/DialogAccessControlUser.vue
```diff
@@ -0,0 +1,433 @@
...
+						cap &&
+						(v.length < cap.minCredentialLength ||
+							v.length > cap.maxCredentialLength)
+					) {
+						return this.lengthHint
+					}
+					return true
+				},
+			]
+		},
```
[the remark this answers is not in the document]
AlCalzone: Fixed in d4eae670 — initial PINs now require digits in both validation paths.

api/lib/ZwaveClient.ts
AlCalzone: Fixed the admin-code exposure in c8385c19 — the snapshot now carries only adminCodeSet. Credential values remain available for the explicit reveal UI.

src/lib/accessControl.ts
AlCalzone: Fixed in d4eae670 — credential masks use a constant width.

src/components/nodes-table/AccessControl.vue
AlCalzone: Fixed in d4eae670 — cancel, retry, and close visibility now use computed state helpers.

src/components/dialogs/DialogAccessControlUser.vue
```diff
@@ -0,0 +1,433 @@
...
+								<v-col cols="8" sm="9">
+									<v-text-field
+										v-model="credential.data"
+										label="Value"
+										:hint="lengthHint"
+										persistent-hint
+										class="font-monospace"
+										:rules="credentialRules"
+										:disabled="saving"
+									/>
```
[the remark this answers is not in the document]
AlCalzone: Fixed in d4eae670 — the initial credential is masked with an explicit reveal control.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — the private command-handler monkeypatch was removed.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — user creation now uses addUser without the setUser follow-up.

src/components/nodes-table/AccessControl.vue
AlCalzone: Fixed in d4eae670 — endpoint event payload selection is centralized in eventPayload.

src/lib/accessControl.ts
AlCalzone: Fixed in d4eae670 — protocol enums now come from @zwave-js/cc and zwave-js.

api/lib/ZwaveClient.ts
```diff
@@ -1397,6 +1482,627 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+					(e) =>
+						e.endpointIndex ===
+						(endpointIndex ?? node.accessControl.primaryEndpoint),
+				)
+				if (epEntry) {
+					epEntry.adminCode = code === '' ? null : code
+					this.emitNodeUpdate(node, {
+						accessControl: node.accessControl,
+					})
+				}
```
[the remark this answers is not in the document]
AlCalzone: Fixed in c8385c19 — only adminCodeSet is stored and emitted.

src/components/nodes-table/AccessControl.vue
AlCalzone: Fixed in d4eae670 — cancellation failures are surfaced and leave the dialog open.

src/components/nodes-table/AccessControl.vue
AlCalzone: Fixed in d4eae670 — retry interval and count now use named constants.

src/components/nodes-table/AccessControl.vue
AlCalzone: Fixed the queue-loss issue in d4eae670 — every unseen event is processed. Per guidance, no disconnect watchdog was added.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — the private handleCommand interception was removed in favor of supervision.

api/lib/ZwaveClient.ts
AlCalzone: No change needed — the operation-specific result shapes are intentional.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — the unused endpoint-clear helper was removed.

src/lib/accessControl.ts
AlCalzone: Fixed in d4eae670 — frontend aliases now reuse the exported backend socket types.

api/lib/ZwaveClient.ts
AlCalzone: No change needed — logging here refers to the Access Control activity tab, not textual server logging.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in c8385c19 — credential data is stripped before events enter the queue or socket stream.

robertsLando (approved):

PR #4645, opened 2026-05-22
feat(ui): vuetify config + design tokens for dashboard rework

AlCalzone:
This PR is the first of many that will modernize the frontend design of Z-Wave JS UI:

grafik

The current PR sets up the design tokens (--zw-*) and Vuetify defaults that the rest of the dashboard rework builds on, so the atoms and components arriving in later commits can just consume the vocabulary instead of redeclaring colours, radii, and component chrome at every call site. Also switches the legacy input family from underlined to outlined+compact so the old pages already look at home next to the new dashboard while we migrate.

This PR also starts the transition to vuetify0 as a headless UI framework, which makes maintaining a custom design language much easier.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,10 @@
       "license": "MIT",
       "dependencies": {
         "@esm2cjs/escape-string-regexp": "^5.0.0",
+        "@fontsource/roboto": "^5.2.10",
+        "@fontsource/roboto-mono": "^5.2.9",
         "@kvaster/zwavejs-prom": "^0.0.3",
+        "@vuetify/v0": "~1.0.0-alpha.5",
         "@zwave-js/log-transport-json": "^3.0.0",
         "@zwave-js/server": "^3.8.0",
         "archiver": "^7.0.1",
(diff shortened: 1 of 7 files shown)


index.html
```diff
@@ -29,6 +29,10 @@
 		<title>Z-Wave JS UI</title>
 		<meta name="description" content="Z-Wave JS UI">
 
+		<link rel="preconnect" href="https://fonts.googleapis.com" />
+		<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+		<link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&family=Roboto+Mono:wght@400;500;600&display=swap" rel="stylesheet" />
+
```
AlCalzone: I think we need to bundle those fonts, not load them at runtime.

index.html
robertsLando, line 1: This should be shipped with frontend to avoid external sources

robertsLando (changes requested):

index.html
AlCalzone: yep, already on it.

robertsLando (commented):
apart from the palette I think this overlaps with #4647

robertsLando (approved):

AlCalzone:
Before you merge, I think I have one more thing to fold into this PR.

robertsLando (approved):

AlCalzone:
@robertsLando This one would be most important to review as it comes first.

robertsLando:
@AlCalzone it's ok 👍🏼

PR #4646, opened 2026-05-22
feat(ui): add Lucide icon set

AlCalzone:
Introduces Lucide icon ESM imports for the new dashboard, with named aliases and a closed ICON_SIZE vocabulary so call sites can't pick arbitrary pixel sizes. Material Icons stays registered for the legacy settings/dialog pages, so this only affects new dashboard components.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "@fontsource/roboto-mono": "^5.2.9",
     "@kvaster/zwavejs-prom": "^0.0.3",
     "@vuetify/v0": "~1.0.0-alpha.5",
+    "@lucide/vue": "^1.16.0",
     "@zwave-js/log-transport-json": "^3.0.0",
     "@zwave-js/server": "^3.8.0",
     "archiver": "^7.0.1",
--- a/src/lib/icons.ts
+++ b/src/lib/icons.ts
@@ -1,0 +1,96 @@
+// src/lib/icons.ts
+//
+// Central re-export of the dashboard's icon vocabulary.
+// One alias per handoff name (`.design-handoff/project/icons.jsx`),
+// suffixed `Icon` so call sites read `<AddIcon :size="ICON_SIZE.button" />`.
+//
+// Call sites import directly:
+//   import { AddIcon, ICON_SIZE } from '@/lib/icons'
+//
+// Lucide ships per-icon ESM entry points; Vite tree-shakes unused aliases.
+//
+// A few Lucide glyphs back two semantic roles in the handoff:
+//   - Sun → DimmerIcon (dimmable lights) and SunIcon (weather/illuminance)
+//   - Thermometer → ThermostatIcon (climate device) and TempIcon (sensor)
+// Kept as distinct aliases so call sites express intent — a future swap
+// to a dedicated glyph touches one line here, not every consumer.
+
+export {
+	Plus as AddIcon,
+	ArrowDown as ArrowDownIcon,
+	ArrowUp as ArrowUpIcon,
+	TriangleAlert as AlertIcon,
+	Battery as BatteryIcon,
+	Bell as BellIcon,
+	Blinds as ShadeIcon,
+	Lightbulb as BulbIcon,
+	ChartLine as GraphIcon,
+	Check as CheckIcon,
+	ChevronDown as ChevronDownIcon,
+	ChevronRight as ChevronRightIcon,
+	ChevronUp as ChevronUpIcon,
+	CircleCheck as StatusIcon,
+	ClipboardList as InterviewIcon,
+	Clock as ClockIcon,
+	Cpu as ControllerIcon,
+	DoorClosed as ContactIcon,
+	Download as DownloadIcon,
+	Droplet as LeakIcon,
+	Filter as FilterIcon,
+	House as LocationsIcon,
+	LayoutGrid as GridIcon,
+	Layers as SceneIcon,
+	Link as LinkIcon,
+	List as ListIcon,
+	Lock as LockIcon,
+	Menu as MenuIcon,
+	Moon as MoonIcon,
+	MoreHorizontal as MoreIcon,
+	Network as NetworkIcon,
+	Pencil as EditIcon,
+	Play as PlayIcon,
+	Plug as PlugIcon,
+	Power as PowerIcon,
+	QrCode as QrIcon,
+	Radio as MotionIcon,
+	RefreshCw as RefreshIcon,
+	Search as SearchIcon,
+	Settings as SettingsIcon,
+	Signal as SignalIcon,
+	SignalHigh as SignalHighIcon,
+	SignalLow as SignalLowIcon,
+	SignalMedium as SignalMedIcon,
+	Siren as SirenIcon,
+	Sun as DimmerIcon,
+	Sun as SunIcon,
+	Tablet as RemoteIcon,
+	Thermometer as ThermostatIcon,
+	Thermometer as TempIcon,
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -28,7 +28,6 @@
 		<meta name="theme-color" content="#ffffff" />
 		<title>Z-Wave JS UI</title>
 		<meta name="description" content="Z-Wave JS UI">
-
```
robertsLando, line 31: remove

robertsLando (commented):
Doesn't this overlap 1-1 with #4647 ?

package.json
```diff
@@ -77,7 +77,11 @@
   ],
   "dependencies": {
     "@esm2cjs/escape-string-regexp": "^5.0.0",
+    "@fontsource/roboto": "^5.2.10",
+    "@fontsource/roboto-mono": "^5.2.9",
     "@kvaster/zwavejs-prom": "^0.0.3",
+    "@vuetify/v0": "~1.0.0-alpha.5",
+    "@lucide/vue": "^1.16.0",
```
robertsLando, line 84: PR title/body says *"Swaps Material Icons for per-icon Lucide ESM imports in the new dashboard"*, but the actual delta in this PR is just this dep + `src/lib/icons.ts` — no SFC swaps anywhere. Could you reword the title/body to reflect that this introduces the vocabulary and the swaps come in follow-up PRs? Otherwise reviewers (and future `git log` readers) will look for changes that aren't here.

src/lib/icons.ts
```diff
@@ -0,0 +1,96 @@
...
+	Menu as MenuIcon,
+	Moon as MoonIcon,
+	MoreHorizontal as MoreIcon,
+	Network as NetworkIcon,
+	Pencil as EditIcon,
+	Play as PlayIcon,
+	Plug as PlugIcon,
+	Power as PowerIcon,
+	QrCode as QrIcon,
+	Radio as MotionIcon,
```
robertsLando, line 56: `Radio` (radio-waves glyph) → `MotionIcon` is the most surprising mapping in the file and the only one without a why. `Sun→Dimmer` and `Thermometer→Thermostat` both get an explanatory note in the header comment; worth adding `Radio→Motion` to that list ("radio waves ≈ motion-detection cone" or similar). Otherwise the next person reading this will think it's a typo.

src/lib/icons.ts
```diff
@@ -0,0 +1,96 @@
...
+	RefreshCw as RefreshIcon,
+	Search as SearchIcon,
+	Settings as SettingsIcon,
+	Signal as SignalIcon,
+	SignalHigh as SignalHighIcon,
+	SignalLow as SignalLowIcon,
+	SignalMedium as SignalMedIcon,
```
robertsLando: Lone truncation: every other alias is a clean semantic rename, but `SignalMedium` becomes `SignalMedIcon`. Either keep `SignalMediumIcon` to match the rest, or rename `SignalHigh`/`SignalLow` to `SignalHi`/`SignalLo` so the shortening is consistent. Pick one — current state reads like an accident.

src/lib/icons.ts
```diff
@@ -0,0 +1,96 @@
...
+	Upload as UploadIcon,
+	X as XIcon,
+	Zap as ZapIcon,
+	Activity as PulseIcon,
+} from '@lucide/vue'
+
+/**
+ * Closed vocabulary of icon sizes used across the dashboard.
+ * Call sites read `<AddIcon :size="ICON_SIZE.button" />` — numeric
+ * literals are forbidden in dashboard components.
```
robertsLando: *"numeric literals are forbidden in dashboard components"* — this is aspirational; nothing actually enforces it. Either soften to *"by convention"*, or queue a follow-up ESLint rule banning numeric `size="N"` on these icon components. Without enforcement the closed vocabulary depends on reviewer vigilance, which is the thing the closed vocabulary was supposed to remove.

src/lib/icons.ts
```diff
@@ -0,0 +1,96 @@
...
+	pill: 10, // pill leading glyph
+	chip: 11, // chip glyph, table-row inline icon, segmented glyph, columns-menu filter
+	sortArrow: 12, // header chevrons, sort-direction arrows
+	button: 14, // primary button leading icon, table-row archetype glyph, search-input prepend
+	nav: 16, // sidebar nav row, top-bar search glyph
+	topbar: 18, // top-bar icon buttons, menu icon, status icons, drawer close button
+	drawerHeader: 22, // drawer header icon stamp
+} as const
+
+export type IconSize = (typeof ICON_SIZE)[keyof typeof ICON_SIZE]
```
robertsLando, line 97: Not introduced by this PR, but worth flagging while the type contract is fresh: `tsconfig.json` only includes `api/**/*.ts`, and there's no `vue-tsc --noEmit` step in the build. So `IconSize`/`ICON_SIZE` are advisory at the consumer call sites — Vite/esbuild will happily transpile `:size="99"` even though it's not in the vocabulary. If the dashboard rework intends type safety, a frontend type-check step is the missing piece (same observation applies to `palette.ts`/`vuetify0.ts` added earlier in the chain).

src/lib/icons.ts
```diff
@@ -0,0 +1,96 @@
...
+ * surfaces); adding a new numeric value is a design decision.
+ */
+export const ICON_SIZE = {
+	pill: 10, // pill leading glyph
+	chip: 11, // chip glyph, table-row inline icon, segmented glyph, columns-menu filter
+	sortArrow: 12, // header chevrons, sort-direction arrows
+	button: 14, // primary button leading icon, table-row archetype glyph, search-input prepend
+	nav: 16, // sidebar nav row, top-bar search glyph
+	topbar: 18, // top-bar icon buttons, menu icon, status icons, drawer close button
+	drawerHeader: 22, // drawer header icon stamp
```
robertsLando, line 94: The 1px steps between `pill: 10 / chip: 11 / sortArrow: 12` are near the edge of visual perceptibility. Worth confirming with the designer that these are intentional and not residue from pixel-pushing in a single mockup — sub-pixel-sensitive icon sizes risk looking accidentally inconsistent across density modes or breakpoints once the dashboard is responsive. Not a blocker, but if the design intent is *"chips and sort arrows look the same size"*, collapsing those into one value would make the vocabulary harder to drift.

robertsLando (commented):
Verified the moving parts before reviewing: all 56 referenced Lucide names exist in `@lucide/vue@1.16.0` (checked the bundled `.d.ts`), tree-shaking will fire (package is `"sideEffects": false` with per-icon `dist/esm/icons/*.mjs` files), and lint passes. The closed `ICON_SIZE` vocabulary and the dual-alias pattern (`Sun→Dimmer+Sun`, `Thermometer→Thermostat+Temp`) are nicely thought-through.

Findings posted inline. Nothing blocking — mostly clarifications and small consistency nits.

src/lib/icons.ts
AlCalzone: Now tracked on https://github.com/orgs/zwave-js/projects/7/views/1

src/lib/icons.ts
AlCalzone: Now tracked in https://github.com/orgs/zwave-js/projects/7/views/1

package.json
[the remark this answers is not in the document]
AlCalzone: Tracked in https://github.com/orgs/zwave-js/projects/7/views/1?pane=issue&itemId=204269864

robertsLando (approved):

PR #4647, opened 2026-05-22
feat(ui): implement atomic dashboard components

AlCalzone:
This PR is based on https://github.com/zwave-js/zwave-js-ui/pull/4646

For review: all commits starting with [feat(ui): ZwPill atom (status pill, 7 tones × 2 sizes)](https://github.com/zwave-js/zwave-js-ui/commit/19ad516bd9cdddf8bb471a42d4e7386516f1743c) - the others are part of the preceding PRs.

Adds the twelve atomic dashboard components — status indicators (ZwPill, ZwChip, ZwStatusDot, ZwBatteryMini), form controls (ZwToggle, ZwSlider, ZwSegmented, ZwSearchInput, ZwButton, ZwToggleMenu), and progress/notifier readouts (ZwActivityReadout, ZwUpdateNotifier) — built against the foundation tokens and Lucide icon set. Each component is self-contained with its own tone/size variants exposed as props, so composites and layouts in later branches can compose them without restyling. Includes a small Floating UI wrapper used by the menu/dropdown atoms when native anchor positioning isn't available in the browser runtime.

grafik

grafik

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "@esm2cjs/escape-string-regexp": "^5.0.0",
     "@fontsource/roboto": "^5.2.10",
     "@fontsource/roboto-mono": "^5.2.9",
+    "@floating-ui/dom": "^1.7.6",
     "@kvaster/zwavejs-prom": "^0.0.3",
     "@vuetify/v0": "~1.0.0-alpha.5",
     "@lucide/vue": "^1.16.0",
--- a/src/components/dashboard/atoms/ZwActivityReadout.vue
+++ b/src/components/dashboard/atoms/ZwActivityReadout.vue
@@ -33,43 +33,43 @@ import {
 	ICON_SIZE,
 } from '@/lib/icons'
 
-type TransientType = 'ota' | 'rebuild' | 'interview'
-interface Transient {
-	type: TransientType
+type ActivityType = 'ota' | 'rebuild' | 'interview'
+interface Activity {
+	type: ActivityType
 	label: string
 	progress?: number
 }
 
 const props = withDefaults(
 	defineProps<{
-		transient: Transient
+		activity: Activity
 		variant?: 'table' | 'card'
 	}>(),
 	{ variant: 'table' },
 )
 
-const TRANSIENT_ICON = {
+const ACTIVITY_ICON = {
 	ota: DownloadIcon,
 	rebuild: RefreshIcon,
 	interview: InterviewIcon,
 } as const
 
-const iconComp = computed(() => TRANSIENT_ICON[props.transient.type])
+const iconComp = computed(() => ACTIVITY_ICON[props.activity.type])
 
 // Progress is optional — when undefined the readout renders an
 // indeterminate sweep across the bar with no numeric label. A literal
 // placeholder (e.g. 30%) would be misleading, especially for OTA where
 // users watch the number.
 const pct = computed(() =>
-	props.transient.progress !== undefined
-		? Math.round(props.transient.progress * 100)
+	props.activity.progress !== undefined
+		? Math.round(props.activity.progress * 100)
 		: undefined,
 )
 
 const title = computed(() =>
 	pct.value !== undefined
-		? `${props.transient.label} · ${pct.value}%`
-		: props.transient.label,
+		? `${props.activity.label} · ${pct.value}%`
+		: props.activity.label,
 )
 </script>
 
--- a/src/components/dashboard/atoms/ZwButton.vue
+++ b/src/components/dashboard/atoms/ZwButton.vue
@@ -15,12 +15,7 @@
 <script setup lang="ts">
 import { Button } from '@vuetify/v0'
 
-type Variant =
-	| 'primary'
-	| 'outline'
-	| 'destructive'
-	| 'ghost'
-	| 'mono-outline'
+type Variant = 'primary' | 'outline' | 'destructive' | 'ghost' | 'mono-outline'
 
 withDefaults(
 	defineProps<{
--- a/src/components/dashboard/atoms/ZwButton.vue
+++ b/src/components/dashboard/atoms/ZwButton.vue
@@ -1,0 +1,133 @@
+<template>
+	<Button.Root
+		class="zw-btn zw-focus-ring"
+		:class="[`zw-btn--${variant}`, `zw-btn--${size}`]"
+		:disabled="disabled"
+		@click="emit('click', $event)"
+	>
+		<span v-if="$slots.icon" class="zw-btn__icon">
+			<slot name="icon" />
+		</span>
+		<slot />
+	</Button.Root>
+</template>
+
+<script setup lang="ts">
+import { Button } from '@vuetify/v0'
+
+type Variant =
+	| 'primary'
+	| 'outline'
+	| 'destructive'
+	| 'ghost'
+	| 'mono-outline'
+
+withDefaults(
+	defineProps<{
+		variant?: Variant
+		size?: 'sm' | 'md'
+		disabled?: boolean
+	}>(),
+	{ variant: 'primary', size: 'md', disabled: false },
+)
+
+const emit = defineEmits<{ click: [MouseEvent] }>()
+</script>
+
+<style>
+/* Styles are unscoped because Vuetify0 primitives set inheritAttrs:false,
+   so Vue does not forward the parent's scoped data-v-* hash onto the
+   rendered <button>. Class names are BEM-namespaced under .zw-btn.
+   Each variant only sets the four --btn-* custom properties below;
+   the base rule applies them so chrome (radius, transition, focus
+   ring) stays declared once. */
+.zw-btn {
+	appearance: none;
+	cursor: pointer;
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+	gap: 6px;
+	background: var(--btn-bg);
+	color: var(--btn-color);
+	border: 1px solid var(--btn-border, transparent);
+	box-shadow: var(--btn-shadow, none);
+	font-family: var(--zw-font);
+	font-weight: 600;
+	letter-spacing: 0.4px;
+	text-transform: uppercase;
+	border-radius: 6px;
+	transition:
+		background 0.12s,
+		border-color 0.12s,
+		color 0.12s,
+		box-shadow 0.12s;
+	white-space: nowrap;
+}
+
+.zw-btn[data-disabled='true'] {
+	opacity: 0.5;
+	pointer-events: none;
+}
+
+.zw-btn:hover:not([data-disabled='true']) {
+	background: var(--btn-bg-hover, var(--btn-bg));
+}
+
+.zw-btn--md {
+	padding: 7px 14px;
+	font-size: 12px;
+}
+
+.zw-btn--sm {
+	padding: 5px 10px;
+	font-size: 11px;
+}
+
+.zw-btn__icon {
+	display: inline-flex;
+	align-items: center;
+}
+
+.zw-btn--primary {
+	--btn-bg: var(--zw-accent);
+	--btn-bg-hover: var(--zw-accent-dark);
+	--btn-color: rgb(var(--v0-on-primary));
+	--btn-shadow: 0 1px 2px rgba(var(--v0-primary), 0.3);
+}
+
+.zw-btn--outline {
+	--btn-bg: var(--zw-card);
+	--btn-bg-hover: var(--zw-bg-soft);
+	--btn-color: rgba(var(--v0-on-surface), 0.78);
+	--btn-border: var(--zw-line);
+}
+
+.zw-btn--destructive {
+	--btn-bg: var(--zw-danger-soft);
+	--btn-bg-hover: rgba(var(--v0-danger-soft), 0.7);
+	--btn-color: rgb(var(--v0-error-darken-1));
+	--btn-border: rgba(var(--v0-error), 0.3);
+}
+
+.zw-btn--ghost {
+	--btn-bg: transparent;
+	--btn-bg-hover: rgba(var(--v0-on-surface), 0.04);
+	--btn-color: var(--zw-fg-soft);
+	text-transform: none;
+	letter-spacing: 0;
+	font-weight: 500;
+}
+
+/* node-details Advanced tab grid — mono caps on an outlined chip. */
+.zw-btn--mono-outline {
+	--btn-bg: var(--zw-card);
+	--btn-bg-hover: var(--zw-bg-soft);
+	--btn-color: var(--zw-fg);
+	--btn-border: var(--zw-line);
+	font-family: var(--zw-mono);
+	font-size: 11px;
+	font-weight: 500;
+	letter-spacing: 0.4px;
+}
+</style>
--- a/src/components/dashboard/atoms/ZwChip.vue
+++ b/src/components/dashboard/atoms/ZwChip.vue
@@ -1,0 +1,36 @@
+<template>
+	<span class="zw-chip" :class="`zw-tone-${tone}`">
+		<slot />
+	</span>
+</template>
+
+<script setup lang="ts">
+type ChipTone = 'neutral' | 'accent' | 'ok' | 'warn' | 'danger' | 'asleep'
+
+withDefaults(defineProps<{ tone?: ChipTone }>(), { tone: 'neutral' })
+</script>
+
+<style scoped>
+/* Tone background / foreground come from the shared .zw-tone-<name>
+   utility in tokens.css; this atom only owns shape, density, and the
+   accent variant's 1-px outline (per handoff). Type role is Mono
+   Micro (handoff design-system.jsx:55 — "pill text, column headers"). */
+.zw-chip {
+	display: inline-flex;
+	align-items: center;
+	gap: 4px;
+	white-space: nowrap;
+	padding: 1px 6px;
+	border-radius: 4px;
+	background: var(--tone-bg);
+	color: var(--tone-fg);
+	font: var(--zw-text-mono-micro);
+	line-height: 16px;
+	letter-spacing: 0.1px;
+	text-transform: none;
+}
+
+.zw-chip.zw-tone-accent {
+	border: 1px solid rgba(var(--v0-primary), 0.3);
+}
+</style>
--- a/src/components/dashboard/atoms/ZwChip.vue
+++ b/src/components/dashboard/atoms/ZwChip.vue
@@ -13,8 +13,7 @@ withDefaults(defineProps<{ tone?: ChipTone }>(), { tone: 'neutral' })
 <style scoped>
 /* Tone background / foreground come from the shared .zw-tone-<name>
    utility in tokens.css; this atom only owns shape, density, and the
-   accent variant's 1-px outline (per handoff). Type role is Mono
-   Micro (handoff design-system.jsx:55 — "pill text, column headers"). */
+   accent variant's 1-px outline. Type role is Mono Micro. */
 .zw-chip {
 	display: inline-flex;
 	align-items: center;
--- a/src/components/dashboard/atoms/ZwPill.vue
+++ b/src/components/dashboard/atoms/ZwPill.vue
@@ -26,8 +26,7 @@ withDefaults(
 <style scoped>
 /* Tone background / foreground come from the shared .zw-tone-<name>
    utility in tokens.css; this atom only owns shape and density.
-   Type role is Caption (handoff design-system.jsx:51 lists "pills"
-   under Caption); the sm variant overrides the size to 10px. */
+   Type role is Caption; the sm variant overrides the size to 10px. */
 .zw-pill {
 	display: inline-flex;
 	align-items: center;
--- a/src/components/dashboard/atoms/ZwPill.vue
+++ b/src/components/dashboard/atoms/ZwPill.vue
@@ -1,0 +1,52 @@
+<template>
+	<span class="zw-pill" :class="[`zw-tone-${tone}`, `zw-pill--${size}`]">
+		<slot />
+	</span>
+</template>
+
+<script setup lang="ts">
+type PillTone =
+	| 'neutral'
+	| 'accent'
+	| 'ok'
+	| 'warn'
+	| 'danger'
+	| 'info'
+	| 'asleep'
+
+withDefaults(
+	defineProps<{
+		tone?: PillTone
+		size?: 'sm' | 'md'
+	}>(),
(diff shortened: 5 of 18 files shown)


robertsLando (commented):
Code looks good overall, as said I would not merge this into main but I would create an `ui-next` branch so we can create beta releases of this. I think also you can drop node 20 in ci and add 24 in case that's the problem with the failing ci

src/components/dashboard/atoms/ZwButton.vue
```diff
@@ -0,0 +1,148 @@
...
+
+.zw-btn--primary:hover:not([data-disabled='true']) {
+	background: var(--zw-accent-dark);
+}
+
+/* ── outline ─────────────────────────────────────────────────── */
+.zw-btn--outline {
+	background: var(--zw-card);
+	border: 1px solid var(--zw-line);
+	color: rgba(0, 0, 0, 0.78);
```
robertsLando: **Dark-mode bug.** `rgba(0, 0, 0, 0.78)` won't flip in dark theme — the whole point of the V0 channel variables is that on-surface flips to white. Should be `var(--zw-fg)` (already alpha-composed against `--v0-on-surface`).

Same file has several other hardcoded literals that bypass the palette:
- L91 `box-shadow: 0 1px 2px rgba(25, 118, 210, 0.3);` → `rgba(var(--v0-primary), 0.3)`
- L112 `border: 1px solid rgba(229, 57, 53, 0.3);` → `rgba(var(--v0-error), 0.3)`
- L113 `color: #c62828;` → `rgb(var(--v0-error-darken-1))`
- L117 hover `rgba(255, 235, 238, 0.7)` — derive from `--v0-error`
- L131 ghost hover `rgba(0, 0, 0, 0.04)` → `rgba(var(--v0-on-surface), 0.04)`

Without these, changing `palette.ts` won't actually change the rendered colors.

src/components/dashboard/atoms/ZwChip.vue
```diff
@@ -0,0 +1,59 @@
...
+
+.zw-chip--neutral {
+	background: var(--zw-chip);
+	color: var(--zw-fg);
+}
+
+.zw-chip--accent {
+	background: var(--zw-accent-soft);
+	color: var(--zw-accent-dark);
+	border: 1px solid rgba(25, 118, 210, 0.3);
```
robertsLando: Three hardcoded color literals that bypass the palette:
- L37 `rgba(25, 118, 210, 0.3)` → `rgba(var(--v0-primary), 0.3)`
- L41 `rgba(67, 160, 71, 0.12)` → `rgba(var(--v0-success), 0.12)`
- L56 `rgba(154, 141, 183, 0.18)` → `rgba(var(--v0-asleep), 0.18)`

If `palette.ts` is the single source of truth, these need to derive from it — otherwise palette changes silently desync.

src/components/dashboard/atoms/ZwPill.vue
```diff
@@ -0,0 +1,83 @@
...
+	color: var(--zw-chip-fg);
+}
+
+.zw-pill--accent {
+	background: var(--zw-accent-soft);
+	color: var(--zw-accent-dark);
+}
+
+.zw-pill--ok {
+	background: rgba(67, 160, 71, 0.12);
```
robertsLando: Same pattern as ZwChip — hardcoded literals for ok/asleep that should flow through V0 channels:
- L60 `rgba(67, 160, 71, 0.12)` → `rgba(var(--v0-success), 0.12)`
- L80 `rgba(154, 141, 183, 0.18)` → `rgba(var(--v0-asleep), 0.18)`

Related: see comment in `tokens.css` about the unused `.zw-tone-*` classes — they were probably meant to absorb exactly this duplication.

src/components/dashboard/atoms/ZwStatusDot.vue
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.zw-status-dot--alive {
+	background: var(--zw-ok);
+}
+
+.zw-status-dot--awake {
+	background: var(--zw-ok);
+	/* Static halo — alpha 0x33 = 20% — matches direction-b.jsx. */
+	box-shadow: 0 0 0 3px rgba(67, 160, 71, 0.2);
```
robertsLando: `rgba(67, 160, 71, 0.2)` should derive from `--v0-success`: `rgba(var(--v0-success), 0.2)`. The comment even calls out the alpha but pins the RGB.

src/components/dashboard/atoms/ZwToggleMenu.vue
```diff
@@ -0,0 +1,239 @@
...
+	cursor: pointer;
+	transition:
+		background 0.12s,
+		border-color 0.12s;
+}
+
+.zw-tm__trigger--open,
+.zw-tm__trigger[data-open] {
+	background: var(--zw-accent-soft);
+	border-color: rgba(25, 118, 210, 0.4);
```
robertsLando: L130 `rgba(25, 118, 210, 0.4)` and L170 `rgba(0, 0, 0, 0.08)` are hardcoded — should be `rgba(var(--v0-primary), 0.4)` and `rgba(var(--v0-on-surface), 0.08)` (or an existing `--zw-line*` token).

Also: L9 `triggerLabel ?? 'Filter'` hardcodes English. If atoms are intentionally not localised at this layer (deferring to consumer-passed labels), document it; otherwise route through vue-i18n.

src/components/dashboard/atoms/ZwUpdateNotifier.vue
```diff
@@ -0,0 +1,99 @@
...
+
+<style>
+/* Unscoped — V0 primitives set inheritAttrs:false; .zw-update namespace
+   is unique to this atom. */
+.zw-update {
+	display: inline-flex;
+	align-items: center;
+	gap: 8px;
+	padding: 6px 8px;
+	border: 1px solid rgba(25, 118, 210, 0.3);
```
robertsLando: L46 `rgba(25, 118, 210, 0.3)` and L56 hover `rgba(227, 242, 253, 0.85)` should derive from `--v0-primary` so dark mode and palette changes propagate.

robertsLando (commented):
Foundation reads cleanly — palette as single source of truth, BEM-namespaced atoms, well-documented V0 workarounds. A few items to address before merge:

**Defects (worth blocking on):**
- Hardcoded color literals across most atoms bypass `palette.ts` and break dark mode in several places (notably `ZwButton` outline/destructive use literal black/`#c62828`).
- `ZwTransientReadout` displays a 30% "placeholder" as if it were real progress.
- `VSlider` default `trackColor: 'line'` references a color not registered in the palette.

**Cleanup:**
- `.zw-tone-*` classes in `tokens.css` are declared but no atom consumes them — the comment claims `ZwPill`/`ZwChip` read them.
- Tautological aliases `--zw-bg2`, `--zw-chip`.
- Alpha dep `@vuetify/v0@^1.0.0-alpha.5` with `^` allows breaking alphas to land automatically.
- Unrelated blank-line removal in `index.html`.
- `src/lib/icons.ts` aliases `Sun` and `Thermometer` to two names each — fine but worth a comment.

**Test coverage:** no tests added for any of the 12 atoms. `ZwBatteryMini` (color thresholds) and `ZwTransientReadout` (placeholder bug above) have real branching logic that's worth a smoke test.

Inline threads below for each item.

src/components/dashboard/atoms/ZwTransientReadout.vue
```diff
@@ -0,0 +1,136 @@
...
+	ota: DownloadIcon,
+	rebuild: RefreshIcon,
+	interview: InterviewIcon,
+} as const
+
+const iconComp = computed(() => TRANSIENT_ICON[props.transient.type])
+
+// 0.3 = indeterminate placeholder matching the handoff.
+const pct = computed(() =>
+	Math.round((props.transient.progress ?? 0.3) * 100),
```
robertsLando: **Misleading UI.** When `progress` is undefined this renders a definite-looking 30% bar and the `title` attribute reads e.g. `OTA · 30%` — users can't distinguish that from real progress.

For an indeterminate state either (a) use `Progress.Root`'s indeterminate mode and omit the `%` label, or (b) render the bar with no fill and a `—` label. The current placeholder is a UX trap, especially for OTA where users watch this number.

src/components/dashboard/atoms/ZwTransientReadout.vue
```diff
@@ -0,0 +1,136 @@
...
+
+.zw-tx__icon {
+	flex: 0 0 auto;
+	color: var(--zw-accent);
+}
+
+.zw-tx__bar {
+	position: relative;
+	height: 4px;
+	background: rgba(25, 118, 210, 0.18);
```
robertsLando: `rgba(25, 118, 210, 0.18)` should be `rgba(var(--v0-primary), 0.18)` so the bar background tracks palette changes.

src/components/dashboard/atoms/ZwToggle.vue
```diff
@@ -0,0 +1,108 @@
...
+.zw-toggle--md[data-state='checked'] {
+	background: var(--zw-accent);
+}
+
+.zw-toggle--md .zw-toggle__thumb {
+	width: 20px;
+	height: 20px;
+	top: 3px;
+	left: 3px;
+	box-shadow: 0 1px 3px rgba(0, 0, 0, 0.18);
```
robertsLando, line 77: L79 `rgba(0, 0, 0, 0.18)` and L102 `rgba(0, 0, 0, 0.2)` are hardcoded thumb shadows — in dark mode the white thumb on the dark track loses contrast because the drop-shadow stays black-on-black. Either use `rgba(var(--v0-on-surface), …)` (counter-intuitive but correct: shadow flips with theme), or override per-theme.

src/components/dashboard/atoms/ZwSlider.vue
```diff
@@ -0,0 +1,121 @@
...
+   `left: <pct>%` so we translate(-50%) to center the disc on the value. */
+.zw-slider__thumb {
+	position: absolute;
+	top: 50%;
+	width: 14px;
+	height: 14px;
+	border-radius: 50%;
+	background: var(--zw-accent);
+	transform: translate(-50%, -50%);
+	box-shadow: 0 1px 2px rgba(0, 0, 0, 0.2);
```
robertsLando: Three `rgba(0, 0, 0, 0.2)` thumb shadows (L98, L106, L113, L119). The halo at L105/112/118 correctly uses `var(--v0-primary)` — the rest-state shadow should follow the same pattern via `--v0-on-surface` so it doesn't disappear in dark mode.

src/assets/css/tokens.css
```diff
@@ -0,0 +1,225 @@
...
+	--zw-warning: rgb(var(--v0-warning));
+	--zw-warn-soft: rgb(var(--v0-warn-soft));
+	--zw-danger: rgb(var(--v0-error));
+	--zw-danger-soft: rgb(var(--v0-danger-soft));
+	--zw-asleep: rgb(var(--v0-asleep));
+
+	/* ── Color · neutral ─────────────────────────────────────── */
+	--zw-bg: rgb(var(--v0-background));
+	--zw-bg-soft: rgb(var(--v0-background-soft));
+	--zw-bg2: var(--zw-bg);
```
robertsLando: Tautological aliases — these resolve to their own source value and add no indirection:
- L33 `--zw-bg2: var(--zw-bg);`
- L45 `--zw-chip: var(--zw-chip-bg);`

If they're future-proofing seams, a comment would help; otherwise drop them.

src/assets/css/tokens.css
```diff
@@ -0,0 +1,225 @@
...
+	--zw-e8: 0 6px 20px rgba(var(--v0-on-surface), 0.16),
+		0 2px 4px rgba(var(--v0-on-surface), 0.08);
+	--zw-e-drawer: -12px 0 32px rgba(var(--v0-on-surface), 0.12);
+}
+
+/* ── Tone palette ────────────────────────────────────────────── */
+/* Each tone sets a (--tone-bg, --tone-fg) pair that ZwPill / ZwChip
+   read in their base rules. Keeps the seven-row tone table in one
+   place instead of repeating it per atom. */
+.zw-tone-neutral {
```
robertsLando, line 99: **Dead code.** The `.zw-tone-{neutral,accent,ok,warn,danger,info,asleep}` block (L102–139) defines `(--tone-bg, --tone-fg)` pairs with a comment saying "ZwPill / ZwChip read in their base rules." They don't — both atoms inline their own per-tone classes (`zw-pill--ok` etc.) with literal colors.

Either wire `ZwPill`/`ZwChip` to apply `.zw-tone-*` and read `--tone-bg`/`--tone-fg` (the DRY pattern the comment promises, and which would resolve several of the hardcoded-color comments above in one shot), or delete this block.

src/plugins/vuetify.js
```diff
@@ -19,47 +30,136 @@ export default createVuetify({
...
+		},
+		VSwitch: {
+			color: 'primary',
+			inset: true,
+			density: 'compact',
+			hideDetails: true,
+		},
+		VSlider: {
+			color: 'primary',
+			trackColor: 'line',
```
robertsLando: `trackColor: 'line'` references a theme color that doesn't exist in `palette.ts` (the `--zw-line*` CSS tokens aren't registered as Vuetify theme entries). Vuetify will emit `var(--v-theme-line)` and resolve to nothing.

Either add `line` to both palettes in `palette.ts`, or pick an existing key (e.g. `'on-surface'` with appropriate opacity).

src/lib/popover-fallback.ts
```diff
@@ -0,0 +1,100 @@
...
+import {
+	computePosition,
+	autoUpdate,
+	flip,
+	shift,
+	offset,
+} from '@floating-ui/dom'
+
+interface FallbackPositionOptions {
+	open: { value: boolean }
```
robertsLando: Nit: `open: { value: boolean }` is morally a `Ref<boolean>` — for consistency with `contentId: MaybeRefOrGetter<string>` on the next line and to give call sites proper inference, prefer `import type { Ref } from 'vue'` and `open: Ref<boolean>`.

package.json
```diff
@@ -77,7 +77,12 @@
   ],
   "dependencies": {
     "@esm2cjs/escape-string-regexp": "^5.0.0",
+    "@fontsource/roboto": "^5.2.10",
+    "@fontsource/roboto-mono": "^5.2.9",
+    "@floating-ui/dom": "^1.7.6",
     "@kvaster/zwavejs-prom": "^0.0.3",
+    "@vuetify/v0": "^1.0.0-alpha.5",
```
robertsLando: `^1.0.0-alpha.5` allows new alphas to land automatically on the next `npm install`. Several atoms in this PR encode workarounds for v0-specific behaviour (the `position-area: none !important` hack, the `inheritAttrs: false` / scoped-styles caveat repeated across files) — a breaking alpha could silently invalidate any of them.

Pin to `~1.0.0-alpha.5` or exact until v0 leaves alpha.

src/components/dashboard/atoms/ZwSearchInput.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+	<div class="zw-search">
+		<SearchIcon class="zw-search__icon" :size="ICON_SIZE.button" />
+		<input
+			type="search"
+			class="zw-search__input"
+			:value="modelValue"
+			:placeholder="placeholder"
+			:aria-label="ariaLabel || 'Search devices'"
```
robertsLando: Nit: `'Search devices'` fallback hardcodes English. Either require callers to always pass `ariaLabel` (drop the fallback) or route through vue-i18n. Same concern as the `'Filter'` fallback in `ZwToggleMenu`.

index.html
```diff
@@ -28,7 +28,6 @@
 		<meta name="theme-color" content="#ffffff" />
```
robertsLando, line 28: Unrelated change (removes a blank line). Per the project's surgical-changes rule, drop from this PR.

src/components/dashboard/atoms/ZwButton.vue
AlCalzone: Good catch with the hardcoded colors. I haven't made the design schema for dark mode yet, this is coming in a later PR.

robertsLando (approved):

package.json
```diff
@@ -77,7 +77,12 @@
   ],
   "dependencies": {
     "@esm2cjs/escape-string-regexp": "^5.0.0",
+    "@fontsource/roboto": "^5.2.10",
+    "@fontsource/roboto-mono": "^5.2.9",
+    "@floating-ui/dom": "^1.7.6",
     "@kvaster/zwavejs-prom": "^0.0.3",
+    "@vuetify/v0": "~1.0.0-alpha.5",
+    "@lucide/vue": "^1.16.0",
```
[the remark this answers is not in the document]
AlCalzone: This is tracked at https://github.com/orgs/zwave-js/projects/7/views/1 and will be done in a followup PR

src/plugins/vuetify.js
```diff
@@ -19,47 +30,129 @@ export default createVuetify({
...
+		// Align with CardsBody grid math: 480 / 760 / 1100 / 1380 cutoffs
+		// yield 1/2/3/4/5 card columns without conditional CSS.
+		thresholds: {
+			xs: 480,
+			sm: 760,
+			md: 1100,
+			lg: 1380,
+			xl: 1920,
 		},
 	},
```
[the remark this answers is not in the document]
AlCalzone: This was done on purpose

src/components/dashboard/atoms/ZwUpdateNotifier.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+	<Button.Root
+		class="zw-update"
+		:class="{ 'zw-update--compact': compact }"
+		:aria-label="`Update available: ${current} → ${available}`"
```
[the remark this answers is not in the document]
AlCalzone: V0's `Button.Root` defaults to `as="button"` and already emits the type itself — its rendered element gets `type: as === "button" ? "button" : undefined`. So there's no implicit `submit` to guard against here, and an explicit `type="button"` would just duplicate what the primitive does. Leaving as-is.

src/plugins/vuetify0.ts
```diff
@@ -0,0 +1,36 @@
...
+//
+// V0 is the headless successor to Vuetify 3 ("Vuetify 5 will be built on
+// v0"). New dashboard components target V0 primitives directly; Vuetify 3
+// remains installed only to host legacy routes during the rework, and is
+// removed once every legacy page has migrated.
+
+import { createThemePlugin, V0StyleSheetThemeAdapter } from '@vuetify/v0'
+import type { App } from 'vue'
+
+import { lightColors, darkColors } from './palette.ts'
```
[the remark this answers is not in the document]
AlCalzone: This is intentional rather than inconsistent. The repo's ESLint config enforces `import/extensions: ['error', 'always', { ignorePackages: true }]` for `**/*.ts` / `**/*.tsx`, so local imports in TypeScript files must carry the extension. `vuetify.js` omits it only because it's a `.js` file, which that override doesn't cover — dropping `.ts` here actually fails `npm run lint`. Keeping the extension.

src/components/dashboard/atoms/ZwToggleMenu.vue
```diff
@@ -0,0 +1,237 @@
...
+import { computed, ref, useId } from 'vue'
+import type { Component } from 'vue'
+import { Popover } from '@vuetify/v0'
+import {
+	CheckIcon,
+	ChevronDownIcon,
+	FilterIcon,
+	ICON_SIZE,
+} from '@/lib/icons'
+import { usePopoverFallback } from '@/lib/popover-fallback.ts'
```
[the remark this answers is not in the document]
AlCalzone: Same as the `vuetify0.ts` thread: the dashboard's TypeScript imports intentionally carry `.ts` extensions (the repo enforces `import/extensions: 'always'` for TS files), so this follows the project convention rather than diverging from it. Keeping the extension.

src/components/dashboard/atoms/ZwToggleMenu.vue
```diff
@@ -0,0 +1,237 @@
+<template>
+	<Popover.Root v-model="open" :id="contentId">
+		<Popover.Activator as="button" class="zw-tm__trigger zw-focus-ring">
```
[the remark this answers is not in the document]
AlCalzone: `Popover.Activator as="button"` is in the same boat as the `Button.Root` threads: V0 sets `type: as === "button" ? "button" : undefined` on the rendered element, so the activator already carries `type="button"` and won't default to `submit`. Leaving as-is.

src/components/dashboard/atoms/ZwButton.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+	<Button.Root
+		class="zw-btn zw-focus-ring"
+		:class="[`zw-btn--${variant}`, `zw-btn--${size}`]"
+		:disabled="disabled"
+		@click="emit('click', $event)"
+	>
```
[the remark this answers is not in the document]
AlCalzone: `Button.Root` defaults to `as="button"` and V0 already emits `type: as === "button" ? "button" : undefined` on the underlying element, so it does not fall back to `submit` inside a form. An explicit `type="button"` would be redundant. Leaving as-is.

src/components/dashboard/atoms/ZwActivityReadout.vue
```diff
@@ -0,0 +1,175 @@
+<template>
+	<span
+		class="zw-tx"
+		:class="[
+			`zw-tx--${variant}`,
```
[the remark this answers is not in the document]
AlCalzone: renamed

src/lib/popover-fallback.ts
```diff
@@ -0,0 +1,110 @@
...
+		cleanup = null
+	}
+
+	watch(
+		() => open.value,
+		(isOpen) => {
+			stopTracking()
+			if (isOpen) startTracking()
+		},
+	)
```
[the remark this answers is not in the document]
AlCalzone: fixed

src/components/dashboard/atoms/ZwUpdateNotifier.vue
```diff
@@ -0,0 +1,97 @@
...
+		@click="emit('click')"
+	>
+		<DownloadIcon :size="compact ? ICON_SIZE.button : ICON_SIZE.update" />
```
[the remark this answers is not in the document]
AlCalzone: fixed

robertsLando (approved):

PR #4648, opened 2026-05-25
refactor(build): improve Docker layer and GHA cache reuse

lightswitch05:
## Why

Docker layers were not structured or cached in a way that allowed reuse across builds. Previously unchanged layers were being re-built and re-pulled unnecessarily:

- \`node_modules\`, \`dist\`, \`dev-dist\`, and \`server\` were not in \`.dockerignore\`, so the build context included ~1 GB of generated/installed content — none of which Docker could use, but all of which had to be uploaded on every build.
- JS/TS build artifacts (\`dist/\`, \`server/\`) were rebuilt on every target platform (\`amd64\`, \`arm64\`, \`arm/v7\`, \`arm/v6\`) even though they are fully platform-agnostic — only native addons like \`@serialport/bindings-cpp\` need per-platform compilation. Running Vite and \`tsc\` under QEMU on \`arm64\`/\`arm/v7\` was also significantly slower than a native \`amd64\` build.
- All \`COPY\` instructions were in a single layer, so any source change invalidated the \`npm ci\` layer even when \`package.json\` hadn't changed.
- CI workflows ran \`npm ci\` and \`npm run build\` before invoking Docker, but Docker couldn't use those artifacts and rebuilt everything anyway — making the pre-build steps wasted work that also broke layer caching.
- Both \`zwave-js-ui\` and \`zwavejs2mqtt\` used the same GHA cache scope, so the \`zwavejs2mqtt\` build silently evicted the \`zwave-js-ui\` cache entry, forcing a full re-pull on the next run.
- \`apk\`/\`apt-get\` packages were re-downloaded on every layer rebuild, with no protection against concurrent writes during multi-platform builds.
- \`COPY --from=build-zui\` in the runtime stage was coupled to the base image layer, so a base image security patch forced the built app layer to be re-pulled even though the app itself hadn't changed.

## What

- **\`.dockerignore\`**: exclude \`node_modules\`, \`dist\`, \`dev-dist\`, and \`server\` to shrink build context from ~1 GB to ~20 MB.
- **Dockerfiles** (\`alpine\` + \`debian\`): add \`# syntax=docker/dockerfile:1\` to both files for consistent BuildKit frontend behavior with \`RUN --mount\` and \`COPY --link\` instructions. Split into two build stages. \`build-js\` (pinned to \`$BUILDPLATFORM\`, resolving to \`linux/amd64\` in CI) installs all dependencies and builds JS/TS artifacts (\`dist/\`, \`server/\`) once natively — these outputs are platform-agnostic. \`build-native\` (runs per target platform) installs production-only dependencies, rebuilds native addons from source (e.g., \`@serialport/bindings-cpp\`), and receives the JS artifacts via \`COPY --link --from=build-js\`. Within each stage, \`package.json\`, \`package-lock.json\` are copied before source so the \`npm ci\` layer is reused across source-only changes. \`--mount=type=cache,sharing=locked\` is applied to all \`npm ci\` commands as well as \`apk\`/\`apt-get\` to cache downloaded packages across layer rebuilds and prevent concurrent write conflicts during multi-platform builds.
- **CI workflows**: remove pre-build \`npm ci\` + \`npm run build\` steps; Docker now owns the full build so its layer cache is authoritative.
- **CI workflows**: add per-image GHA cache scopes (\`zwave-js-ui\` vs \`zwavejs2mqtt\`) in both \`docker-release.yml\` and \`docker-release-test.yml\` so builds no longer evict each other's cached layers; \`zwavejs2mqtt\` also reads from the \`zwave-js-ui\` scope to warm its cache on a cold start.
- **Dockerfiles**: add \`--link\` to all \`COPY\` instructions for independent layer reuse, most notably \`COPY --link --from=build-native\` in the runtime stage so a base image security patch no longer invalidates the built app layer.
- **\`api/lib/MqttClient.ts\`**: replace \`native-url\` with \`node:url\` to fix a TypeScript build error exposed by the clean Docker build. The old CI pre-built on the host runner and Docker skipped the build step when \`dist/\` existed, masking a latent \`TS2306\` error — \`native-url\`'s type definition uses an ambient \`declare module 'url'\` block rather than a proper module export. Since \`MqttClient.ts\` is backend-only code, \`node:url\` (available since Node.js 0.x, well within the \`>= 20.19\` requirement) is the correct import.
- **\`patch-package\` and \`patches/\`**: the \`native-url\` default-import patch is no longer needed since \`MqttClient.ts\` switched to \`node:url\`. With no patches remaining, \`patch-package\` was removed from \`dependencies\` and \`postinstall\`, and the \`COPY patches/\` instruction was dropped from both Dockerfiles.

## Testing

I ran the \`docker-release\` build in a [test PR](https://github.com/lightswitch05/zwave-js-ui/pull/8). First, I deleted all cache (\`gh cache delete --all --repo lightswitch05/zwave-js-ui\`) to ensure a fresh run. The [first run](https://github.com/lightswitch05/zwave-js-ui/actions/runs/26454761341/job/77904085169?pr=8) was pretty slow - about 21 minutes. The \`zwavejs2mqtt\` reported 83% build cache. The \`zwave-js-ui\` image only had 6% cache use. I ran the build a [second time](https://github.com/lightswitch05/zwave-js-ui/actions/runs/26454761341?pr=8) and both images reported 83% cached and an entire run time of just under 1 minute. This isn't a perfect test, as any _real_ release is going to include various changes, but I believe the overall cache and layer re-use will be improved.


.github/workflows/docker-release-test.yml
```diff
@@ -59,20 +54,15 @@ jobs:
             org.opencontainers.image.url=https://zwave-js.github.io/zwave-js-ui/#/
             maintainer=robertsLando
 
-      - name: Pre-build frontend and backend files
-        run: |
-          npm ci
-          npm run build
```
lightswitch05, line 65: There is a tradeoff with removing this, but I believe with the improved docker cache re-use, its ok to let docker manage this cache on its own

docker/Dockerfile
```diff
@@ -6,16 +6,17 @@ FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION} AS node
...
+    apk add \
     openssl \
     libusb \
     tzdata \
     eudev
 
-COPY --from=node /usr/lib /usr/lib
-COPY --from=node /usr/local/lib /usr/local/lib
-COPY --from=node /usr/local/include /usr/local/include
-COPY --from=node /usr/local/bin /usr/local/bin
```
lightswitch05, line 18: The `--link` flag allows you to preserve a layer, even when the underlying layers have changed. You can read more [in the docs](https://docs.docker.com/reference/dockerfile/#copy---link).

docker/Dockerfile
```diff
@@ -24,23 +25,22 @@ WORKDIR /usr/src/app
...
 FROM base AS build-zui
 
-RUN \
-    apk add --no-cache --virtual .build-dependencies \
+RUN --mount=type=cache,target=/var/cache/apk,sharing=locked \
+    apk add --virtual .build-dependencies \
         jq \
         build-base \
         linux-headers \
         python3-dev
```
lightswitch05, line 35: Cache mounts won't help any with with CI builds, but it has the same effect as `--no-cache` while significantly improving any local build times

docker/Dockerfile
```diff
@@ -24,23 +25,22 @@ WORKDIR /usr/src/app
...
-COPY . .
-
-# if node_modules does not exist, install deps, if dist is missing, install also dev deps
-RUN [ -d 'node_modules' ] && echo "Skipping install" || npm ci $([ -d 'dist' ] && echo '--omit=dev' || echo '')
-
+# Cache dependency layer independently of source changes
+COPY --link package.json package-lock.json patches/ ./
+RUN npm ci
 # Fix issue with serialport bindings #2349
 RUN npm_config_build_from_source=true npm rebuild @serialport/bindings-cpp
```
lightswitch05: I believe this rebuild is no longer necessary - since each arch is already downloading its own dependencies (and reusing its own cache). But I'm not sure how to test it, so I left it alone.

.github/workflows/docker-release-test.yml
robertsLando: I removed this on purpose because building ui inside container was so painful due to dev deps, that was very flaky and I would like to avoid having to deal with that again, why not keeping build of frontend external?

docker/Dockerfile
robertsLando: Didn't know about that! Thanks 🙏🏼

robertsLando (changes requested):
As said I would keep the ui build external to docker

docker/Dockerfile
```diff
@@ -6,16 +6,17 @@ FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION} AS node
 
 FROM alpine:${ALPINE_VERSION} AS base
 
-RUN apk add --no-cache \
+RUN --mount=type=cache,target=/var/cache/apk,sharing=locked \
+    apk add \
```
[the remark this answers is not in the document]
lightswitch05: Fixed in 116edb97 — added `# syntax=docker/dockerfile:1` as the first line.

docker/Dockerfile.debian
```diff
@@ -5,24 +5,21 @@ FROM node:22.20.0-bookworm AS build-zui
 
 WORKDIR /usr/src/app
 
-RUN apt-get update && apt-get install -y \
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt,sharing=locked \
+    apt-get update && apt-get install -y \
```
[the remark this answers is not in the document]
lightswitch05: Fixed in 116edb97 — added `# syntax=docker/dockerfile:1` as the first line.

.github/workflows/docker-release.yml
```diff
@@ -75,20 +71,15 @@ jobs:
             org.opencontainers.image.authors=Daniel Lando <daniel.sorridi@gmail.com>
             maintainer=robertsLando
 
-      - name: Pre-build frontend and backend files
-        run: |
-          npm ci
-          npm run build
-
       - name: build+push
         timeout-minutes: 120 # prevents this to run indeterminatly
```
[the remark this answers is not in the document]
lightswitch05: Fixed in 116edb97.

.github/workflows/docker-release-test.yml
lightswitch05: Thanks for the review @robertsLando! This is a good point about the UI build and it not needing to be platform specific. When the build happens in docker, we get great cache and layer re-use results, which is what I was focused on. I think there is a solution that solves both issues. Using a build stage like `FROM --platform=$BUILDPLATFORM base AS build-js`, that stage only gets built once, regardless of the target platform. Then the individual platform-specific builds can all pull from it. This keeps all the cache within docker, but hopefully bypasses any painful duplication. What do you think?

This PR ended up being larger than I had intended. I'm happy to try to break it up into smaller individual PRs if you like. If you still want to move the build out of docker completely, then I'm happy to do that as well, just wanted to explore this other option first.

Thank you for this project!

robertsLando: I think that's a nice idea! :) Thanks to you for your contribution 🙏🏼

api/lib/MqttClient.ts
```diff
@@ -20,7 +20,7 @@ import { storeDir } from '../config/app.ts'
 import { ensureDir } from './utils.ts'
 import { Manager } from 'mqtt-jsonl-store'
 import { join } from 'node:path'
-import url from 'native-url'
+import url from 'node:url'
```
robertsLando, line 23: why this change?

.github/workflows/docker-release-test.yml
robertsLando: No worries about splitting the pr. it's ok to keep it in one

api/lib/MqttClient.ts
robertsLando, line 23: Since `MqttClient.ts` no longer imports `native-url` (and the remaining frontend usage in `src/views/Settings.vue` is a named `import { parse }`, not a default import), the `patches/native-url+0.3.4.patch` file is now dead weight — it exists solely to add `export default url` to `native-url`'s type declaration so the old `import url from 'native-url'` would type-check. Could you delete `patches/native-url+0.3.4.patch` in this same PR? Otherwise it becomes orphan tech debt that'll be hard to associate with this change later.

lightswitch05: There was build issue with `native-url` and one of the specific platform builds:

> api/lib/MqttClient.ts(23,17): error TS2306: File '/usr/src/app/node_modules/native-url/third_party/url.d.ts' is not a module.

Now that the builds are sharing some resources, I'll see if it can be removed.

lightswitch05: Coming back to this after reading the rest of the comments - maybe I had just messed up the patching, but it seems making this switch is still an overall improvement and the patch logic can be removed.

lightswitch05: Removed! 🥳

robertsLando (approved):
LGTM. Thanks! :)

PR #4649, opened 2026-05-27
fix(ui): reduce nodes-table re-render pressure and clear inclusion ghosts

robertsLando:
## Summary

On meshes with many chatty nodes the control panel becomes laggy and visibly broken — see discussion #4639 (94-node setup, logs attached there). Three distinct UI symptoms, two of them sharing a root cause:

1. **The expanded config panel collapses mid-edit.** Every socket-driven node update was both `Object.assign`ing into the existing node object *and* `splice`ing it back into `this.nodes` — the splice was an array-level reactivity pulse that invalidated v-data-table's row identity and dropped the expanded slot.
2. **Hover tooltips on the table columns orphan and stack up** until something forces a global refresh. `RichValue.vue` used the `v-tooltip` directive form, which leaks overlays when the activator unmounts during a burst of re-renders. The component form (`<v-tooltip activator="parent">`) ties the overlay to its own mount lifecycle and is properly torn down.
3. **Failed inclusions leave permanent `ProtocolInfo` ghost entries.** `_onNodeFound` always pushes a partial node to the UI; `_onInclusionFailed` cleared `tmpNode`/`isReplacing` but never removed the ghost — and zwave-js does not consistently emit `node removed` when inclusion aborts before interview completes.

The reporter's logs also showed that **~50% of all `valueUpdated` socket emits were no-ops** (`X => X` — e.g. `0 => 0`, `true => true`, `100 => 100`). Suppressing those alone halves the re-render pressure on the frontend, which incidentally also widens the orphan-tooltip and collapse windows by a lot.

### `api/lib/ZwaveClient.ts`

- `emitValueChanged` now skips the socket broadcast when `changed === false`. The local `'valueChanged'` event is still emitted unchanged so the MQTT gateway (`Gateway._onValueChanged`) sees every refresh — only the UI socket traffic is gated. `valueId.lastUpdate` is still updated server-side so the next real change carries a fresh timestamp.
- Stateless value reset (`statelessTimeouts`) now passes `changed=true` to `emitValueChanged` because the reset (X → undefined) is itself a transition the UI must see.
- `_onInclusionFailed` removes any node that came in via `node found` but never reached `node added`. Tracking lives in a new `_pendingInclusionNodeIds: Set<number>`, populated in `_onNodeFound`, cleared in `_onNodeAdded` and `_removeNode`.

### `src/stores/base.js`

- `updateNode` no longer `splice`s the merged node back into `this.nodes`. The `Object.assign` mutation propagates via Vue reactivity without invalidating the v-data-table's row identity.

### `src/components/nodes-table/RichValue.vue`

- Replaced the `v-tooltip:bottom` directive with `<v-tooltip activator="parent" :open-delay="300">`. The 300ms delay also keeps transient mouse-overs from opening a tooltip at all during update bursts.

## Notes / non-goals

- MQTT/HASS behavior is unchanged: the `'valueChanged'` local event still fires on every update (Gateway has its own filtering/publishing logic).
- Other `v-tooltip` directive uses elsewhere in the codebase are not touched in this PR — they aren't on hot paths in chatty mesh setups. Worth converting later if the same symptoms appear elsewhere.
- No batching/debouncing of socket events was added. With the no-op suppression alone the reported load drops by ~50%, and adding a coalescing window felt like over-reach for this fix. Easy follow-up if needed.

## Test plan

- [ ] Stand up the demo fleet (`npm run fake-stick:fleet`) and confirm the control panel still updates correctly on value change.
- [ ] Expand a node's Config panel, write a value via the `>` button; confirm the panel stays expanded after the round-trip.
- [ ] Hover repeatedly over Last Active / Status / Power on multiple rows; confirm tooltips don't stack and dismiss cleanly on mouse-out.
- [ ] Stateless button press: confirm UI shows the press and then clears after 1s.
- [ ] Force an inclusion failure (start inclusion, never present a device, wait for the 30s timeout) and confirm no ghost node remains in the table.
- [ ] MQTT gateway: confirm value updates including unchanged ones still publish.

Refs: #4639

### DIFF
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -768,6 +768,10 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 	private tmpNode: utils.DeepPartial<ZUINode>
 	// tells if a node replacement is in progress
 	private isReplacing = false
+	// node ids surfaced to the UI via `node found` that have not yet hit
+	// `node added` — if inclusion fails before the interview completes
+	// they're orphans stuck at ProtocolInfo and must be cleared (see #4639)
+	private _pendingInclusionNodeIds: Set<number> = new Set()
 
 	private hasUserCallbacks = false
 
@@ -2625,7 +2629,13 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 			this.getNode(valueId.nodeId)?.getValueTimestamp(valueId) ??
 			Date.now()
 
-		this.sendToSocket(socketEvents.valueUpdated, valueId)
+		// Skip the socket broadcast when the value didn't actually change.
+		// Chatty meshes can produce ~50% no-op value updates (see #4639) and
+		// each one triggers an unnecessary re-render of the nodes table.
+		// MQTT/HASS still get every event via the local 'valueChanged' below.
+		if (changed) {
+			this.sendToSocket(socketEvents.valueUpdated, valueId)
+		}
 
 		this.emit('valueChanged', valueId, node, changed)
 	}
@@ -5823,6 +5833,23 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		const message = 'Inclusion failed'
 		this.isReplacing = false
 		this.tmpNode = undefined
+
+		// Clear any ghost nodes the driver surfaced via `node found` but never
+		// promoted to `node added`. zwave-js doesn't always emit `node removed`
+		// for these, so they would stay stuck on ProtocolInfo in the UI forever.
+		for (const nodeId of this._pendingInclusionNodeIds) {
+			const node = this._nodes.get(nodeId)
+			if (node && !node.ready) {
+				this.logNode(
+					nodeId,
+					'info',
+					'Removing ghost node after failed inclusion',
+				)
+				this._removeNode(nodeId)
+			}
+		}
+		this._pendingInclusionNodeIds.clear()
+
 		this._updateControllerStatus(message)
 		this.emit('event', EventSource.CONTROLLER, 'inclusion failed')
 	}
@@ -5844,6 +5871,8 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 		if (this.driverReady) {
 			node = this._createNode(nodeId)
 			this.sendToSocket(socketEvents.nodeFound, { node })
+			// track until `node added` clears it; cleaned up on inclusion failure
+			this._pendingInclusionNodeIds.add(nodeId)
 		} else {
 			node = this._nodes.get(nodeId)
 		}
@@ -5860,6 +5889,8 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 	 */
 	private async _onNodeAdded(zwaveNode: ZWaveNode, result: InclusionResult) {
 		let node: ZUINode
+		// node made it past `node found` — no longer a ghost candidate
+		this._pendingInclusionNodeIds.delete(zwaveNode.id)
 		// the driver is ready so this node has been added on fly
 		if (this.driverReady) {
 			node = this._addNode(zwaveNode)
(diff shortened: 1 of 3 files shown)


api/lib/ZwaveClient.ts
```diff
@@ -5823,6 +5833,23 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
+			if (node && !node.ready) {
+				this.logNode(
+					nodeId,
+					'info',
+					'Removing ghost node after failed inclusion',
+				)
+				this._removeNode(nodeId)
+			}
+		}
+		this._pendingInclusionNodeIds.clear()
```
robertsLando, line 5851: @AlCalzone is this legit? I wonder if it should be zwave js driver emitting the node removed in case this happens?

src/stores/base.js
```diff
@@ -323,7 +323,10 @@ const useBaseStore = defineStore('base', {
 			}
 
 			if (index >= 0) {
-				this.nodes.splice(index, 1, n)
+				// node already exists: Object.assign above mutated this.nodes[index]
+				// in place, so Vue reactivity already propagated the per-field changes.
+				// Splicing would invalidate v-data-table's row tracking and drop the
+				// expanded-row state mid-edit (see #4639).
```
robertsLando: this shuuld be verified

robertsLando: Verified end-to-end in Chrome against the running app — Vue 3 deep reactivity propagates `Object.assign` mutations to the DOM without needing the splice. Test battery results (Pinia store called directly + DOM diffed):

| Test | Result |
|---|---|
| Partial updateNode mutates name+loc → DOM reflects change | ✅ `Front Door: Lock` → `Renamed Lock`, `Entrance` → `Garage` |
| Status flip Alive→Dead→Alive → status icon SVG path actually changes | ✅ path length 419 (Dead emoticon) → 115 (Alive check) |
| Battery value mutation via `updateValue` | ✅ stored value matches, DOM Power column updates |
| Full (non-partial) `updateNode` with new values payload | ✅ node array ref stays stable, values converted to array |
| Add new node via `updateNode` (push path still intact) | ✅ row appears, `nodesMap` updated |
| Remove node via `removeNode` | ✅ row gone, `nodesMap` consistent |
| `setValue` (the write-path setter) replaces single valueId | ✅ |
| 1000 partial `updateNode` calls in a tight loop | ✅ node ref stable, no leak, DOM still live |
| 500 idempotent `updateValue` calls with same payload | ✅ exactly one value entry, no duplicates |
| `nodesMap` integrity after all mutations | ✅ `mapSize === arrLen` |
| Real `>` send on a config value (round-trip) → panel stays expanded | ✅ (verified visually) |
| Expanded panel survives 200 mixed partial updates | ✅ (visually, panel remained open through bursts) |

The splice was a Vue 2 carry-over — in Vue 3, the `Object.assign(this.nodes[index], n)` above is enough; the proxy intercepts every property write and notifies dependents. The splice would just trigger an extra array-level invalidation that no one needed (and was the actual cause of the v-data-table dropping the expanded row).

Also fixed the empty-if-block in 9b48127 (Copilot's comment) by inverting to `if (index === undefined) push(...)`.

api/lib/ZwaveClient.ts
robertsLando: Worth asking @AlCalzone yes — ideally the driver should emit `node removed` for this scenario so we don't need a client-side safety net. From reading the driver code paths, `inclusion failed` fires when the controller has assigned a NodeID but the device never completed the secure exchange / interview, and in some of those branches the driver doesn't synthesize a `node removed` because, from its POV, the node was never fully added.

I kept the cleanup conservative — it only removes nodes that:

1. Were surfaced to the UI via `_onNodeFound` (added to `_pendingInclusionNodeIds`)
2. Have not had `_onNodeAdded` fire (still in the set)
3. Have `node.ready === false` (extra safety — never removes a node mid-interview if for some reason it did make it past `node added`)

So even if zwave-js does start emitting `node removed` for this case in the future, our cleanup is a no-op (the node is removed via `_removeNode` which deletes from `_pendingInclusionNodeIds` first). Happy to drop the safety net once a driver-side fix lands.

AlCalzone: That sequence of events is intended
- node found means there is a node wanting to be included
- inclusion failed means it was not included
- node removed implies the node was part of the network in the first place, which isn't the case here

AlCalzone: it might make sense to include the nodeId in the inclusion failed event though.

AlCalzone (approved):
Z-Wave portion looks valid

PR #4662, opened 2026-06-11
fix(gateway): apply numeric `postOperation` scaling directly

AlCalzone:
A value's `postOperation` is a single +, -, *, / against a literal number, so compute it with a small arithmetic helper instead of using `eval`. `_isValidOperation` now enforces that shape, and only finite numeric values are scaled; any other payload or malformed operation is returned untouched.


robertsLando (changes requested):
I would move the appy operation to utils and write some tests

robertsLando (dismissed):

robertsLando (approved):

PR #4664, opened 2026-06-11
fix(config): enforce a strong session secret

AlCalzone:
Until now we used a know default session "secret" to sign auth tokens. With this PR we instead generate a random secret when none is given or already persisted from a previous startup.


robertsLando (commented):
I think this change may invalidate some sessions but not a big deal 👍🏼

robertsLando (approved):

PR #4678, opened 2026-06-19
fix: prevent symlinks from escaping the store directory

AlCalzone:
The `PUT/POST` `/api/store-multi` endpoints and the backup-restore flow took user-supplied paths and applied only a string-prefix check, so an authenticated user could delete or ZIP-export files outside storeDir, and a restore archive could plant a symlink that escaped `storeDir` to write arbitrary files. The shared `getSafePath()` helper never resolved real paths, so symlinked components defeated the confinement.

With this PR, `getSafePath()` now normalizes paths relative to the `storeDir` and resolves the targets of symlinks, optionally throwing if a path ends up pointing outside the store.
Also, the restore flow now stages the ZIP file contents, and checks them for symlinks that would point outside the store, before restoring the archive.

### DIFF
--- a/api/app.ts
+++ b/api/app.ts
@@ -2014,7 +2014,6 @@ app.put(
 		try {
 			const files = req.body.files || []
 			for (const f of files) {
-				// lgtm [js/path-injection]
 				await rm(await getSafePath(f), { recursive: true, force: true })
 			}
 			res.json({ success: true })
--- a/api/app.ts
+++ b/api/app.ts
@@ -395,32 +395,28 @@ async function assertRealPathInStore(target: string) {
 }
 
 /**
- * Get the `path` param from a request. Throws if the path is not safe.
- * Performs a string confinement check and, when `resolveReal` is set,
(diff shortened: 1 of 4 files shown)


api/app.ts
```diff
@@ -360,22 +363,60 @@ async function getSnippets() {
...
+ * existing and not-yet-created paths.
  */
-function getSafePath(req: Request | string) {
-	let reqPath = typeof req === 'string' ? req : req.query.path
+async function assertRealPathInStore(target: string) {
+	let current = target
+
+	// Walk up until we hit an existing ancestor (the target itself may be new)
+
+	while (true) {
```
robertsLando, line 376: It would be better to guard this for recursive symlinks

AlCalzone: `realpath` already does this by throwing `ELOOP` when encountering cyclic symlinks.
The while loop is bounded by walking up the directory with `path.dirname` and checking `parent === current`

robertsLando: oh nice, didn't know that

PR #4691, opened 2026-06-24
feat(ui): implement more complex UI components

AlCalzone:
This PR builds the composite component layer on top of the atoms we added in #4647.
These are the reusable building blocks the dashboard shell (followup work) assembles into views.

Primary-value display system — a type-dispatched renderer for a device's main control/value:
- `ZwPrimaryDisplay.vue` (dispatcher) + `registry.ts` mapping each `PrimaryValueType` to a renderer
- For now there are 7 renderers under primary-display/: Toggle, Dim, Lock, Reading, State, Thermostat, Controller
- `ZwCompactPrimary.vue` — condensed variant for dense rows

Device list presentation
- `ZwDeviceCard.vue` — card-view device chrome
- `ZwDeviceRow.vue` + gridTemplateB.ts — table-view row sharing a single column-grid template

Node-details surfaces
- `ZwNodeDetailsBody.vue` — shared detail body (used by both drawer and inline expansion)
- `ZwDeviceDrawer.vue` — right-side overlay drawer; `ZwExpandedRow.vue` — inline row expansion wrapper
- `ZwPropTable.vue` (mono key-value table), `ZwStatsCard.vue` (label/value stats), `ZwSecurityPanel.vue` (S0/S2 key-grant panel)

Top-bar / toolbar pieces — `ZwAddDeviceSplitButton.vue` (split CTA), `ZwColumnsMenu.vue` (table column visibility)

Supporting
- `src/lib/dashboard-types.ts` — the `Device` / `PrimaryValue` / `Activity` / `DeviceAction` type contract these components consume
- `tsconfig.eslint.json` — adds `@/*` path-alias resolution so type-aware lint resolves the new imports

---

Sneak preview:
image

### DIFF
--- a/src/components/dashboard/components/ZwAddDeviceSplitButton.vue
+++ b/src/components/dashboard/components/ZwAddDeviceSplitButton.vue
@@ -58,7 +58,7 @@ import {
 } from '@/lib/icons'
 import { usePopoverFallback } from '@/lib/popover-fallback.ts'
 
-type Action = 'include' | 'replace' | 'exclude'
+type Action = 'include' | 'replace-failed' | 'exclude'
 
 defineProps<{
 	compact?: boolean
@@ -81,7 +81,7 @@ const MENU = [
 		destructive: false,
 	},
 	{
-		id: 'replace' as const,
+		id: 'replace-failed' as const,
 		icon: RefreshIcon,
 		title: 'Replace failed device',
 		desc: 'Swap an unresponsive node in place',
--- a/src/components/dashboard/components/ZwAddDeviceSplitButton.vue
+++ b/src/components/dashboard/components/ZwAddDeviceSplitButton.vue
@@ -19,11 +19,7 @@
 				<ChevronDownIcon :size="ICON_SIZE.button" />
 			</Popover.Activator>
 		</Button.Group>
-		<Popover.Content
-			as="div"
-			class="zw-asb__menu"
-			role="menu"
-		>
+		<Popover.Content as="div" class="zw-asb__menu" role="menu">
 			<button
 				v-for="item in MENU"
 				:key="item.id"
--- a/src/components/dashboard/components/ZwAddDeviceSplitButton.vue
+++ b/src/components/dashboard/components/ZwAddDeviceSplitButton.vue
@@ -1,0 +1,259 @@
+<template>
+	<Popover.Root v-model="open" :id="contentId">
+		<Button.Group class="zw-asb__group">
+			<Button.Root
+				class="zw-asb__face"
+				:class="{ 'zw-asb__face--compact': compact }"
+				@click="onFaceClick"
+			>
+				<AddIcon :size="ICON_SIZE.button" />
+				<span v-if="!compact">{{ wide ? 'Add device' : 'Add' }}</span>
+			</Button.Root>
+			<span class="zw-asb__divider" />
+			<Popover.Activator
+				as="button"
+				class="zw-asb__chev"
+				:class="{ 'zw-asb__chev--open': open }"
+				aria-haspopup="menu"
+			>
+				<ChevronDownIcon :size="ICON_SIZE.button" />
+			</Popover.Activator>
+		</Button.Group>
+		<Popover.Content
+			as="div"
+			class="zw-asb__menu"
+			role="menu"
+		>
+			<button
+				v-for="item in MENU"
+				:key="item.id"
+				type="button"
+				role="menuitem"
+				class="zw-asb__item"
+				:class="{ 'zw-asb__item--destructive': item.destructive }"
+				@click="onItemClick(item.id)"
+			>
+				<span
+					class="zw-asb__item-icon"
+					:class="{
+						'zw-asb__item-icon--destructive': item.destructive,
+					}"
+				>
+					<component :is="item.icon" :size="ICON_SIZE.button" />
+				</span>
+				<span class="zw-asb__item-text">
+					<span class="zw-asb__item-title">{{ item.title }}</span>
+					<span class="zw-asb__item-desc">{{ item.desc }}</span>
+				</span>
+			</button>
+		</Popover.Content>
+	</Popover.Root>
+</template>
+
+<script setup lang="ts">
+import { ref, useId } from 'vue'
+import { Button, Popover } from '@vuetify/v0'
+import {
+	AddIcon,
+	ChevronDownIcon,
+	ICON_SIZE,
+	RefreshIcon,
+	TrashIcon,
+} from '@/lib/icons'
+import { usePopoverFallback } from '@/lib/popover-fallback.ts'
+
+type Action = 'include' | 'replace' | 'exclude'
+
+defineProps<{
+	compact?: boolean
+	wide?: boolean
+}>()
+
+const emit = defineEmits<{ action: [Action] }>()
+
+const open = ref(false)
+const contentId = `zw-asb-${useId()}`
+
+usePopoverFallback({ open, contentId })
+
+const MENU = [
+	{
+		id: 'include' as const,
+		icon: AddIcon,
+		title: 'Include device',
+		desc: 'Add a new device to the network',
+		destructive: false,
+	},
+	{
+		id: 'replace' as const,
+		icon: RefreshIcon,
+		title: 'Replace failed device',
+		desc: 'Swap an unresponsive node in place',
+		destructive: false,
+	},
+	{
+		id: 'exclude' as const,
+		icon: TrashIcon,
+		title: 'Exclude device',
+		desc: 'Remove a device from the network',
+		destructive: true,
+	},
+]
+
+function onFaceClick(): void {
+	emit('action', 'include')
+}
+
+function onItemClick(id: Action): void {
+	emit('action', id)
+	open.value = false
+}
+</script>
+
+<style>
+/* Unscoped — V0 primitives set inheritAttrs:false; .zw-asb namespace is
+   unique to this component. */
+.zw-asb__group {
+	display: inline-flex;
+	border-radius: 6px;
+	overflow: hidden;
+	box-shadow: 0 1px 2px rgba(25, 118, 210, 0.3);
+}
+
+.zw-asb__face,
+.zw-asb__chev {
+	appearance: none;
+	border: none;
+	cursor: pointer;
+	background: var(--zw-accent);
+	color: #fff;
+	font-family: var(--zw-font);
+	font-weight: 600;
+	letter-spacing: 0.4px;
+	text-transform: uppercase;
+	display: inline-flex;
+	align-items: center;
+	gap: 6px;
+	transition: background 0.12s;
+}
+
+.zw-asb__face {
+	font-size: 12px;
+	padding: 7px 14px;
+}
+
+.zw-asb__face--compact {
+	padding: 7px 9px;
+}
+
+.zw-asb__face:hover,
+.zw-asb__chev:hover {
+	background: var(--zw-accent-dark);
+}
+
+.zw-asb__divider {
+	width: 1px;
+	background: rgba(255, 255, 255, 0.28);
+}
+
+.zw-asb__chev {
+	padding: 7px 10px;
+}
+
+.zw-asb__chev--open,
+.zw-asb__chev[data-open] {
+	background: var(--zw-accent-dark);
+}
+
+.zw-asb__chev--open svg,
+.zw-asb__chev[data-open] svg {
+	transform: rotate(180deg);
+	transition: transform 0.15s;
+}
+
+/* V0 hard-codes `position-area: bottom` inline on the popover and Vue
+   re-applies it on every render, so neutralise via stylesheet so Floating
+   UI's manual top/left writes drive the actual position. See ZwToggleMenu. */
+.zw-asb__menu {
+	position-area: none !important;
+	position-anchor: auto !important;
+	min-width: 260px;
+	background: var(--zw-card);
+	border: 1px solid var(--zw-line);
+	border-radius: 6px;
+	box-shadow: var(--zw-e8);
+	padding: 4px;
+	animation: zw-fade-in 0.12s;
+}
+
+.zw-asb__menu::backdrop {
+	display: none;
+}
+
+.zw-asb__item {
+	display: flex;
+	align-items: center;
+	gap: 10px;
+	width: 100%;
+	background: transparent;
+	border: none;
+	cursor: pointer;
+	padding: 8px 10px;
+	border-radius: 4px;
+	text-align: left;
+	color: var(--zw-fg);
+	font-family: var(--zw-font);
+	font-size: 13px;
+}
+
+.zw-asb__item:hover {
+	background: var(--zw-row-hover);
+}
+
+.zw-asb__item--destructive {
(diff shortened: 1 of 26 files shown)


src/components/dashboard/components/ZwColumnsMenu.vue
```diff
@@ -0,0 +1,36 @@
...
+	'update:modelValue': [ToggleableCol[]]
+}>()
+
+const TOGGLEABLE_COLS = [
+	{ id: 'activity', label: 'Activity' },
+	{ id: 'location', label: 'Location' },
+	{ id: 'value', label: 'State / Value' },
+	{ id: 'power', label: 'Power' },
+	{ id: 'lastSeen', label: 'Last seen' },
+]
```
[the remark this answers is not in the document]
AlCalzone: We should have one single source of truth for the list of toggleable columns

src/components/dashboard/components/ZwDeviceRow.vue
```diff
@@ -0,0 +1,250 @@
+<template>
+	<div
+		class="zw-row"
+		:class="{
+			'zw-row--expanded': expanded,
+			'zw-row--mobile': viewport < 600,
+		}"
+		:style="rowStyle"
+		@click="emit('expand', device.id)"
+	>
```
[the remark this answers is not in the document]
AlCalzone: The button to toggle a device row was made keyboard accessible, wasn't it?

src/components/dashboard/components/ZwExpandedRow.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+	<div class="zw-expanded-row">
+		<ZwNodeDetailsBody
+			:device="device"
+			:viewport="viewport"
+			@action="(d, a) => emit('action', d, a)"
+		/>
```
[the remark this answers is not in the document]
AlCalzone: is this still relevant?

src/components/dashboard/components/ZwDeviceRow.vue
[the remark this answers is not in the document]
AlCalzone: Already keyboard-accessible: the expand chevron is a focusable `<button>` with `aria-expanded` and a label, so rows are operable by keyboard/AT. The row `<div>` click is a pointer-only convenience — making the div itself interactive would nest it with the in-row toggle/columns controls, which is the a11y anti-pattern this deliberately avoids.

src/components/dashboard/components/ZwColumnsMenu.vue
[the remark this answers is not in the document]
AlCalzone: Addressed — the toggleable-column list now lives only in `deviceRowGrid.ts` (`TOGGLEABLE_COLS`, next to the `ToggleableCol` union). `ZwColumnsMenu` and every other consumer import it instead of re-declaring it (7334a060; the 05/06 consumers were de-duplicated in follow-up commits).

src/lib/dashboard-types.ts
```diff
@@ -0,0 +1,194 @@
...
+	| 'siren'
+	| 'remote'
+	| 'rgb'
+	| 'controller'
+
+export interface Archetype {
+	kind: ArchetypeKind
+	label: string
+	// Aliased Lucide component (e.g. SwitchIcon from @/lib/icons).
+	icon: unknown
```
robertsLando: **[Major] · Design** — `Archetype.icon: unknown`.

**Why:** it's consumed by `<component :is="device.archetype.icon">` in ZwDeviceCard/Row — `unknown` gives zero type-checking at the one place a wrong value silently renders nothing, and is the weakest possible type on a load-bearing field.

**Fix:** type it as Vue's `Component` (the same `import type { Component } from 'vue'` already used in `registry.ts`). The comment even says "Aliased Lucide component," so the intent is known.

src/components/dashboard/components/primary-display/ZwPrimaryState.vue
```diff
@@ -0,0 +1,52 @@
...
+	</div>
+</template>
+
+<script setup lang="ts">
+import { computed } from 'vue'
+import type { Device, PrimaryValueState } from '@/lib/dashboard-types'
+
+const props = defineProps<{ device: Device; compact?: boolean }>()
+
+const pv = computed(() => props.device.primaryValue as PrimaryValueState)
```
robertsLando: **[Major] · Design** — representative of the ~22 `as PrimaryValueX` casts across the renderers.

**Why:** the unchecked cast severs the link between a registry entry and its matching `PrimaryValue` subtype. The registry maps `key → Component`, but nothing ties a component to its subtype, so a mis-wire (e.g. `reading: Toggle`) type-checks and fails only at runtime.

**Fix:** a generic `PrimaryRenderer<T extends PrimaryValueType>` contract to make the registry type-safe end-to-end, or an internal discriminant guard (`if (pv.type !== 'state') return null`) so narrowing is checked rather than asserted. (In `ZwCompactPrimary` the casts are also redundant — the `pv?.type ===` guards already narrow.)

src/components/dashboard/components/ZwDeviceRow.vue
```diff
@@ -0,0 +1,275 @@
...
+
+const props = defineProps<{
+	device: Device
+	expanded: boolean
+	columns: ToggleableCol[]
+	viewport: number
+}>()
+
+const emit = defineEmits<{
+	expand: [Device['id']]
```
robertsLando: **[Minor] · Design** — sibling "show details" intents diverge: this row emits `expand: [Device['id']]`, while `ZwDeviceCard` emits `open: [Device]` for the same concept.

**Why:** a parent listening to both must handle two names and two payload shapes for one logical intent, making the future dispatcher/view wiring more fragile.

**Fix:** align on one name + payload, or document the distinction if open-drawer vs expand-row are genuinely different.

src/components/dashboard/components/deviceRowGrid.ts
```diff
@@ -0,0 +1,54 @@
...
+	power: '72px',
+	signal: '18px',
+	lastSeen: '88px',
+}
+
+export function deviceRowGrid(
+	viewport: number,
+	columns: readonly ToggleableCol[],
+): string {
+	if (viewport < 600) {
```
robertsLando: **[Minor] · Readability** — hardcoded `600` viewport breakpoint, duplicated here and in `ZwDeviceRow.vue:6` and `ZwDeviceDrawer.vue:61`.

**Why:** the breakpoint strategy is opaque and error-prone to keep in sync across three files.

**Fix:** extract a shared `MOBILE_BREAKPOINT = 600` constant.

src/components/dashboard/components/ZwCompactPrimary.vue
```diff
@@ -0,0 +1,141 @@
...
+	PrimaryValueState,
+	PrimaryValueThermostat,
+} from '@/lib/dashboard-types'
+
+const props = defineProps<{ device: Device }>()
+const emit = defineEmits<{ action: [Device, DeviceAction] }>()
+
+const pv = computed<PrimaryValue | null>(() => props.device.primaryValue)
+
+const stateChipTone = computed(() => {
```
robertsLando, line 75: **[Minor] · DRY** — the alert predicate (`stateIdx === 1 && colors[1] ∈ {red, amber}`) is implemented here, in `ZwPrimaryState.vue:21` (`isAlert`), and in `ZwDeviceCard`.

**Why:** identical load-bearing logic in three places — if the alert condition changes, all three must change together (and the triplication already caused a false-positive bug report during review).

**Fix:** extract a shared `isStateAlert(state)` helper (boolean) and keep tone-mapping separate.

src/components/dashboard/components/ZwSecurityPanel.vue
```diff
@@ -0,0 +1,84 @@
...
+	font-family: var(--zw-mono);
+	font-size: 11px;
+}
+
+.zw-sec__label {
+	color: var(--zw-fg);
+}
+
+.zw-sec__state {
+	color: #047857;
```
robertsLando: **[Minor] · Readability** — raw hex `#047857` (also `ZwPrimaryLock.vue:52` `#a14b1f`, and `ZwPrimaryState.vue`).

**Why:** these are the only 3 of 19 files using raw hex — everything else uses `var(--zw-*)` design tokens, so they break the established convention.

**Fix:** use a theme token (e.g. `var(--zw-success)`), adding one if missing.

robertsLando (commented):
## Deep review — feat(ui): implement more complex UI components

**Verdict: Ship with minor changes.** The approach is sound — clean container/presentational separation over the atoms layer, atoms correctly reused (no hand-rolled markup), the grid template genuinely unified between header and rows, and the parent owns the resize observer (no per-component listeners). No blockers.

**Top risks**
1. `Archetype.icon: unknown` erases type-checking at a dynamic `<component :is>` render site.
2. Pervasive `as PrimaryValueX` casts defeat the registry's discriminated union — a mis-wire (`reading: Toggle`) type-checks and fails only at runtime.
3. The alert-detection predicate (`stateIdx === 1 && colors[1] ∈ {red, amber}`) is copy-pasted across 3 components.

**Verification note:** a candidate blocker — `ZwPrimaryState.isAlert` reading `.stateIdx`/`.colors` off a string — was investigated and **dropped as a false positive**. `pv` is a `ComputedRef`, so in script context `pv.value` unwraps to the `PrimaryValueState` object; it matches the verified-correct `ZwCompactPrimary.stateChipTone`.

**Security:** clean — no `v-html`, no secret-bearing fields (`ZwSecurityPanel` renders only grant booleans + literal key labels), all device strings auto-escaped via `{{ }}`.

**Tests:** repo norm is backend-only (mocha in `test/lib/`); no Vue component tests exist, so none here is consistent with precedent. Cheapest high-value targets if wanted: `deviceRowGrid.ts` packing math + a `registry` exhaustiveness assertion.

*(Stacked-PR context acknowledged: no view wiring / no tests yet is expected, not a defect.)*

src/components/dashboard/components/ZwAddDeviceSplitButton.vue
```diff
@@ -0,0 +1,255 @@
...
+const MENU = [
+	{
+		id: 'include' as const,
+		icon: AddIcon,
+		title: 'Include device',
+		desc: 'Add a new device to the network',
+		destructive: false,
+	},
+	{
+		id: 'replace' as const,
```
robertsLando: **[Minor] · Correctness** — the middle action id is `'replace'`, but the corresponding `DeviceAction` shape is `'replace-failed'` (there is no controller-level `{type:'replace'}`).

**Why:** the split button is self-contained (emits its own `'include'|'replace'|'exclude'` union, not a `DeviceAction`), so nothing breaks in this PR — but the plan-70 dispatcher will need an explicit `'replace' → 'replace-failed'` mapping or it silently mis-routes.

**Fix:** rename the id to `'replace-failed'`, or document the mapping at the emit site.

src/components/dashboard/components/ZwDeviceRow.vue
AlCalzone: we should unify this.

src/components/dashboard/components/primary-display/ZwPrimaryState.vue
AlCalzone: generic sounds good

src/components/dashboard/components/ZwCompactPrimary.vue
AlCalzone: sounds good.

src/components/dashboard/components/deviceRowGrid.ts
AlCalzone: shared constant sounds good. We should check if we can derive this from the already defined breakpoints.

src/components/dashboard/components/ZwSecurityPanel.vue
AlCalzone: yes!

src/components/dashboard/components/ZwAddDeviceSplitButton.vue
AlCalzone: yes, we should avoid unnecessary mappings.

src/lib/dashboard-types.ts
AlCalzone: Done in `cf9b7ff` — typed as Vue's `Component`, the same contract `primary-display/registry.ts` uses.

src/components/dashboard/components/primary-display/ZwPrimaryState.vue
AlCalzone: Done in `25a7d39` — added a `PrimaryValueByType` discriminant→shape map and a generic `usePrimaryValue<T>` composable. Renderers narrow via a runtime `pv.type === type` guard and get `PrimaryValueByType[T] | null`; the lone remaining cast lives behind that guard, and a `v-if="pv"` makes a registry mis-wire render nothing instead of throwing. Also dropped the now-redundant casts in `ZwCompactPrimary` and `ZwDeviceCard`.

src/components/dashboard/components/ZwDeviceRow.vue
AlCalzone: Done in `a21405a` — `ZwDeviceRow` now emits `open: [Device]` to match `ZwDeviceCard` (one name, one payload; parent owns inline-expand vs drawer). Layout-layer consumers realigned on `05-ui-layout`.

src/components/dashboard/components/deviceRowGrid.ts
AlCalzone: Done in `78e9325` — hoisted to `MOBILE_BREAKPOINT` in `src/lib/dashboard-breakpoints.ts`. It can't derive from the Vuetify thresholds (480/760/1100/1380 in `plugins/vuetify.js` — window-based, and none is 600); this is a container-width cutoff fed by the dashboard's own ResizeObserver `viewport`, so it stays a standalone constant. The drawer's CSS `@media (max-width:600px)` keeps its literal (a stylesheet can't read a JS const) — noted in the constant's doc comment.

src/components/dashboard/components/ZwCompactPrimary.vue
AlCalzone: Done in `85d6f31` — extracted `isStateAlert(state)` to `src/lib/primaryValue.ts` as the single boolean predicate; tone-mapping (red→danger, amber→warn) kept separate at the `ZwCompactPrimary` call site.

src/components/dashboard/components/ZwSecurityPanel.vue
AlCalzone: Done in `550a220` — `#047857`→`var(--zw-ok)` (granted/success), `#a14b1f`→`var(--zw-warning)` (alert). The two alert sites now match `ZwDeviceCard`'s alert icon, which already used `var(--zw-warning)`.

src/components/dashboard/components/ZwAddDeviceSplitButton.vue
AlCalzone: Done in `e14a41c` — renamed the id to `replace-failed` and dropped the spurious `{ type: 'replace' }` from `DeviceAction` (no controller-level "replace" exists). `ZwNodeDetailsBody` updated to match, so every emit site already carries the canonical type — no dispatcher mapping needed.

PR #4697, opened 2026-06-26
feat: implement values table and polish events log

AlCalzone:
This PR implements the values tab for a node, both for the table view (with the events tab next to it)...
image

...and the card view:
image

### DIFF
--- a/src/components/dashboard/components/ZwNodeDetailsBody.vue
+++ b/src/components/dashboard/components/ZwNodeDetailsBody.vue
@@ -164,7 +164,12 @@ import ZwStatsCard, { type StatsItem } from './ZwStatsCard.vue'
 import ZwNodeEvents from './ZwNodeEvents.vue'
 import ZwValuesView from './ZwValuesView.vue'
 import { signalDisplay } from '@/lib/deviceSignal'
-import { TWO_PANE_BREAKPOINT } from '@/lib/dashboard-breakpoints'
+import {
+	RAIL_WIDTH_BREAKPOINT,
+	RAIL_WIDTH_COMPACT,
+	RAIL_WIDTH_SPACIOUS,
+	TWO_PANE_BREAKPOINT,
+} from '@/lib/dashboard-breakpoints'
 import type { Device, DeviceAction } from '@/lib/dashboard-types'
 
 // `viewport` is the host panel's available width (the shell width passed down
@@ -220,8 +225,11 @@ const tabs = computed(() =>
 		: ALL_TABS,
 )
 
-// Rail follows the design's two widths: roomier on desktop, tighter below.
-const railWidth = computed(() => (props.viewport >= 1200 ? 340 : 300))
+const railWidth = computed(() =>
+	props.viewport >= RAIL_WIDTH_BREAKPOINT
+		? RAIL_WIDTH_SPACIOUS
+		: RAIL_WIDTH_COMPACT,
+)
 
 function defaultTab(): TabId {
 	return props.device.isController ? 'summary' : 'values'
--- a/src/components/dashboard/components/ZwNodeDetailsBody.vue
+++ b/src/components/dashboard/components/ZwNodeDetailsBody.vue
@@ -1,14 +1,54 @@
 <template>
-	<div class="zw-nd">
-		<header class="zw-nd__header">
+	<div class="zw-nd" :class="`zw-nd--${mode}`">
+		<!-- ── Two-pane left rail: primary · status · signal · activity ──
+		     Only in the wide (split) layout. The rail surfaces what would
+		     otherwise be the Events tab, which is dropped from the tab bar
+		     in this mode. -->
+		<aside
+			v-if="mode === 'split'"
+			class="zw-nd__rail"
+			:style="{ width: railWidth + 'px' }"
+		>
+			<div class="zw-nd__rail-primary">
+				<span class="zw-nd__overline">PRIMARY</span>
+				<div class="zw-nd__primary">
+					<ZwPrimaryDisplay :device="device" @action="onAction" />
+				</div>
+			</div>
+			<div class="zw-nd__rail-section">
+				<span class="zw-nd__overline">STATUS</span>
+				<span class="zw-nd__status">
+					<ZwStatusDot :status="dotStatus" />
+					<span class="zw-nd__status-label">{{ statusLabel }}</span>
+					<span class="zw-nd__bullet">·</span>
+					<span class="zw-nd__lastseen"
+						>last seen {{ device.lastSeen }}</span
+					>
+				</span>
+			</div>
+			<div class="zw-nd__rail-section">
+				<span class="zw-nd__overline">SIGNAL</span>
+				<span class="zw-nd__signal">
+					<component
+						:is="signal.icon"
+						:size="14"
+						:style="{ color: signal.color }"
+					/>
+					<span>{{ signal.label }}</span>
+				</span>
+			</div>
+			<div class="zw-nd__rail-activity">
+				<span class="zw-nd__overline">RECENT ACTIVITY</span>
+				<ZwNodeEvents :device="device" :max="25" />
+			</div>
+		</aside>
+
+		<!-- ── Stacked header: full-width above the tabs (drawer / mobile) ── -->
+		<header v-else class="zw-nd__header">
 			<div class="zw-nd__header-top">
 				<span class="zw-nd__overline">PRIMARY</span>
 				<span class="zw-nd__status">
-					<ZwStatusDot
-						:status="
-							device.isController ? 'controller' : device.status
-						"
-					/>
+					<ZwStatusDot :status="dotStatus" />
 					<span class="zw-nd__status-label">{{ statusLabel }}</span>
 					<span class="zw-nd__bullet">·</span>
 					<span class="zw-nd__lastseen"
@@ -17,112 +57,98 @@
 				</span>
 			</div>
 			<div class="zw-nd__primary">
-				<ZwPrimaryDisplay
-					:device="device"
-					@action="(d, a) => emit('action', d, a)"
-				/>
+				<ZwPrimaryDisplay :device="device" @action="onAction" />
 			</div>
 		</header>
 
-		<Tabs.Root v-model="tab">
-			<Tabs.List as="nav" class="zw-nd__tabs">
-				<Tabs.Item
-					v-for="t in tabs"
-					:key="t.id"
-					:value="t.id"
-					class="zw-nd__tab"
-				>
-					{{ t.label }}
-				</Tabs.Item>
-			</Tabs.List>
-
-			<Tabs.Panel value="values" class="zw-nd__content">
-				<div v-if="device.isController" class="zw-nd__empty">
-					Controller exposes no command-class values.
-				</div>
-				<div v-else class="zw-nd__values-stub">
-					<!-- Placeholder until the full values view is built. -->
-					Values pane — full view coming soon.
-				</div>
-			</Tabs.Panel>
-
-			<Tabs.Panel value="summary" class="zw-nd__content zw-nd__summary">
-				<ZwPropTable :rows="manufacturerRows" />
-				<ZwPropTable :rows="firmwareRows" />
-				<ZwSecurityPanel :device="device" />
-			</Tabs.Panel>
-
-			<Tabs.Panel value="groups" class="zw-nd__content">
-				<ZwPropTable :rows="groupsStub" />
-			</Tabs.Panel>
-
-			<Tabs.Panel value="updates" class="zw-nd__content zw-nd__updates">
-				<div v-if="device.hasUpdate" class="zw-nd__update-card">
-					<div class="zw-nd__update-current">
-						Current {{ device.firmware?.node ?? '—' }}
-					</div>
-					<div class="zw-nd__update-line">
-						Update available — see the device drawer's footer to
-						apply.
-					</div>
-				</div>
-				<div v-else class="zw-nd__empty">
-					No firmware update available.
-				</div>
-			</Tabs.Panel>
-
-			<Tabs.Panel value="events" class="zw-nd__content zw-nd__events">
-				<template v-if="events.length === 0">
-					<div class="zw-nd__empty">No recent events</div>
-				</template>
-				<template v-else>
-					<div
-						v-for="(ev, i) in events"
-						:key="i"
-						class="zw-nd__event"
+		<!-- ── Tabbed detail content — identical in both layouts. Sits in
+		     the right pane (split) or below the header (stacked). -->
+		<div class="zw-nd__main">
+			<Tabs.Root v-model="tab">
+				<Tabs.List as="nav" class="zw-nd__tabs">
+					<Tabs.Item
+						v-for="t in tabs"
+						:key="t.id"
+						:value="t.id"
+						class="zw-nd__tab"
 					>
-						<span class="zw-nd__event-time">{{
-							relativeTime(toMs(ev.time), now)
-						}}</span>
-						<div class="zw-nd__event-body">
-							<div class="zw-nd__event-name">{{ ev.event }}</div>
-							<div
-								v-if="eventDetail(ev)"
-								class="zw-nd__event-detail"
-							>
-								{{ eventDetail(ev) }}
-							</div>
+						{{ t.label }}
+					</Tabs.Item>
+				</Tabs.List>
+
+				<Tabs.Panel value="values" class="zw-nd__content">
+					<div v-if="device.isController" class="zw-nd__empty">
+						Controller exposes no command-class values.
+					</div>
+					<ZwValuesView v-else :device="device" @action="onAction" />
+				</Tabs.Panel>
+
+				<Tabs.Panel
+					value="summary"
+					class="zw-nd__content zw-nd__summary"
+				>
+					<ZwPropTable :rows="manufacturerRows" />
+					<ZwPropTable :rows="firmwareRows" />
+					<ZwSecurityPanel :device="device" />
+				</Tabs.Panel>
+
+				<Tabs.Panel value="groups" class="zw-nd__content">
+					<ZwPropTable :rows="groupsStub" />
+				</Tabs.Panel>
+
+				<Tabs.Panel
+					value="updates"
+					class="zw-nd__content zw-nd__updates"
+				>
+					<div v-if="device.hasUpdate" class="zw-nd__update-card">
+						<div class="zw-nd__update-current">
+							Current {{ device.firmware?.node ?? '—' }}
+						</div>
+						<div class="zw-nd__update-line">
+							Update available — see the device drawer's footer to
+							apply.
 						</div>
 					</div>
-					<div
-						v-if="totalEvents > MAX_EVENTS"
-						class="zw-nd__event-footer"
-					>
-						+{{ totalEvents - MAX_EVENTS }} earlier events
+					<div v-else class="zw-nd__empty">
+						No firmware update available.
 					</div>
-				</template>
-			</Tabs.Panel>
+				</Tabs.Panel>
 
-			<Tabs.Panel value="debug" class="zw-nd__content">
-				<div v-if="device.isController" class="zw-nd__stats">
-					<ZwStatsCard title="Communication" :items="commStats" />
-					<ZwStatsCard title="Messages" :items="messageStats" />
-				</div>
-				<ZwPropTable v-else :rows="debugRows" />
-			</Tabs.Panel>
-
-			<Tabs.Panel value="advanced" class="zw-nd__content zw-nd__advanced">
-				<ZwButton
-					v-for="cmd in advancedCommands"
-					:key="cmd.label"
-					variant="mono-outline"
-					size="sm"
-					@click="emit('action', device, cmd.action)"
+				<!-- Events is omitted from the tab bar in the split layout
+				     (its content lives in the rail), so the panel is only
+				     mounted in the stacked layout. -->
+				<Tabs.Panel
+					v-if="mode === 'stacked'"
+					value="events"
+					class="zw-nd__content"
 				>
-					{{ cmd.label }}
-				</ZwButton>
-			</Tabs.Panel>
-		</Tabs.Root>
+					<ZwNodeEvents :device="device" />
+				</Tabs.Panel>
+
+				<Tabs.Panel value="debug" class="zw-nd__content">
+					<div v-if="device.isController" class="zw-nd__stats">
+						<ZwStatsCard title="Communication" :items="commStats" />
+						<ZwStatsCard title="Messages" :items="messageStats" />
+					</div>
+					<ZwPropTable v-else :rows="debugRows" />
+				</Tabs.Panel>
+
+				<Tabs.Panel
+					value="advanced"
+					class="zw-nd__content zw-nd__advanced"
+				>
+					<ZwButton
+						v-for="cmd in advancedCommands"
+						:key="cmd.label"
+						variant="mono-outline"
+						size="sm"
+						@click="emit('action', device, cmd.action)"
+					>
+						{{ cmd.label }}
+					</ZwButton>
+				</Tabs.Panel>
+			</Tabs.Root>
+		</div>
 	</div>
 </template>
 
@@ -135,12 +161,29 @@ import ZwPrimaryDisplay from './ZwPrimaryDisplay.vue'
 import ZwPropTable from './ZwPropTable.vue'
 import ZwSecurityPanel from './ZwSecurityPanel.vue'
 import ZwStatsCard, { type StatsItem } from './ZwStatsCard.vue'
+import ZwNodeEvents from './ZwNodeEvents.vue'
+import ZwValuesView from './ZwValuesView.vue'
+import { signalDisplay } from '@/lib/deviceSignal'
+import { TWO_PANE_BREAKPOINT } from '@/lib/dashboard-breakpoints'
 import type { Device, DeviceAction } from '@/lib/dashboard-types'
-import useBaseStore from '@/stores/base'
-import { relativeTime, useNow } from '@/lib/time'
 
-const props = defineProps<{ device: Device }>()
+// `viewport` is the host panel's available width (the shell width passed down
+// by the table, NOT the device viewport). It selects the layout and feeds the
+// rail width. `layout="stacked"` forces the single-column layout regardless of
+// width — the card-view drawer uses it so it always keeps the stacked layout.
+const props = withDefaults(
+	defineProps<{
+		device: Device
+		viewport?: number
+		layout?: 'auto' | 'stacked'
+	}>(),
+	{ viewport: 0, layout: 'auto' },
+)
 const emit = defineEmits<{ action: [Device, DeviceAction] }>()
+
+function onAction(d: Device, a: DeviceAction) {
+	emit('action', d, a)
+}
 
 type TabId =
 	| 'values'
@@ -151,7 +194,7 @@ type TabId =
 	| 'debug'
 	| 'advanced'
 
-const tabs: { id: TabId; label: string }[] = [
+const ALL_TABS: { id: TabId; label: string }[] = [
 	{ id: 'values', label: 'Values' },
 	{ id: 'summary', label: 'Summary' },
 	{ id: 'groups', label: 'Groups' },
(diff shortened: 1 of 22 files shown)


src/lib/valueGroups.ts
```diff
@@ -0,0 +1,191 @@
+// Projects a node's `values` into the grouped, control-ready shape the Values
+// pane renders. Pure — the component memoizes it with a computed.
+
+import { CommandClasses, type ValueID } from '@zwave-js/core'
+import type { ZUINode, ZUIValueId } from '../../api/lib/ZwaveClient.ts'
```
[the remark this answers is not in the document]
AlCalzone: linting passed though.

src/components/dashboard/components/ZwValueRow.vue
```diff
@@ -0,0 +1,689 @@
...
+)
+
+function coerce(raw: unknown): unknown {
+	if (props.param.kind === 'number' || props.param.kind === 'level') {
+		return Number(raw)
+	}
+	return raw
+}
+
+function commit(value: unknown) {
```
robertsLando, line 314: **[Blocker] · Correctness × Operability**

The optimistic `draft` is never rolled back on write failure. `commit()` pins `draft` to the attempted value for readable params, and the only thing that clears it is `watch(() => props.param.value)` (line ~310), which fires *only* when the device confirms a new value. On a failed or un-acked write the value is unchanged, `busy` clears via the `finally` in `ControlPanelNew.onAction`, the row re-enables — and it keeps showing the attempted value while the error snackbar says the write failed. The control actively contradicts the toast.

**Why:** Common for sleeping/unreachable Z-Wave nodes; the user can't tell whether the write took, and the UI misrepresents device state until the next genuine report.

**Fix:** Propagate success/failure back into the row (e.g. a second injected `failedKeys` Set the row watches, reverting `draft` to `null` on failure), or skip the optimistic draft for instant-commit kinds (switch/enum/level) and keep showing the live value until the watcher fires.

src/lib/valueGroups.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// Picks the control *shape* (editability is `readonly`, set elsewhere). A value
+// can match several predicates, so order matters — most specific first.
+function paramKind(v: ZUIValueId): ValueParamKind {
+	// write-only boolean = a command (e.g. Meter reset), not a toggle
+	if (v.type === 'boolean' && v.writeable && !v.readable) return 'button'
+	if (v.type === 'boolean') return 'switch'
+	if (v.type === 'color') return 'color'
+	if (isLevel(v)) return 'level'
```
robertsLando, line 114: **[Major] · Correctness**

`isLevel(v)` is tested before the `hasStates → enum` branch (line 109), so any 0–99 number that carries discrete `states` renders as a level slider and loses its labeled dropdown.

**Why:** Violates the "most specific first" rule the comment promises — an enumerated value is more specific than "any 0–99 number."

**Fix:** Gate the generic `type === 'number' && min === 0 && max === 99` case behind `!hasStates(v)`, keeping the explicit Multilevel-Switch CC check ahead of `enum`.

src/lib/valueGroups.ts
```diff
@@ -0,0 +1,191 @@
...
+	// states → dropdown, unless free entry is allowed (keep arbitrary values reachable)
+	if (hasStates(v) && !(v.writeable && v.allowManualEntry)) return 'enum'
+	if (v.type === 'number') return 'number'
+	if (v.type === 'string' || v.type === 'buffer') return 'text'
+	return 'reading' // duration / any / unsupported → read-only, never written
```
robertsLando, line 119: **[Major] · Design**

`paramKind` returns `'reading'` for duration/any/unsupported types, but `ZwValueRow`'s template only renders those via the `v-if="param.readonly"` branch (template line ~97). A *writeable* such value (`readonly: false`) matches none of switch/enum/number/level/text/color/button → it renders an empty control (just the busy span). The discriminant looks exhaustive but isn't.

**Why:** Silent dead control for any writeable duration/any value.

**Fix:** Force `readonly: true` for `reading` in `projectParam`, or add an explicit `reading` fallback branch in the template.

src/views/ControlPanelNew.vue
```diff
@@ -38,17 +44,28 @@ function appInstance(): AppLike | null {
...
-		errorSnack: true,
-	})
+	const key = actionPendingKey(device, action)
+	if (key) pending.add(key)
+	try {
+		await app.apiRequest(req.api, req.args, {
+			infoSnack: false,
+			errorSnack: true,
+		})
+	} finally {
```
robertsLando, line 72: **[Major] · Operability**

`pending.delete(key)` lives only in the `finally`, which runs only when `apiRequest` settles. If the promise never settles (node unreachable, driver queue stall), the key is never removed: the row spinner spins forever and the control stays disabled for the rest of the session with no recourse.

**Why:** Queued RF writes to sleeping/dead nodes are the field's most common failure mode for a dashboard operator.

**Fix:** Race the request against a deadline (~30–60s) that rejects with a recognizable timeout error and surfaces a specific toast, while ensuring cleanup still runs; or track per-key start timestamps and sweep stale keys.

src/components/dashboard/components/ZwValuesView.vue
```diff
@@ -0,0 +1,138 @@
...
+import type { ValueID } from '@zwave-js/core'
+
+const props = defineProps<{ device: Device }>()
+const emit = defineEmits<{ action: [Device, DeviceAction] }>()
+
+const baseStore = useBaseStore()
+
+// Reads live from the store so writes/polls re-render as values land.
+const groups = computed<ValueGroup[]>(() =>
+	buildValueGroups(baseStore.getNode(props.device.nodeId)),
```
robertsLando, line 51: **[Major] · Performance**

`buildValueGroups` runs inside a `computed` that tracks the entire reactive node, so any value report rebuilds all n `ValueParam`/`ValueID`/options objects with fresh references. Every `ZwValueRow` then sees its `param` prop change and re-runs its ~7 computeds (`cur`, `dirty`, `busy`, `levelValue`, `rangeMeta`, `sending`, `refreshing`). At n≈200 that is ~1,400 computed runs per socket event, even when only one value changed.

**Why:** Value reports are the most frequent Z-Wave event; this is an O(n) rebuild + re-render storm on the hot path.

**Fix:** Cache `ValueParam` objects keyed by value `id`, recreating only when content changes; or have each row subscribe to its own valueId slice so updates are surgical.

src/lib/valueGroups.ts
```diff
@@ -0,0 +1,191 @@
...
+			? v.states.map((s) => ({ value: s.value, label: s.text }))
+			: undefined,
+		paramNumber: isConfig ? String(v.property) : undefined,
+		target: toValueId(v),
+	}
+}
+
+// Groups a node's values by command class (first-seen order), each projected to
+// a render-ready `ValueParam`. `[]` for a node without values (e.g. controller).
+export function buildValueGroups(node?: ZUINode | null): ValueGroup[] {
```
robertsLando, line 194: **[Major] · Tests**

The core projection logic — `paramKind`, `formatValue`, `projectParam`, `buildValueGroups` — ships with no tests, while `src/lib/deviceProjection.test.ts` establishes that `src/lib/` projection is expected to be covered.

**Why:** This is branch-heavy logic that *drives rendering decisions*; the ordering and exhaustiveness bugs flagged elsewhere in this review are exactly what a unit test would pin.

**Fix:** Unit-test `paramKind` (ordering across kinds incl. 0–99-with-states and writeable-reading), `formatValue` (enum index, duration, buffer, undefined), and `projectParam` (Configuration CC defaults/modified flag).

src/lib/deviceActionPending.ts
```diff
@@ -0,0 +1,50 @@
...
+export type PendingSet = Set<string>
+
+export const DeviceActionPendingKey: InjectionKey<PendingSet> = Symbol(
+	'zwDeviceActionPending',
+)
+
+// Stable per-node value identity (`cc-endpoint-property-key`); keys pending
+// requests and matches event args back to values. Partial: event args may
+// carry only some of the fields.
+export function valueIdKey(v: Partial<ValueID>): string {
```
robertsLando, line 22: **[Major] · Tests**

The key-identity helpers (`valueIdKey`, `setPendingKey`, `pollPendingKey`, `ccPendingKey`, `actionPendingKey`) are untested. These hash ValueIDs into the strings that drive every spinner; a format or collision bug here silently breaks pending state across all values.

**Why:** Provider (`ControlPanelNew`) and consumers (`ZwValueRow`, `ZwNodeEvents`) rely on these keys matching exactly.

**Fix:** Tests verifying stable key format, edge cases (undefined endpoint/propertyKey), and no collisions across nodes/CCs.

src/components/dashboard/components/ZwValueRow.vue
```diff
@@ -0,0 +1,689 @@
...
+
+// Once the confirmed value updates, drop the optimistic draft.
+watch(
+	() => props.param.value,
+	() => {
+		draft.value = null
+	},
+)
+
+function coerce(raw: unknown): unknown {
```
robertsLando, line 307: **[Minor] · Correctness**

Clearing a number input then pressing Apply sends `Number('') === 0` rather than a no-op, and `dirty` compares `String(draft) !== String(value)` so `"05"` vs `5` reads as dirty.

**Why:** Unintended `0` writes and spurious dirty state on edge inputs.

**Fix:** Guard empty/NaN before committing numbers.

src/lib/device-actions.ts
```diff
@@ -50,6 +50,19 @@ export const ACTION_DISPATCHERS: {
 		api: 'writeValue',
 		args: [{ nodeId: d.nodeId, ...a.valueId }, a.mode],
 	}),
+	// Generic value-pane interactions.
+	'set-value': (d, a) => ({
```
robertsLando, line 55: **[Minor] · Design**

The Multilevel-Switch ≤99 clamp lives in the `'dim'` dispatcher and in `ZwValueRow.commitLevel`, but the generic `'set-value'` path the values table actually emits for levels does not clamp.

**Why:** A level write that goes through `set-value` (bypassing `commitLevel`) is unclamped; the invariant is enforced in two unrelated places and absent from the one that needs it.

**Fix:** Centralize the clamp, or document that `set-value` trusts the caller.

src/components/dashboard/deviceActionPending.ts
```diff
@@ -0,0 +1,50 @@
+// Reactive Set of in-flight value-pane request keys. The host fills it around
+// `apiRequest` so spinners clear on response (not a timer, and not only when a
+// value changes — a poll often re-reads the same value).
+
+import type { InjectionKey } from 'vue'
+import type { ValueID } from '@zwave-js/core'
+import type { Device, DeviceAction } from '@/lib/dashboard-types.ts'
+
+export type PendingSet = Set<string>
```
robertsLando: **[Minor] · Design**

The pure key projections (`valueIdKey`, `setPendingKey`, …) live under `components/` while every sibling projection lib (`valueGroups`, `deviceSignal`, `device-actions`) sits in `src/lib/`; `ZwNodeEvents.vue` even reaches into the `components/` path to import `valueIdKey`.

**Why:** Placement inconsistency — only the `InjectionKey` truly belongs to the component layer.

**Fix:** Move the pure helpers to `src/lib/`, keep the `InjectionKey` with the provider (or co-locate everything in lib).

src/lib/valueGroups.ts
```diff
@@ -0,0 +1,191 @@
...
+	| 'text'
+	| 'color'
+	| 'reading'
+
+export interface ValueOption {
+	value: number | string
+	label: string
+}
+
+export interface ValueParam {
```
robertsLando, line 30: **[Minor] · Design**

`ValueParam` is a flat interface with all kind-specific fields optional (`options?`, `min?`, `step?`, `paramNumber?`) rather than a discriminated union on `kind`. `kind` drives all rendering yet gives zero compile-time narrowing, so the row leans on `?.` and prose comments instead of the type.

**Why:** Weaker contract for the most important branch point in the feature.

**Fix:** A `kind`-discriminated union would harden it. (Acknowledged: sibling `ZUIValueId` is also flat, so this is a deliberate-convention call, not a must.)

src/views/ControlPanelNew.vue
```diff
@@ -38,17 +44,28 @@ function appInstance(): AppLike | null {
 	return (inst as unknown as AppLike) ?? null
 }
 
-function onAction(device: Device, action: DeviceAction) {
+// Value-pane components watch this to show a spinner only while their request
+// is actually in flight.
+const pending = reactive(new Set<string>()) as PendingSet
```
robertsLando: **[Minor] · Performance**

`pending` is a `reactive(new Set())`. Vue's reactive `Set` invalidates every `.has()` tracker on any `add`/`delete`, so each action start/finish re-runs `sending`/`refreshing` in all n mounted rows (2n re-evaluations per state change).

**Why:** Unnecessary fan-out on every value action.

**Fix:** Use a `shallowRef<Set<string>>` replaced on mutation, or scope the signal to the affected rows.

src/components/dashboard/components/ZwNodeEvents.vue
```diff
@@ -0,0 +1,198 @@
...
+
+// A value-updated/notification event arg, loosely typed — queue args are
+// untyped, so every field may be absent.
+type ValueEventArg = Partial<
+	TranslatedValueID & ValueUpdatedArgs & ValueNotificationArgs
+>
+
+// value-id → metadata label, rebuilt per node-values change. O(1) lookup avoids
+// re-scanning values per event per render; raw event payloads lack the label.
+const valueLabelIndex = computed(() => {
```
robertsLando, line 84: **[Minor] · Performance**

`valueLabelIndex` rebuilds a full Map over all node values on every value update (same reactive dependency as `buildValueGroups`). In split mode both `ZwValuesView` and `ZwNodeEvents` are mounted, so each value update drives two O(n) passes over `node.values`.

**Fix:** Derive label lookups from the cached `ValueParam` set, or memoize on a values-length/timestamp sentinel.

src/components/dashboard/components/ZwNodeDetailsBody.vue
```diff
@@ -161,21 +204,55 @@ const tabs: { id: TabId; label: string }[] = [
...
+
+// In split mode the Events tab is dropped — its content lives in the rail.
+const tabs = computed(() =>
+	mode.value === 'split'
+		? ALL_TABS.filter((t) => t.id !== 'events')
+		: ALL_TABS,
+)
+
+// Rail follows the design's two widths: roomier on desktop, tighter below.
+const railWidth = computed(() => (props.viewport >= 1200 ? 340 : 300))
```
robertsLando: **[Minor] · DRY**

Hardcoded `1200` / `340` / `300` for rail-width selection is inconsistent with the `TWO_PANE_BREAKPOINT` pattern in `dashboard-breakpoints.ts`.

**Why:** Easy to miss this magic number if the threshold ever needs adjusting alongside the others.

**Fix:** Extract to named constants in `dashboard-breakpoints.ts` (e.g. `RAIL_WIDTH_BREAKPOINT`, `RAIL_WIDTH_SPACIOUS`, `RAIL_WIDTH_COMPACT`).

src/lib/deviceSignal.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export interface SignalDisplay {
+	icon: Component // Lucide, rendered via `<component :is>`
+	color: string
+	label: string
+}
+
+// Device link health → signal glyph/color/label, shared by the table row and
+// the details rail. Only `weak` reads as degraded today.
+export function signalDisplay(health: Device['health']): SignalDisplay {
```
robertsLando, line 13: **[Minor] · Tests**

`signalDisplay` is untested. It branches on `health === 'weak'` to pick icon/color/label and is now used in multiple surfaces (table row + details rail).

**Fix:** Two-case test (weak vs strong health).

src/components/dashboard/components/ZwValueRow.vue
```diff
@@ -0,0 +1,689 @@
...
+						:key="o.value"
+						:value="String(o.value)"
+					>
+						[{{ o.value }}] {{ o.label }}
+					</option>
+				</select>
+			</template>
+
+			<!-- number -->
+			<template v-else-if="param.kind === 'number'">
```
robertsLando: **[Minor] · Readability**

The `number` and `text` input blocks share essentially the same template structure (input + optional unit + apply button).

**Why:** Duplicated structure across kinds increases maintenance burden in an already-large component.

**Fix:** Extract the shared input+apply scaffold despite the semantic difference.

src/components/dashboard/components/ZwNodeEvents.vue
```diff
@@ -0,0 +1,198 @@
+<template>
+	<div class="zw-node-events">
+		<div v-if="rows.length === 0" class="zw-node-events__empty">
+			No recent events
+		</div>
+		<template v-else>
+			<div v-for="(row, i) in rows" :key="i" class="zw-node-events__row">
```
robertsLando: **[Nit] · Performance**

`v-for` key is the index `i` over a newest-first reversed list, so each new event shifts every index by one and Vue re-patches all rows instead of inserting one at the front.

**Fix:** Use a stable key (e.g. `row.timeMs + row.name`), or keep natural order and reverse with CSS `flex-direction: column-reverse`.

src/components/dashboard/components/ZwValueRow.vue
```diff
@@ -0,0 +1,689 @@
...
+const dirty = computed(
+	() =>
+		draft.value !== null &&
+		String(draft.value) !== String(props.param.value),
+)
+const busy = computed(() => sending.value || refreshing.value)
+const levelValue = computed(() => Number(cur.value) || 0)
+
+// Range hint under an editable number; default segment only when one exists.
+const rangeMeta = computed(() => {
```
robertsLando: **[Nit] · Readability**

`rangeMeta` is a vague name for what is a human-readable range hint string.

**Fix:** Rename to `rangeHint` / `rangeDescription`.

src/components/dashboard/components/ZwValueRow.vue
```diff
@@ -0,0 +1,689 @@
...
+}
+
+function copyId() {
+	// Clipboard API is absent in insecure contexts (`?.`) and can reject
+	// (permission denied / not focused); swallow both — it's only feedback.
+	void navigator.clipboard?.writeText(props.param.id).catch(() => {})
+	copied.value = true
+	setTimeout(() => {
+		copied.value = false
+	}, 1200)
```
robertsLando: **[Nit] · Readability**

Magic numbers without named constants: the `1200`ms copy-feedback timeout here, and the hardcoded `80px` / `180px` control widths in the style block.

**Fix:** Extract to module-level constants / named CSS variables.

src/views/ControlPanelNew.vue
```diff
@@ -38,17 +44,28 @@ function appInstance(): AppLike | null {
...
 	if (!app) {
 		console.warn('[ControlPanelNew] App instance not registered yet')
 		return
 	}
-	void app.apiRequest(req.api, req.args, {
-		infoSnack: false,
-		errorSnack: true,
-	})
+	const key = actionPendingKey(device, action)
+	if (key) pending.add(key)
```
robertsLando: **[FYI] · Correctness**

Two concurrent requests with the same key (same valueId set, or repeated refresh-cc) share one Set entry; the first `finally` to run deletes it, clearing the spinner while the second is still in flight. Controls are `:disabled="busy"` so the UI rarely triggers this — flagging for awareness, not action.

src/components/dashboard/components/ZwValueRow.vue
```diff
@@ -0,0 +1,689 @@
...
+	param: ValueParam
+	nodeId: number
+}>()
+const emit = defineEmits<{
+	set: [ValueID, unknown]
+	poll: [ValueID]
+}>()
+
+// Pending edit until applied; null = show the live value. Clears on confirm.
+const draft = ref<unknown>(null)
```
robertsLando, line 259: **[FYI] · Design**

The `draft === null` sentinel conflates "no pending edit" with a legitimately-`null` draft value. Safe today since drafts only originate from user input on editable controls, but a distinct sentinel (or a brief note) would harden it against future kinds.

src/views/ControlPanelNew.vue
```diff
@@ -38,17 +44,28 @@ function appInstance(): AppLike | null {
 	return (inst as unknown as AppLike) ?? null
 }
```
robertsLando, line 44: this is not sometihng specific to control panel, should be moved elsewhere. app is the app.vue vue instance, this used to be a good approach with options api but I think we can refactor it to use a better approach in vue 3

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: Should be addressed by the latest commit, but double check.

src/views/ControlPanelNew.vue
AlCalzone: This seems incorrect. Unless there is a major bug, Z-Wave JS won't stall indefinitely.

src/lib/valueGroups.ts
AlCalzone: yes

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: yes, we should not allow committing empty numeric inputs

src/lib/device-actions.ts
AlCalzone: set-value should be generic and not clamp anything. the caller is in charge of making sensible API calls.

src/lib/valueGroups.ts
AlCalzone: I do like discriminated unions more than a flat catch-all shape.

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: yes

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: rangeHint sounds good

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: sounds reasonable. we should also briefly comment on those magic numbers to explain the rationale behind them

src/views/ControlPanelNew.vue
AlCalzone: we should make sure the UI does not allow double-triggering the same action

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: Fixed in f486af9c: `commit()` no longer pins an optimistic draft. The row shows the live value during the write (`busy` covers the in-flight window) and reflects the real result on confirm, so a failed write no longer contradicts the error toast.

src/lib/valueGroups.ts
AlCalzone: Fixed: a 0–99 number that carries discrete `states` now resolves to `enum` (added `&& !hasStates(v)` to `isLevel`), so the labeled dropdown wins.

src/lib/valueGroups.ts
AlCalzone: Fixed: `projectParam` now forces `readonly: true` for the `reading` kind, so a writeable duration/any value renders read-only instead of an empty control.

src/views/ControlPanelNew.vue
AlCalzone: Per @AlCalzone, leaving as-is — Z-Wave JS won't stall indefinitely, so the `finally` always runs and clears the key.

src/components/dashboard/components/ZwValuesView.vue
AlCalzone: Fixed: `valueGroups.ts` memoizes `ValueParam` by source-object reference (WeakMap). The base store splice-replaces value objects on update, so unchanged values keep stable refs and their rows skip re-render — only the changed value rebuilds.

src/lib/valueGroups.ts
AlCalzone: Done: added `src/lib/valueGroups.test.ts` covering paramKind ordering (incl. 0–99-with-states and writeable-reading), formatValue (enum/number/undefined/object), and Configuration modified/paramNumber.

src/lib/deviceActionPending.ts
AlCalzone: Done: added `src/lib/deviceActionPending.test.ts` — key format, undefined endpoint/propertyKey, and no collisions across nodes/ops/CCs.

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: Fixed: `send()` now blocks empty/NaN numeric input so it can't write a spurious `0`.

src/lib/device-actions.ts
AlCalzone: Per @AlCalzone, `set-value` stays generic — documented that it writes the value verbatim and the caller is responsible for sending a sensible value.

src/components/dashboard/deviceActionPending.ts
AlCalzone: Done: moved to `src/lib/deviceActionPending.ts` (the `InjectionKey` co-locates there, as `deviceSignal` already does); updated importers.

src/lib/valueGroups.ts
AlCalzone: Deferred to a followup PR (plan noted in `.plans`). @AlCalzone prefers the discriminated union; keeping it out of this PR to stay reviewable.

src/views/ControlPanelNew.vue
AlCalzone: Fixed: `pending` is now a `shallowRef<ReadonlySet<string>>` swapped on each change, so membership checks re-run only on a real change rather than via reactive-Set per-key tracking.

src/components/dashboard/components/ZwNodeEvents.vue
AlCalzone: The expensive part (param rebuild + row re-renders) is handled by the WeakMap cache in `valueGroups.ts`; `valueLabelIndex` is mounted only for the open node and is now a cheap O(n) scan. Left as-is for now.

src/components/dashboard/components/ZwNodeDetailsBody.vue
AlCalzone: Fixed: extracted `RAIL_WIDTH_BREAKPOINT` / `RAIL_WIDTH_SPACIOUS` / `RAIL_WIDTH_COMPACT` to `dashboard-breakpoints.ts`.

src/lib/deviceSignal.ts
AlCalzone: Done: added `src/lib/deviceSignal.test.ts` (weak vs non-weak).

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: Done: merged the number and text input blocks into one shared scaffold.

src/components/dashboard/components/ZwNodeEvents.vue
AlCalzone: Fixed: now keyed by a stable content string (`timeMs|event|detail`) instead of the index.

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: Done: renamed to `rangeHint`.

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: Done: `1200` → `COPY_FEEDBACK_MS` (commented); added comments explaining the 80px / 180px input widths.

src/views/ControlPanelNew.vue
AlCalzone: Inline controls are `:disabled="busy"` (busy flips synchronously on first trigger) and menu actions close the menu on click; also added `:disabled="busy"` to the Apply buttons — so the UI prevents double-triggering the same action.

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: Added a note documenting that `null` is safe as the no-edit sentinel (drafts only originate from user input on editable controls, never a literal null).

src/views/ControlPanelNew.vue
AlCalzone: Agreed — this is a broader App.vue / instance-bridge refactor (move the provider out, drop the Options-API instance access) beyond this PR's scope. Deferring to a followup.

PR #4698, opened 2026-06-28
fix(ui): sort values by propertyKey within command class groups

ejohnson-lhi:
## Description

Values within each command class group in the expanded node view (`NodeDetails.vue`) are displayed in arbitrary insertion order. For command classes with keyed properties — most notably **User Code (0x63)** — this means lock code slots appear unsorted (e.g. slot 5, slot 1, slot 12…) rather than in numerical order, making it difficult to find and manage specific slots.

## Changes

Sort each command class group's values by `propertyKey` in the `commandGroups` computed property, so values display in consistent ascending order.

**Files changed:** `src/components/nodes-table/NodeDetails.vue` (4 lines added)

## How to test

1. Open a node that has User Code CC (e.g. a Kwikset or Schlage lock)
2. Expand the node and view the "User Code" command class group under the Node tab
3. Verify slots are displayed in ascending order (1, 2, 3, …)

### DIFF
--- a/src/components/nodes-table/NodeDetails.vue
+++ b/src/components/nodes-table/NodeDetails.vue
@@ -570,6 +570,12 @@ export default {
 					groups.Configuration = []
 				}
 
+				for (const key in groups) {
+					if (groups[key][0]?.commandClass === 99) {
+						groups[key].sort((a, b) => (a.propertyKey ?? 0) - (b.propertyKey ?? 0))
+					}
+				}
+
 				return groups
 			} else {
 				return {}


src/components/nodes-table/NodeDetails.vue
```diff
@@ -570,6 +570,10 @@ export default {
 					groups.Configuration = []
 				}
 
+				for (const key in groups) {
+					groups[key].sort((a, b) => (a.propertyKey ?? 0) - (b.propertyKey ?? 0))
+				}
+
```
robertsLando, line 578: Good catch on the ordering, but this comparator has two issues worth fixing before merge:

1. **`NaN` on string `propertyKey`.** `propertyKey` is `string | number | null` (see `api/lib/utils.ts:151-152`, `ZwaveClient.ts:442`). CCs like Color Switch (`red`/`green`/`blue`), Notification/Alarm and Indicator use string keys, so `(a.propertyKey ?? 0) - (b.propertyKey ?? 0)` evaluates to `NaN` and leaves those groups in an undefined order — potentially scrambling CCs that were previously fine.
2. **Sorting by `propertyKey` alone ignores `property` and `endpoint`.** A group can contain multiple properties and endpoints. For Configuration, partial (bitmask) params carry a `propertyKey` while base params don't, so a partial param gets pulled away from its base param.

Suggested type-safe comparator (endpoint → property → propertyKey, with natural-numeric string ordering so "slot 2" < "slot 10"):

```suggestion
				// sort each group by endpoint → property → propertyKey so keyed
				// CCs (e.g. User Code slots) display in a stable, natural order.
				// Uses a type-safe comparator because propertyKey may be a string
				// (Color Switch, Notification, …) — numeric subtraction would NaN.
				const cmp = (a, b) => {
					if (typeof a === 'number' && typeof b === 'number') {
						return a - b
					}
					return String(a ?? '').localeCompare(String(b ?? ''), undefined, {
						numeric: true,
					})
				}

				for (const key in groups) {
					groups[key].sort(
						(x, y) =>
							cmp(x.endpoint ?? 0, y.endpoint ?? 0) ||
							cmp(x.property, y.property) ||
							cmp(x.propertyKey, y.propertyKey),
					)
				}
```

Note this now reorders *all* CC groups (rather than relying on insertion order), but the comparator is total and stable so it's safe.

AlCalzone (changes requested):
This is a special case that should be done on a per CC basis. Like @robertsLando pointed out, this change would mess up the sorting for `Configuration CC` at least.

I also wonder if this would be made obsolete by a dedicated user management UI:
https://github.com/zwave-js/zwave-js-ui/pull/4638

ejohnson-lhi:
Thanks for the feedback! I've updated the PR to address your concerns:

- Sort is now scoped to **User Code CC (0x63) only** — Configuration CC and all other command classes are left in their original order.
- Aware of #4638 — happy to close this if the dedicated user management UI supersedes it. In the meantime this is a minimal targeted fix for the current slot display.

AlCalzone:
FYI, the credential management PR was merged. Should be in the next release.

PR #4699, opened 2026-06-29
chore: correct TS/ESLint editor config and fix found issues

AlCalzone:
Now that we also have TypeScript in the frontend, I made sure that the editor setup (TS, ESLint) works everywhere, not just in a subset of source files.

This PR corrects that configuration and fixes a few issues it found.

### DIFF
--- a/src/components/dashboard/components/ZwValuesView.vue
+++ b/src/components/dashboard/components/ZwValuesView.vue
@@ -4,7 +4,7 @@
 			v-model="query"
 			class="zw-values__search"
 			placeholder="Filter values…"
-			aria-label="Filter values"
+			ariaLabel="Filter values"
 		/>
 
 		<ZwValueGroup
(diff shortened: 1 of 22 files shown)


src/components/dashboard/components/ZwValuesView.vue
```diff
@@ -4,7 +4,7 @@
 			v-model="query"
 			class="zw-values__search"
 			placeholder="Filter values…"
-			aria-label="Filter values"
+			ariaLabel="Filter values"
```
robertsLando: not sure why this change, I see other props use the `-` as word separator

src/components/dashboard/layout/ZwTopbar.vue
```diff
@@ -25,7 +25,7 @@
 			<ZwSearchInput
 				:model-value="query"
 				:placeholder="searchPlaceholder"
-				:aria-label="'Search devices'"
+				:ariaLabel="'Search devices'"
```
robertsLando: same

src/components/dashboard/components/ZwValuesView.vue
AlCalzone: this change was unnecessary.

PR #4701, opened 2026-06-29
feat(ui): implement new Dropdown and NumericInput controls

AlCalzone:
This PR implements two specialized controls in the new dashboard design:

## Numeric input
Includes a button to submit after changing, also reacts to up/down arrows, <kbd>ENTER</kbd> and <kbd>ESC</kbd> (resets).

image

## Dropdown

Closed:
image

Open:
image

With a filter for more than 6 entries:
image

And a manual input variant:
image
...which doubles as a filter:
image

### DIFF
--- a/src/components/dashboard/atoms/ZwDropdown.vue
+++ b/src/components/dashboard/atoms/ZwDropdown.vue
@@ -138,8 +138,6 @@ const contentId = `zw-dd-${useId()}`
 
 usePopoverFallback({ open, contentId, placement: 'bottom-start', offsetPx: 4 })
 
-// Autocomplete shows a filter when options exceed the plain threshold.
-// Combobox (allowManual) always shows the filter since it doubles as entry.
 const showFilter = computed(
 	() => props.allowManual || props.options.length > OPTIONS_FILTER_THRESHOLD,
 )
@@ -162,11 +160,9 @@ const filteredOptions = computed(() => {
 	)
 })
 
-// Number typed in the filter field — only meaningful in combobox mode.
 const manualNum = computed<number | null>(() => {
 	if (!props.allowManual) return null
 	const q = filter.value.trim()
-	// Accept integers and decimals — allowManualEntry params may have a decimal step.
 	if (!/^-?\d+(\.\d+)?$/.test(q)) return null
 	return Number(q)
 })
@@ -176,12 +172,9 @@ const manualValid = computed(() => {
 	if (n === null) return false
 	if (props.min !== undefined && n < props.min) return false
 	if (props.max !== undefined && n > props.max) return false
-	// Don't offer "set custom" for a value already in the list.
 	return !props.options.some((o) => String(o.value) === String(n))
 })
 
-// Selected value as a string key — options compare against this, so it's
-// coerced once rather than per option per render.
 const selectedKey = computed(() => String(props.modelValue))
 
 const curOpt = computed(() =>
@@ -216,10 +209,7 @@ function onFilterKeyDown(e: KeyboardEvent): void {
 	}
 }
 
-// On open: match panel width to trigger (min 220px) and focus the filter.
-// On close: clear the filter so it doesn't persist across sessions.
-// Registered after usePopoverFallback so its watcher fires first and the
-// element is already positioned when we set width here.
+// Match panel width to trigger on open; clear filter on close.
 watch(
 	() => open.value,
 	(isOpen) => {
--- a/src/components/dashboard/atoms/ZwDropdown.vue
+++ b/src/components/dashboard/atoms/ZwDropdown.vue
@@ -1,0 +1,447 @@
+<template>
+	<Popover.Root v-model="open" :id="contentId">
+		<Popover.Activator
+			as="button"
+			type="button"
+			class="zw-dropdown__trigger zw-focus-ring"
+			:disabled="disabled"
+			:title="triggerLabel"
+		>
+			<span class="zw-dropdown__trigger-text">{{ triggerLabel }}</span>
+			<ChevronDownIcon
+				class="zw-dropdown__chevron"
+				:size="ICON_SIZE.caret"
+			/>
+		</Popover.Activator>
+
+		<Popover.Content as="div" class="zw-dropdown__panel">
+			<div v-if="showFilter" class="zw-dropdown__filter-row">
+				<component
+					:is="allowManual ? EditIcon : SearchIcon"
+					class="zw-dropdown__filter-icon"
+					:size="ICON_SIZE.caret"
+				/>
+				<input
+					ref="filterInputRef"
+					v-model="filter"
+					type="text"
+					class="zw-dropdown__filter-input"
+					:placeholder="filterPlaceholder"
+					@keydown="onFilterKeyDown"
+				/>
+			</div>
+
+			<div class="zw-dropdown__list" role="listbox">
+				<!-- Combobox manual-entry confirm row -->
+				<button
+					v-if="manualValid"
+					type="button"
+					class="zw-dropdown__item zw-dropdown__item--custom"
+					role="option"
+					@click="pick(manualNum!)"
+				>
+					<!-- Spacer preserves alignment with the check column -->
+					<span
+						class="zw-dropdown__check-spacer"
+						aria-hidden="true"
+					/>
+					<span class="zw-dropdown__item-body">
+						Set custom value
+						<span class="zw-dropdown__manual-value">
+							{{ manualNum }}{{ unit ? ` ${unit}` : '' }}
+						</span>
+					</span>
+					<span class="zw-dropdown__enter-hint">Enter ↵</span>
+				</button>
+
+				<button
+					v-for="opt in filteredOptions"
+					:key="String(opt.value)"
+					type="button"
+					class="zw-dropdown__item"
+					:class="{
+						'zw-dropdown__item--selected': isSelected(opt.value),
+					}"
+					role="option"
+					:aria-selected="isSelected(opt.value)"
+					@click="pick(opt.value)"
+				>
+					<CheckIcon
+						class="zw-dropdown__check"
+						:class="{
+							'zw-dropdown__check--invisible': !isSelected(
+								opt.value,
+							),
+						}"
+						:size="ICON_SIZE.dense"
+						aria-hidden="true"
+					/>
+					<span class="zw-dropdown__item-num">[{{ opt.value }}]</span>
+					<span class="zw-dropdown__item-body">{{ opt.label }}</span>
+				</button>
+
+				<div
+					v-if="filteredOptions.length === 0 && !manualValid"
+					class="zw-dropdown__empty"
+				>
+					{{ emptyLabel }}
+				</div>
+			</div>
+		</Popover.Content>
+	</Popover.Root>
+</template>
+
+<script setup lang="ts">
+import { computed, ref, watch, useId } from 'vue'
+import { Popover } from '@vuetify/v0'
+import {
+	CheckIcon,
+	ChevronDownIcon,
+	EditIcon,
+	SearchIcon,
+	ICON_SIZE,
+} from '@/lib/icons'
+import { usePopoverFallback } from '@/lib/popover-fallback.ts'
+
+type OptionValue = number | string | boolean
+
+interface DropdownOption {
+	value: OptionValue
+	label: string
+}
+
+const props = withDefaults(
+	defineProps<{
+		modelValue: OptionValue | null
+		options: DropdownOption[]
+		allowManual?: boolean
+		min?: number
+		max?: number
+		unit?: string
+		disabled?: boolean
+	}>(),
+	{ allowManual: false, disabled: false },
+)
+
+const emit = defineEmits<{ 'update:modelValue': [OptionValue] }>()
+
+const open = ref(false)
+const filter = ref('')
+const filterInputRef = ref<HTMLInputElement | null>(null)
+const contentId = `zw-dd-${useId()}`
+
+usePopoverFallback({ open, contentId, placement: 'bottom-start', offsetPx: 4 })
+
+// Autocomplete shows a filter when options exceed the plain threshold.
+// Combobox (allowManual) always shows the filter since it doubles as entry.
+const showFilter = computed(() => props.allowManual || props.options.length > 6)
+
+const filterPlaceholder = computed(() => {
+	if (!props.allowManual) return 'Filter…'
+	if (props.min !== undefined && props.max !== undefined)
+		return `Enter a value (${props.min}–${props.max})…`
+	return 'Enter a value…'
+})
+
+const filteredOptions = computed(() => {
+	const q = filter.value.trim()
+	if (!q) return props.options
+	const lower = q.toLowerCase()
+	return props.options.filter(
+		(o) =>
+			o.label.toLowerCase().includes(lower) ||
+			String(o.value).includes(q),
+	)
+})
+
+// Integer typed in the filter field — only meaningful in combobox mode.
+const manualNum = computed<number | null>(() => {
+	if (!props.allowManual) return null
+	const q = filter.value.trim()
+	if (!/^-?\d+$/.test(q)) return null
+	return Number(q)
+})
+
+const manualValid = computed(() => {
+	const n = manualNum.value
+	if (n === null) return false
+	if (props.min !== undefined && n < props.min) return false
+	if (props.max !== undefined && n > props.max) return false
+	// Don't offer "set custom" for a value already in the list.
+	return !props.options.some((o) => String(o.value) === String(n))
+})
+
+// Selected value as a string key — options compare against this, so it's
+// coerced once rather than per option per render.
+const selectedKey = computed(() => String(props.modelValue))
+
+const curOpt = computed(() =>
+	props.options.find((o) => String(o.value) === selectedKey.value),
+)
+
+const triggerLabel = computed(() => {
+	if (curOpt.value) return `[${curOpt.value.value}] ${curOpt.value.label}`
+	if (props.modelValue !== null && props.modelValue !== undefined)
+		return `${props.modelValue}${props.unit ? ` ${props.unit}` : ''} · custom`
+	return '—'
+})
+
+const emptyLabel = computed(() =>
+	props.allowManual ? 'No preset matches — enter a number' : 'No matches',
+)
+
+function isSelected(value: OptionValue): boolean {
+	return String(value) === selectedKey.value
+}
+
+function pick(value: OptionValue): void {
+	open.value = false
+	emit('update:modelValue', value)
+}
+
+function onFilterKeyDown(e: KeyboardEvent): void {
+	if (e.key === 'Enter' && manualValid.value && manualNum.value !== null) {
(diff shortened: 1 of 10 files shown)


src/components/dashboard/components/ZwValueRow.vue
```diff
@@ -313,11 +358,9 @@ function coerce(raw: unknown): unknown {
...
 function commit(value: unknown) {
 	menuOpen.value = false
-	// Don't optimistically pin the value. On a failed or un-acked write the
-	// device never confirms, so a pinned draft would keep showing the attempted
-	// value while the error toast says it failed. Show the live value instead;
-	// `busy` covers the in-flight window and the watch reflects the real result.
-	draft.value = null
+	sendFailed.value = false
+	// Show the value optimistically; the send-completion watch reverts on failure.
+	draft.value = value
```
robertsLando: **[Blocker] · Correctness**

`commit()` now pins `draft.value = value` for *every* param. A write-only param (`writeable && !readable`, kind `number`/`level`) never reports a value back, so neither the `() => props.param.value` watch nor the success branch ever clears the draft.

**Why:** After a *successful* write the row stays dirty forever — the apply button / optimistic label stick permanently. The pre-PR code (`draft.value = null`) did not regress this. (The sibling `08-values-table` branch already guards this exact line with `props.param.readable ? value : null`.)

**Fix:** Only hold the optimistic draft when `props.param.readable`, else set `null`.

src/views/ControlPanelNew.vue
```diff
@@ -65,10 +80,13 @@ async function onAction(device: Device, action: DeviceAction) {
 	const key = actionPendingKey(device, action)
 	if (key) setPending(key, true)
 	try {
-		await app.apiRequest(req.api, req.args, {
+		const response = await app.apiRequest(req.api, req.args, {
```
robertsLando, line 81: **[Major] · Operability**

No deadline on `app.apiRequest`. If the promise never settles (WS disconnect mid-write, driver/node stall), the `finally` never runs, `setPending(key,false)` never fires, and the row is stuck in "Setting…" with no escape.

**Why:** Z-Wave transport stalls are a real production path (node offline / driver hang during a write).

**Fix:** Race the request against a timeout (~30s) and treat the timeout as a failure result.

src/views/ControlPanelNew.vue
```diff
@@ -65,10 +80,13 @@ async function onAction(device: Device, action: DeviceAction) {
...
 			infoSnack: false,
 			errorSnack: true,
 		})
+		// Record the outcome before clearing pending: the row's watch fires on
+		// the pending flip and reads the result, so it must be set first.
+		if (key) setResult(key, isRequestSuccess(req.api, response))
```
robertsLando: **[Major] · Design / API**

Failure detection depends on an implicit temporal ordering split across two files: `setResult` must run before `setPending(key,false)` so the row's `watch(sending)` (ZwValueRow.vue:341-343) reads the result synchronously on the pending flip.

**Why:** It works today (the read-before-clear is race-free within the synchronous tick), but it is an undocumented cross-file invariant — moving `setResult` into a `finally`, or batching state, silently breaks failure-revert with no type/test signal.

**Fix:** Carry the outcome *with* the pending transition — a single `Map<string,'pending'|'ok'|'fail'>` flip instead of two coupled maps.

src/components/dashboard/components/ZwValueRow.vue
```diff
@@ -296,14 +328,27 @@ const rangeHint = computed(() => {
...
 )
 
+// On send completion (pending cleared): the recorded outcome is readable
+// synchronously. On failure, revert the optimistic draft and flag the error.
+watch(sending, (isNowSending, wasSending) => {
+	if (!wasSending || isNowSending) return
+	if (results.value.get(setKey.value) === false) {
+		draft.value = null
+		sendFailed.value = true
+		setTimeout(() => {
```
robertsLando: **[Minor] · Correctness / Operability**

The 3s error timer is untracked. Back-to-back failures within 3s let the first timer clear the *newer* error early; the timer is also never cleared on unmount (detached-ref write). The `3000` literal is also an unexplained magic number.

**Why:** Error indicator flashes off prematurely on rapid successive failures; orphaned callback on unmount.

**Fix:** Store the timer id, `clearTimeout` it before each new schedule and in `onBeforeUnmount`; extract `3000` as a named constant.

src/components/dashboard/atoms/ZwDropdown.vue
```diff
@@ -0,0 +1,447 @@
...
+			o.label.toLowerCase().includes(lower) ||
+			String(o.value).includes(q),
+	)
+})
+
+// Integer typed in the filter field — only meaningful in combobox mode.
+const manualNum = computed<number | null>(() => {
+	if (!props.allowManual) return null
+	const q = filter.value.trim()
+	if (!/^-?\d+$/.test(q)) return null
```
robertsLando: **[Minor] · Correctness**

`manualNum` uses `/^-?\d+$/`, so manual entry rejects decimals — but a writeable numeric param with `allowManualEntry` can carry a decimal `step`.

**Why:** Valid decimal custom values cannot be entered via the combobox.

**Fix:** Allow an optional fractional part (e.g. `/^-?\d+(\.\d+)?$/`) and parse with `Number`.

src/lib/valueGroups.ts
```diff
@@ -145,15 +145,15 @@ function formatValue(v: ZUIValueId, kind: ValueParamKind): string {
...
+// Keeps each row's `param` referentially stable while unchanged, so Vue can skip
+// re-rendering rows that didn't change.
 const paramCache = new WeakMap<ZUIValueId, ValueParam>()
 
 function projectParam(v: ZUIValueId): ValueParam {
 	const cached = paramCache.get(v)
-	if (cached) return cached
+	// Compare the value too: the store can mutate `.value` in place (same object),
+	// so the key alone would return a stale param.
+	if (cached && Object.is(cached.value, v.value)) return cached
```
robertsLando, line 154: **[Minor] · Correctness**

The cache invalidates on `Object.is(cached.value, v.value)` only, but `ValueParam` also projects `min/max/step/default/unit/options/display/modified` from `v`.

**Why:** If metadata changes in place while `value` is unchanged, a stale `param` is returned.

**Fix:** Fine if metadata is immutable per `v`; otherwise also key on the projected fields.

src/lib/device-actions.ts
```diff
@@ -99,3 +103,21 @@ export function dispatchAction(
...
+	success: boolean
+	result?: unknown
+}
+
+export function isRequestSuccess(api: string, response: ApiResponse): boolean {
+	if (!response.success) return false
+	// `writeValue` carries a SetValue status; Working/Success/Unsupervised count
+	// as accepted. Other APIs report only the top-level success flag.
+	if (api === 'writeValue' && response.result) {
+		return setValueWasUnsupervisedOrSucceeded(
```
robertsLando, line 116: **[Minor] · Operability**

`isRequestSuccess` reduces the Z-Wave `SetValueResult` to a boolean; the actual failure code (NoAck / Fail / unconfirmed) is discarded, so "Could not be set" is the only row-level signal.

**Why:** A user debugging a non-responding device cannot distinguish transport failure from protocol rejection.

**Fix:** Store a coarse failure category (or the raw status) in the result map instead of a bare boolean.

src/views/ControlPanelNew.vue
```diff
@@ -48,13 +53,23 @@ function appInstance(): AppLike | null {
 const pending = shallowRef<ReadonlySet<string>>(new Set())
 provide(DeviceActionPendingKey, pending)
 
+// Last outcome per key, so a row can tell a rejected write from a successful one.
+const result = shallowRef<ReadonlyMap<string, boolean>>(new Map())
```
robertsLando: **[Minor] · Operability**

The `result` map accretes one entry per distinct write target and is never evicted.

**Why:** Bounded by the dataset so not a true leak, but asymmetric with the pending set and unbounded over a long session.

**Fix:** Delete the entry in the `finally` block after clearing pending, so cleanup mirrors the pending set.

src/lib/device-actions.ts
```diff
@@ -99,3 +103,21 @@ export function dispatchAction(
...
 	return dispatcher(device, action)
 }
+
+// Shape the UI reads from an `apiRequest` response.
+export interface ApiResponse {
+	success: boolean
+	result?: unknown
+}
+
+export function isRequestSuccess(api: string, response: ApiResponse): boolean {
```
robertsLando, line 112: **[Minor] · Tests**

`isRequestSuccess` is a new pure, testable function added to a file that already has a test suite, but no cases were added.

**Why:** Special-case writeValue logic can break silently when changed.

**Fix:** Cover: non-writeValue failure, writeValue success/fail `SetValueResult`, and undefined result.

src/components/dashboard/atoms/ZwDropdown.vue
```diff
@@ -0,0 +1,447 @@
...
+const open = ref(false)
+const filter = ref('')
+const filterInputRef = ref<HTMLInputElement | null>(null)
+const contentId = `zw-dd-${useId()}`
+
+usePopoverFallback({ open, contentId, placement: 'bottom-start', offsetPx: 4 })
+
+// Autocomplete shows a filter when options exceed the plain threshold.
+// Combobox (allowManual) always shows the filter since it doubles as entry.
+const showFilter = computed(() => props.allowManual || props.options.length > 6)
```
robertsLando: **[Minor] · Readability**

Magic number `6` for the filter-visibility threshold is unexplained.

**Why:** Future readers cannot tell the intent or how to tune it.

**Fix:** Extract as a named constant (e.g. `OPTIONS_FILTER_THRESHOLD`).

src/components/dashboard/atoms/ZwDropdown.vue
```diff
@@ -0,0 +1,447 @@
...
+			return
+		}
+		const c = document.getElementById(contentId)
+		const a = document.querySelector<HTMLElement>(
+			`[popovertarget="${contentId}"]`,
+		)
+		if (a && c) {
+			c.style.setProperty(
+				'min-width',
+				`${Math.max(a.offsetWidth, 220)}px`,
```
robertsLando: **[Minor] · Readability**

Magic number `220` for the minimum panel width is unexplained.

**Why:** Appears arbitrary; hard to find/tune.

**Fix:** Extract as a named constant (e.g. `MIN_PANEL_WIDTH_PX`).

src/components/dashboard/atoms/ZwDropdown.vue
```diff
@@ -0,0 +1,447 @@
...
+	value: OptionValue
+	label: string
+}
+
+const props = withDefaults(
+	defineProps<{
+		modelValue: OptionValue | null
+		options: DropdownOption[]
+		allowManual?: boolean
+		min?: number
```
robertsLando, line 124: **[Minor] · Design / API**

`min` / `max` / `unit` props are only meaningful in combobox (`allowManual`) mode; in plain/autocomplete mode they are inert, with nothing signaling the coupling to callers.

**Why:** Conditional-validity props are an API smell.

**Fix:** Acceptable given shared scaffold, but document on the props that they apply only with `allowManual`.

src/components/dashboard/atoms/ZwNumericInput.vue
```diff
@@ -0,0 +1,232 @@
...
+import {
+	CheckIcon,
+	ChevronDownIcon,
+	ChevronUpIcon,
+	ICON_SIZE,
+} from '@/lib/icons'
+
+const props = withDefaults(
+	defineProps<{
+		modelValue: number | string
```
robertsLando: **[Nit] · Design / API**

`modelValue` accepts `number | string` but the emit is always `[string]`; a parent binding a `number` ref has it silently reassigned to `string` after the first edit.

**Why:** Asymmetric v-model contract (ZwValueRow sidesteps it via `String(...)`, but the public API is inconsistent).

**Fix:** Narrow the input type to `string` — callers already stringify.

src/lib/device-actions.ts
```diff
@@ -99,3 +103,21 @@ export function dispatchAction(
...
+	if (!response.success) return false
+	// `writeValue` carries a SetValue status; Working/Success/Unsupervised count
+	// as accepted. Other APIs report only the top-level success flag.
+	if (api === 'writeValue' && response.result) {
+		return setValueWasUnsupervisedOrSucceeded(
+			response.result as SetValueResult,
```
robertsLando, line 117: **[FYI] · Operability**

Unsupervised writes (no device ack) are equated with success via `setValueWasUnsupervisedOrSucceeded`; the row shows no error despite no confirmation.

**Why:** Impossible to distinguish "confirmed OK" from "sent, nobody home" — likely intentional given how common unsupervised Z-Wave commands are.

**Fix:** Consider a neutral "sent (unconfirmed)" indicator to aid debugging unresponsive devices.

src/views/ControlPanelNew.vue
robertsLando: **[FYI] · Operability**

No `console.warn`/debug trace when `isRequestSuccess` returns false.

**Why:** When a physical write fails, zero diagnostic context (key, api, raw response) is emitted.

**Fix:** Add a single `console.warn('[ControlPanelNew] write failed', { key, api: req.api, response })` before `setResult`.

robertsLando (commented):
## Deep review — feat(ui): new Dropdown and NumericInput controls

**Verdict:** Needs work — one confirmed bug + two robustness gaps on the device-write path; otherwise a clean, well-fitted change.

**Top 3 risks**
1. Write-only numeric / `level` params stay permanently "dirty" after a *successful* write (`commit()` pins the optimistic draft unconditionally).
2. A write that never settles hangs the row in "Setting…" forever (no deadline on `apiRequest`).
3. Failure UX collapses every failure reason to one boolean, and the result-map ordering contract is fragile + unbounded.

**Strengths:** custom atoms slot cleanly into the existing dashboard-neo atom system (same `@vuetify/v0` headless Popover + `--zw-*` tokens as `ZwToggleMenu`); `usePopoverFallback` is *extended* (backward-compatible defaults), not forked; `DeviceActionResultKey` mirrors the existing pending-inject pattern; failure logic correctly reuses `@zwave-js/cc`'s `setValueWasUnsupervisedOrSucceeded`. DRY: clean.

**Theme:** all three top risks live in the same new surface — the optimistic-write → result → clear lifecycle. The happy path is solid; failure / never-settles / write-only cases are under-specified.

**Coverage:** Correctness, DRY, Design/API, Operability, Tests, Readability. Skipped Security (no auth/parse/network surface) and Performance (no hot path). Pre-flight: approach validated — the custom atoms are the deliberate dashboard-neo pattern, not wheel-reinvention.

**FYI (not anchored to a diff line) — docs:** the new ENTER/ESC/arrow affordances on these controls may warrant a dashboard-docs sync (out of scope for this PR).

src/views/ControlPanelNew.vue
AlCalzone: I don't think so. Setting a value on a sleeping device can take hours to go through - but an indication that the value update is still queued would be good to have.

src/lib/device-actions.ts
AlCalzone: No, I think this is good enough for 99% of use cases. All others can look into the logs.

src/lib/device-actions.ts
AlCalzone: No I think this is fine - this is the nature of Z-Wave.

src/views/ControlPanelNew.vue
AlCalzone: No, I think the Z-Wave logs are enough.

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: Fixed in e05c8805: `commit()` now pins the optimistic draft only when `props.param.readable`, otherwise sets `null` — so write-only params no longer stick dirty after a successful write.

src/views/ControlPanelNew.vue
AlCalzone: Hardened in e05c8805: the two writes now live in a single `completeAction(key, ok)` so they can't drift apart across files. Kept the two maps rather than one status map — merging would make poll/refresh/cc keys accrete status entries too (worsening the growth point). The row's `watch` is `flush:'pre'` (async), so it reads both after they're set in the same tick; strict ordering isn't load-bearing, but the encapsulation prevents an `await` ever creeping between them.

src/components/dashboard/components/ZwValueRow.vue
AlCalzone: Fixed in e05c8805: the timer id is stored and `clearTimeout`'d before each reschedule and in `onBeforeUnmount`; `3000` is now `SEND_ERROR_MS`.

src/components/dashboard/atoms/ZwDropdown.vue
AlCalzone: Fixed in e05c8805: regex is now `/^-?\d+(\.\d+)?$/`, so decimal manual entry is accepted.

src/lib/valueGroups.ts
AlCalzone: Not a bug: `metadataUpdated` → store `setValue` does `node.values.splice(index, 1, valueId)`, replacing the value object — so a metadata change is a *new* `WeakMap` key → cache miss → rebuild. Only `.value` mutates in place (`valueUpdated`), which the `Object.is` guard catches. Metadata is effectively immutable per `v`, so no stale projection. Leaving unresolved for your call.

src/lib/device-actions.ts
AlCalzone: Added in e05c8805: `isRequestSuccess` tests covering non-writeValue failure, non-writeValue success, writeValue with no result, and writeValue success/fail `SetValueResult` (real `SetValueStatus` values).

src/components/dashboard/atoms/ZwDropdown.vue
AlCalzone: Fixed in e05c8805: extracted as `OPTIONS_FILTER_THRESHOLD`.

src/components/dashboard/atoms/ZwDropdown.vue
AlCalzone: Fixed in e05c8805: extracted as `MIN_PANEL_WIDTH_PX`.

src/components/dashboard/atoms/ZwDropdown.vue
AlCalzone: Documented in e05c8805: added a comment on the props that `min`/`max`/`unit` apply only with `allowManual` (combobox manual entry).

src/components/dashboard/atoms/ZwNumericInput.vue
AlCalzone: Fixed in e05c8805: narrowed `modelValue` to `string` — the caller already passes `String(...)` and the emit is `[string]`.

robertsLando (approved):

src/views/ControlPanelNew.vue
AlCalzone: These now get deleted again on the next tick, so the pre-flush watchers can read the outcome.

PR #4706, opened 2026-06-30
feat(ui): add action list components and rework Debug tab

AlCalzone:
Introduce ZwActionList and ZwActionBtn — reusable action-row components with idle→busy→done transient states.

Replace the Debug tab's property table (for nodes) with four diagnostic action rows: Ping, Re-interview, Rebuild routes, and Export JSON (with UI/Driver sub-buttons). Controller Debug tab stays unchanged (communication stats).

Add `export-ui` device action type for local node JSON export.

image


AlCalzone:
Copilots feedback got lost on the way to this PR, but the last commit addresses it.

PR #4715, opened 2026-07-02
fix(ui): show group id instead of redundant endpoint in association dropdown

### LINKED ISSUE #4668: Group Association Options Lacking Detail
### Checklist

### Deploy method

Docker

### Z-Wave JS UI version

zwave-js-ui: 11.19.1

### ZwaveJS version

zwave-js: 15.24.2

### Describe the bug

I am trying to configure basic set group 3 so that I can receive double tap events to action upon with home assistant. This is a classic older GE dimmer motion switch that does not support central scene. However, the options are not labled!!

Image

### To Reproduce

Go to any device under Groups and try to add a group :)

### Expected behavior

I would expect more information such as Group ID number and better description. In this case, should see the following:

Group 1: Lifeline (Reports status to hub). 
Group 2: Basic Set (Single Tap / Local Load). 
Group 3: Basic Set (Double Tap).

### Additional context

_No response_

robertsLando:
## Summary

Fixes #4668.

In the "New Association" dialog, each group option in the **Group** dropdown showed the node endpoint as its second (subtitle) row. This is redundant because the endpoint is already chosen in the **Node Endpoint** field right above it, and — as noted in the issue discussion — it provided no way to tell which numeric group you were selecting when multiple groups share the same label (e.g. two "Basic Set" groups).

This replaces the redundant endpoint subtitle with the **group id**, which was previously not surfaced anywhere in the association UI.

## Changes

- `DialogAssociation.vue`: group dropdown item subtitle now shows `Group <id>` instead of the endpoint label.
- Removed the now-unused `getEndpointLabel` method.

## Before / After

Before: second row repeated the already-selected endpoint (e.g. "Root Endpoint").
After: second row shows the group id (e.g. "Group 3").

### DIFF
--- a/src/components/dialogs/DialogAssociation.vue
+++ b/src/components/dialogs/DialogAssociation.vue
@@ -36,14 +36,7 @@
 										<v-list-item
 											v-bind="props"
 											:title="item.raw.title"
-											:subtitle="
-												item.raw.endpoint >= 0
-													? getEndpointLabel(
-															node,
-															item.raw.endpoint,
-														)
-													: 'No Endpoint'
-											"
+											:subtitle="`Group ${item.raw.value}`"
 										>
 										</v-list-item>
 									</template>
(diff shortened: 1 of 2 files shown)


polskikrol:
Can you also add a column to the table? If you review existing associations, on the node, the group ID is also not surfaced there anywhere.

robertsLando:
Done in dcf91ea8 — added a **Group ID** column to the existing associations table so the numeric group id is now surfaced when reviewing a node's associations (the previous "Group" column only shows the group title, which can be duplicated across groups).

AlCalzone (approved):
LGTM

AlCalzone:
Coveralls is failing again. All tests are green.

PR #4717, opened 2026-07-03
feat: support trusted API on separate network/IP/port

AlCalzone:
This PR adds support for two environment variables that let a host application like Home Assistant establish a secure channel to Z-Wave JS UI's API, without interfering with the user's auth configuration.

This is done by specifying a network/IP/port on which a second API server is started, and an IP allowlist (which IPs may use this). The configuration works entirely through environment variables:
- `TRUSTED_API_LISTEN`: Enables the trusted listener. Three forms are accepted:
  - An absolute path starting with `/`: bind a unix socket at that path. Access control is left to filesystem permissions. A stale socket file from an unclean shutdown is removed automatically on startup.
  - `host:port` with a literal IP (IPv6 in brackets, e.g. `[fd00::1]:8092`): bind exactly that address.
  - `cidr:port` (e.g. `172.30.32.0/23:8092`): resolve the local interface address inside that network at startup and bind to it. This lets the host name a (docker) network without either side knowing the container's dynamic IP. Startup fails if no (or more than one) local address matches.
- `TRUSTED_API_ALLOWED_IPS`: Comma-separated list of IPs/CIDRs (e.g. `172.30.32.2,127.0.0.1`) allowed to connect to a TCP binding. **Required for TCP bindings** — startup fails without it. Ignored for unix sockets.


api/config/app.ts
```diff
@@ -87,3 +88,107 @@ export const sessionSecret: string = resolveSessionSecret()
...
+			// Literal IP, or a network CIDR resolved to the local interface address at bind time
+			host: string | ParsedCidr
+			port: number
+			allowedIps: ParsedCidr[]
+	  }
+	| { kind: 'unix'; path: string }
+
+function parseCidrOrThrow(raw: string, context: string): ParsedCidr {
+	try {
+		return ipaddr.parseCIDR(raw)
```
robertsLando: do we relly need a dep for this?

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,171 @@
...
+		return undefined
+	}
+
+	const server = createServer(app)
+
+	server.on('connection', (socket) => {
+		if (config.kind === 'tcp') {
+			const remote = socket.remoteAddress
+			if (!remote || !isAllowedAddress(remote, config.allowedIps)) {
+				logRejection(remote ?? 'unknown')
```
robertsLando, line 259: I'm wondering in case of brute forece attacks if this could fill up the log memory somehow

api/config/app.ts
robertsLando, line 1: I think this changes should live in an utils file (dedicated)

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,171 @@
...
+				socket.destroy()
+				return
+			}
+		}
+		trustedSockets.add(socket)
+	})
+
```
robertsLando, line 266: **[Major] · Operability / Design / Correctness**

`server.on('error')` → `process.exit(1)`, and the handler stays attached for the whole server lifetime. A bind failure (`EADDRINUSE`/`EACCES`/`EADDRNOTAVAIL`) — *or any later runtime error* — on the auth-bypassed side listener tears down the primary Z-Wave gateway that controls physical devices. It also creates two divergent failure paths for one feature: config errors reject/throw at import, bind errors hard-exit, bypassing `www.ts`'s own startup `.catch`. The only breadcrumb is `error.message` — no `code`/`syscall`, and `describeBinding`/`host`/`port` are in scope but unused. *(Flagged independently by 5 of 6 review lenses.)*

**Why:** blast radius (a low-traffic optional listener can kill the whole app) plus thin context makes 3am root-causing slow.

**Fix:** decide deliberately whether a side-listener bind failure should be fatal; if yes, reject the returned promise so `startServer` handles it uniformly — it already prints friendly `EACCES`/`EADDRINUSE` messages (app.ts ~284-303) that could be reused. After `listen` succeeds, drop or downgrade this handler to log-only, and log `code`/`syscall`/the resolved bind address.

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,171 @@
...
+	let host: string | undefined
+	if (config.kind === 'unix') {
+		// Node only unlinks the socket file on server.close(), so a previous
+		// unclean shutdown leaves a stale file that would fail the bind
+		await unlink(config.path).catch((error) => {
+			if (error.code !== 'ENOENT') throw error
+		})
+		await new Promise<void>((resolve) =>
```
robertsLando: **[Major] · Security**

The unix socket is bound via `server.listen(path)` with no `chmod`; its mode is whatever the ambient umask yields (0777 under umask 000, common in root containers) — potentially world-writable.

**Why:** any local user/process connecting to a world-writable socket gets unauthenticated control of the Z-Wave API (locks, security devices). The docs state "access control is left to filesystem permissions," but the code sets none, so it relies on ambient umask.

**Fix:** explicitly `chmod(path, 0o660)` after bind (fail-closed) and/or document the required umask, rather than trusting the ambient one.

api/app.ts
```diff
@@ -926,7 +947,7 @@ async function parseJWT(req: Request) {
 // middleware to check if user is authenticated
 async function isAuthenticated(req: Request, res: Response, next: () => void) {
 	// if user is authenticated in the session, carry on
-	if (req?.session?.user || !isAuthEnabled()) {
+	if (isTrustedRequest(req) || req?.session?.user || !isAuthEnabled()) {
```
robertsLando, line 948: **[Major] · Tests**

This is the primary REST auth-bypass gate the PR exists to add, and no test exercises `isAuthenticated()` with `isTrustedRequest(req)` true vs false. The trust *primitive* is well tested through a real listener, but its composition here (`isTrustedRequest(req) || session || !isAuthEnabled()`) is not.

**Why:** a wiring bug (wrong branch order, wrong property path, a future refactor) that silently bypasses or fails auth would ship undetected.

**Fix:** add an integration test hitting a protected route over a real trusted listener — expect 200 without a session, and 401/redirect for the same route on a non-trusted connection.

api/app.ts
```diff
@@ -121,9 +126,14 @@ const logger = loggers.module('App')
 
 const verifyJWT = promisify(jwt.verify.bind(jwt))
 
+// The trusted listener skips rate limiting: the host application is trusted,
+// and over a unix socket req.ip is undefined which express-rate-limit rejects
+const skipTrusted = (req: Request) => isTrustedRequest(req)
```
robertsLando, line 131: **[Major] · Tests**

No test asserts that `storeLimiter`/`loginLimiter`/`apisLimiter` actually skip rate limiting for trusted requests *and still enforce it for everyone else*.

**Why:** a wiring bug (e.g. `skip` returning true unconditionally) would silently disable brute-force protection on `/login` for all clients, not just trusted ones — a security regression with no failing test.

**Fix:** add a test that a trusted request bypasses the limiter while a non-trusted one is still limited.

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,171 @@
...
+const logger = module('TrustedListener')
+
+// Sockets accepted by the trusted listener; keyed by the raw net.Socket so
+// both plain requests and websocket upgrades (which reuse it) are covered
+const trustedSockets = new WeakSet<NetSocket>()
+
+// A single page load opens several TCP connections per remote address, so
+// log at most one rejection per address per window instead of one per socket
+const REJECTION_LOG_WINDOW_MS = 60_000
+const lastRejectionLogAt = new Map<string, number>()
```
robertsLando, line 128: **[Minor] · Security / Operability / Correctness**

`lastRejectionLogAt` accumulates one never-evicted entry per distinct rejected remote address. *(Flagged by 3 lenses.)*

**Why:** a TCP binding scanned by many source IPs grows the Map unbounded — a slow memory leak on exactly the attack path it's meant to observe.

**Fix:** evict entries older than the window (sweep on insert) or use a small LRU cap.

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,171 @@
...
+	} else {
+		host =
+			typeof config.host === 'string'
+				? config.host
+				: resolveListenAddress(config.host)
+		await new Promise<void>((resolve) =>
+			server.listen(config.port, host, resolve),
+		)
+	}
+
```
robertsLando, line 298: **[Minor] · Operability**

The startup log for an authentication-bypassed listener is `logger.info`.

**Why:** at default log levels this security-relevant "auth is bypassed" notice can be filtered out, so operators may not realize the listener is active.

**Fix:** raise to `logger.warn`. (Consider also a `warn` when a stale unix socket file is unlinked — currently silent.)

test/lib/trustedListener.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+		const trustedServer = await startTrustedListener(
+			app,
+			tcpConfig('127.0.0.0/8'),
+		)
+		servers.push(trustedServer)
+
+		const socketManager = new SocketManager()
+		socketManager.authMiddleware = (
+			socket,
```
robertsLando, line 163: **[Minor] · Tests**

This test hand-rolls an `authMiddleware` inline instead of exercising the real one in app.ts (`isTrustedSocket(socket.request?.socket)` → `!isAuthEnabled()` → token check), so a bug in the real branch order/wiring isn't caught — it only proves `isTrustedSocket` itself works. Relatedly, the "NOT on main server" assertion (~line 167) only tries the `polling` transport; `websocket` is a distinct engine.io handshake path and is never tried against the main server.

**Why:** the auth-bypass wiring and the main-server enforcement across both transports are the security-critical behaviors, and both are only partially covered.

**Fix:** exercise the real `socketManager.authMiddleware` (export it if needed) and add the `websocket`-against-main-server case.

api/app.ts
```diff
@@ -948,7 +969,12 @@ async function isAuthenticated(req: Request, res: Response, next: () => void) {
 
 // logout the user
 app.get('/api/auth-enabled', apisLimiter, function (req, res) {
-	res.json({ success: true, data: isAuthEnabled() })
+	// Report auth as disabled on the trusted listener so an embedded UI
+	// skips the login screen
+	res.json({
+		success: true,
+		data: isTrustedRequest(req) ? false : isAuthEnabled(),
```
robertsLando, line 972: **[Minor] · Tests**

`/api/auth-enabled` returning `false` for trusted requests (vs real `isAuthEnabled()` otherwise) is untested.

**Why:** an embedded UI relies on this to skip the login screen; a regression here silently breaks that flow.

**Fix:** assert the trusted vs non-trusted response for this route.

robertsLando (commented):
## Deep review — trusted API listener

**Verdict: Ship with minor changes.** The approach is sound and the security model holds up under scrutiny. Trust is decided on the kernel-reported TCP peer address in the `connection` event *before any HTTP parsing*, `X-Forwarded-For`/`TRUST_PROXY` are deliberately excluded, and the `WeakSet<net.Socket>` keyed on the raw socket cleanly covers keep-alive, HTTP/1.1 pipelining, and websocket-upgrade reuse. Verified fail-closed on `remoteAddress === undefined`, and the `::ffff:` IPv4-mapped divergence between allowlist-parse and remote-match is fail-closed (no over-permit path). Backcompat on the main listener is clean — every new branch (`isAuthenticated`, `/api/auth-enabled`, the three rate-limiter skips) evaluates exactly as before.

Review run across 6 lenses (correctness, security, operability, design, tests, DRY). Findings below are robustness / hardening / test-coverage — none challenge the design.

**Top 3 risks**
1. An error on the *optional secondary* listener hard-kills the *primary* gateway via `process.exit(1)`, with divergent failure paths and thin diagnostics (flagged independently by 5 of 6 lenses).
2. The unix socket is created with no explicit mode — under a permissive container umask it can be world-writable, i.e. unauthenticated full API control for any local process, despite the docs claiming access control is left to filesystem permissions.
3. The core auth-bypass wiring in `app.ts` (`isAuthenticated` + rate-limit `skip`) is not integration-tested; only the underlying trust primitive is.

**Strengths:** correct threat model (raw peer address, headers excluded), the WeakSet abstraction reaching both HTTP and the socket.io handshake, fail-fast config validation, and strong helper/config-level test coverage.

**Coverage:** correctness, security, operability, design, tests, DRY. No changed file left uncovered; no Blockers surfaced by any lens.

test/lib/trustedListenerConfig.test.ts
```diff
@@ -0,0 +1,179 @@
...
+		).toThrow(/invalid IP/)
+		expect(() =>
+			parseTrustedListenerConfig({
+				TRUSTED_API_LISTEN: '127.0.0.1:8092',
+				TRUSTED_API_ALLOWED_IPS: '10.0.0.0/99',
+			}),
+		).toThrow(/invalid CIDR/)
+	})
+
+	it('rejects an invalid port', () => {
```
robertsLando, line 110: **[Minor] · Tests**

"rejects an invalid port" only covers `0`; the `> 65535` branch (e.g. `99999`) is never exercised.

**Fix:** add a case for a port above the valid range.

api/lib/SocketManager.ts
```diff
@@ -54,6 +54,15 @@ class SocketManager extends TypedEventEmitter<SocketManagerEventCallbacks> {
 			.on('connection', this._onConnection.bind(this))
 	}
 
+	/**
+	 * Attaches the already-bound socket.io instance to an additional http
+	 * server: connections from it flow into the same io, namespaces and
+	 * middleware
+	 */
+	attachServer(server: HttpServer) {
```
robertsLando, line 62: **[Nit] · Design**

`attachServer` silently assumes `bindServer` already created `this.io` — `this.io.attach` throws on undefined. The ordering is correct in app.ts today, but this is a new public method with an implicit precondition.

**Fix:** guard with a clear error or document the precondition in the JSDoc.

api/config/app.ts
```diff
@@ -87,3 +88,107 @@ export const sessionSecret: string = resolveSessionSecret()
...
+ *    resolved at startup, so the host application can name a (docker)
+ *    network without either side knowing the container's dynamic IP
+ *
+ * A bare port is rejected: the trusted listener is never bound to all
+ * interfaces. TCP bindings additionally require `TRUSTED_API_ALLOWED_IPS`
+ * (comma-separated IPs/CIDRs) because the listener bypasses authentication —
+ * the TCP peer allowlist is the only gate. Throws on any misconfiguration
+ * so the process fails to start instead of silently degrading.
+ */
+export function parseTrustedListenerConfig(env: {
```
robertsLando: **[Minor] · Design**

~100 lines of CIDR/port parsing plus the `ParsedCidr`/`TrustedListenerConfig` types live here in `config/app.ts`, while the matching runtime (`isAllowedAddress`, `resolveListenAddress`) lives in `trustedListener.ts` — one feature split across two files, with the config side far heavier than its thin env-read siblings.

**Why:** minor cohesion cost; a reader must cross-reference both files.

**Fix:** optional — move `parseTrustedListenerConfig` + types into `trustedListener.ts`, leaving only the `export const trustedListener = parseTrustedListenerConfig(...)` binding here. The `resolveSessionSecret` precedent makes the current placement defensible, so this is low priority.

api/config/app.ts
```diff
@@ -87,3 +88,107 @@ export const sessionSecret: string = resolveSessionSecret()
...
+
+	const match = listen.match(/^(?:\[([^\]]+)\]|([^:\][]+)):(\d+)$/)
+	if (!match) {
+		throw new Error(
+			`TRUSTED_API_LISTEN must be a unix socket path, "host:port" or "cidr:port" (IPv6 in brackets), got "${listen}"`,
+		)
+	}
+
+	const rawHost = match[1] ?? match[2]
+	const listenPort = parseInt(match[3], 10)
```
robertsLando: **[FYI] · Correctness**

A leading-zero port (`:080`) parses via `parseInt` to `80` and binds silently — not what the config literally said. Harmless in practice. The regex itself is correct: it parses `[ipv6]:port` / `ipv4:port` / `cidr:port`, rejects a bare port and unbracketed IPv6, and the `1..65535` bound + overflow handling are right. No action expected — noting for the record.

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,171 @@
...
+	server.on('error', (error: NodeJS.ErrnoException) => {
+		logger.error(`Trusted listener error: ${error.message}`)
+		process.exit(1)
+	})
+
+	let host: string | undefined
+	if (config.kind === 'unix') {
+		// Node only unlinks the socket file on server.close(), so a previous
+		// unclean shutdown leaves a stale file that would fail the bind
```
robertsLando: **[FYI] · Correctness**

The stale-socket `unlink` has no liveness check: if another *running* instance owns that unix path, this unlinks its live socket and binds a new one, silently hijacking it. The TOCTOU between `unlink` and `listen` is benign (a real conflict fails the bind → exit), and the `ENOENT` swallow is correct. `server.close()` does unlink the path (test confirms). No action expected — noting for the record.

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,171 @@
...
+ * Resolve the local interface address inside `cidr`, so a host application
+ * can name a (docker) network without knowing the container's dynamic IP
+ */
+export function resolveListenAddress(
+	cidr: ParsedCidr,
+	interfaces = networkInterfaces(),
+): string {
+	const candidates: string[] = []
+	for (const addrs of Object.values(interfaces)) {
+		for (const { address } of addrs ?? []) {
```
robertsLando, line 181: **[FYI] · Correctness**

IPv6 zone-id interface addresses (`fe80::1%eth0`) make `ipaddr.process` throw; the `catch` in `isAllowedAddress` returns false, so such addresses are silently skipped in `resolveListenAddress`. Fine for the docker/private-IP use case this targets. `isAllowedAddress` kind()-guards before `match()` (the `as ipaddr.IPv4` cast is safe) and returns false on unparsable/empty input. No action expected.

api/config/app.ts
AlCalzone: agreed, we should look into just copying the 1-2 utilities we need.

api/lib/trustedListener.ts
AlCalzone: Lets leave this unresolved for now. I'm not yet certain whether the host application or Z-UI should create the unix socket.

api/lib/SocketManager.ts
AlCalzone: lets look into avoiding this implicit dependency.

api/lib/trustedListener.ts
AlCalzone: This should not happen, but worth flagging in the docs that no two instances may share a socket.

api/lib/trustedListener.ts
AlCalzone: We should have some reasonable upper bound, yes.

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,171 @@
...
+
+function logRejection(remote: string) {
+	const now = Date.now()
+	const last = lastRejectionLogAt.get(remote)
+	if (last !== undefined && now - last < REJECTION_LOG_WINDOW_MS) {
+		return
+	}
+	lastRejectionLogAt.set(remote, now)
+	logger.warn(`Trusted listener: rejected connection from ${remote}`)
+}
```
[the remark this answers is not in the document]
AlCalzone: Fixed in c4ddfcb7: the map is now recency-ordered and capped at 100 entries, evicting the least recently seen address.

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,171 @@
...
+		// process() unwraps IPv4-mapped IPv6 (::ffff:a.b.c.d)
+		addr = ipaddr.process(remote)
+	} catch {
+		return false
+	}
+	return allowed.some(
+		([net, prefix]) =>
+			addr.kind() === net.kind() &&
+			(addr as ipaddr.IPv4).match(net as ipaddr.IPv4, prefix),
+	)
```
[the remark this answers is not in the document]
AlCalzone: Fixed in c4ddfcb7 — the ipaddr.js dependency is gone; the new matchesCidr compares raw bytes with an explicit kind check, no casts.

api/app.ts
```diff
@@ -2281,6 +2307,7 @@ process.removeAllListeners('SIGINT')
 async function gracefuShutdown() {
 	logger.warn('Shutdown detected: closing clients...')
 	try {
+		trustedServer?.close() // also unlinks a unix socket file
 		if (gw) await gw.close()
 		await destroyPlugins()
 	} catch (error) {
```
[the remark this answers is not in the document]
AlCalzone: Fixed in c4ddfcb7 — the shutdown handler now awaits the close so the socket file is unlinked before exit.

test/lib/trustedListener.test.ts
```diff
@@ -0,0 +1,171 @@
...
+	}
+}
+
+function serverPort(server: HttpServer): number {
+	return (server.address() as AddressInfo).port
+}
+
+describe('startTrustedListener', () => {
+	const servers: HttpServer[] = []
+	const cleanups: (() => Promise<void>)[] = []
```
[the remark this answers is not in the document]
AlCalzone: Fixed in c4ddfcb7 — the cleanup is now wrapped in a Promise<void> resolved from the close callback.

test/lib/trustedListenerConfig.test.ts
```diff
@@ -0,0 +1,179 @@
...
+			TRUSTED_API_ALLOWED_IPS: '::1',
+		})
+		expect(config).toMatchObject({ kind: 'tcp', host: '::1', port: 8092 })
+	})
+
+	it('parses a cidr:port host as a network to resolve at bind time', () => {
+		const config = parseTrustedListenerConfig({
+			TRUSTED_API_LISTEN: '172.30.32.0/23:8092',
+			TRUSTED_API_ALLOWED_IPS: '172.30.32.2',
+		})
```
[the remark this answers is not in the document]
AlCalzone: Fixed in c4ddfcb7 with a toBeDefined() check plus non-null assertions.

test/lib/trustedListenerConfig.test.ts
```diff
@@ -0,0 +1,179 @@
...
+		if (config.kind === 'tcp') {
+			expect(config.host).toEqual(ipaddr.parseCIDR('172.30.32.0/23'))
+		}
+	})
+
+	it('parses allowlist entries as IPs and CIDRs', () => {
+		const config = parseTrustedListenerConfig({
+			TRUSTED_API_LISTEN: '127.0.0.1:8092',
+			TRUSTED_API_ALLOWED_IPS: ' 172.30.32.2, 10.0.0.0/8 ',
+		})
```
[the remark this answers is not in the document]
AlCalzone: Fixed in c4ddfcb7.

docs/guide/env-vars.md
```diff
@@ -39,3 +39,17 @@ This is the list of the supported environment variables:
...
+
+## Trusted listener (host API)
+
+A host application that embeds or manages Z-Wave JS UI (for example Home Assistant) can enable a second, **unauthenticated** listener. It serves the full application (UI, REST APIs and Socket.IO) with authentication bypassed, so it must only ever be reachable by the host itself. It is configured exclusively via environment variables so that only whoever controls the process environment can enable it:
+
+- `TRUSTED_API_LISTEN`: Enables the trusted listener. Three forms are accepted:
+  - An absolute path starting with `/`: bind a unix socket at that path. Access control is left to filesystem permissions. A stale socket file from an unclean shutdown is removed automatically on startup.
+  - `host:port` with a literal IP (IPv6 in brackets, e.g. `[fd00::1]:8092`): bind exactly that address.
+  - `cidr:port` (e.g. `172.30.32.0/23:8092`): resolve the local interface address inside that network at startup and bind to it. This lets the host name a (docker) network without either side knowing the container's dynamic IP. Startup fails if no (or more than one) local address matches.
+
```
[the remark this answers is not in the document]
AlCalzone: Fixed in c4ddfcb7.

api/config/app.ts
AlCalzone: Dropped in c4ddfcb7 — replaced by a small local api/lib/ipUtils.ts (parse, CIDR match, IPv4-mapped unwrap, format).

api/lib/trustedListener.ts
AlCalzone: Bounded in c4ddfcb7: recency-ordered map capped at 100 entries.

api/config/app.ts
AlCalzone: Done in c4ddfcb7 — the parsing/types moved out of config/app.ts: IP utilities into api/lib/ipUtils.ts, parseTrustedListenerConfig into trustedListener.ts.

api/lib/trustedListener.ts
AlCalzone: Fixed in c4ddfcb7 — bind errors now reject the returned promise (with the error code attached) and surface through startServer's startup handling; after a successful bind the handler is log-only and includes code/syscall/binding.

api/app.ts
AlCalzone: Added in c4ddfcb7 — trustedAppIntegration.test.ts runs the real app over a plain server and a trusted listener and asserts /api/settings succeeds without a session on the trusted one and fails on the other.

api/app.ts
AlCalzone: Added in c4ddfcb7 — the test exhausts the login limiter untrusted (6th request blocked), then verifies the same client IP still passes on the trusted listener.

api/lib/trustedListener.ts
AlCalzone: Fixed in c4ddfcb7: capped at 100 entries with least-recently-seen eviction.

api/lib/trustedListener.ts
AlCalzone: Done in c4ddfcb7 — the startup notice and the stale-socket removal both log at warn now.

test/lib/trustedListener.test.ts
AlCalzone: Done in c4ddfcb7 — the integration test now exercises the real socketManager.authMiddleware (exported from app.ts), and the main server is checked over the websocket transport too.

api/app.ts
AlCalzone: Added in c4ddfcb7 — asserts data:false on the trusted listener and data:true on the main server with auth enabled.

test/lib/trustedListenerConfig.test.ts
AlCalzone: Added in c4ddfcb7.

api/lib/SocketManager.ts
AlCalzone: Guarded in c4ddfcb7 — attachServer now throws "bindServer must be called before attachServer" instead of failing implicitly.

api/config/app.ts
AlCalzone: Done in c4ddfcb7 — moved to trustedListener.ts (parsing/types/env binding) and ipUtils.ts; config/app.ts is back to thin env reads.

api/lib/trustedListener.ts
AlCalzone: Documented in c4ddfcb7 — env-vars.md now states that no two instances may share a socket path.

api/config/app.ts
AlCalzone: Acknowledged — this behavior is unchanged in c4ddfcb7 (a leading-zero port still parses via parseInt and binds as the decimal value).

api/lib/trustedListener.ts
AlCalzone: Acknowledged — c4ddfcb7 preserves this behavior: the new ipUtils parser explicitly rejects zone-id addresses, so they are still skipped in resolveListenAddress. The IPv4 cast is gone; matching is byte-wise with a kind check.

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,314 @@
+import type { Express } from 'express'
+import { createServer, type Server as HttpServer } from 'node:http'
+import type { Socket as NetSocket } from 'node:net'
+import { networkInterfaces } from 'node:os'
+import { unlink } from 'node:fs/promises'
```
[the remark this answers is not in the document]
AlCalzone: Fixed in f91430ff — lstat() before unlink; a non-socket at the path now throws instead of being deleted, with tests for both the stale-socket and regular-file cases.

api/lib/trustedListener.ts
```diff
@@ -0,0 +1,314 @@
...
+		let unlinked = true
+		await unlink(config.path).catch((error) => {
+			unlinked = false
+			if (error.code !== 'ENOENT') throw error
+		})
+		if (unlinked) {
+			logger.warn(
+				`Trusted listener: removed stale socket file ${config.path}`,
+			)
+		}
```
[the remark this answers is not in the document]
AlCalzone: Fixed in f91430ff — see above: the path is only unlinked if lstat() reports a socket, otherwise startup fails.

test/lib/trustedListenerConfig.test.ts
```diff
@@ -0,0 +1,180 @@
...
+		}
+	})
+})
+
+describe('isAllowedAddress', () => {
+	const allowed = [
+		parseCidr('172.30.32.0/23'),
+		parseCidr('127.0.0.1/32'),
+		parseCidr('fd00::/8'),
+	]
```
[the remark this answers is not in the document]
AlCalzone: False positive — the repo compiles with strict: false (no strictNullChecks), so this passes tsc and type-aware lint as-is; eslint even flags the ! assertion as unnecessary here.

test/lib/trustedListenerConfig.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+	it('throws when multiple local addresses match', () => {
+		const multi = {
+			eth0: [{ address: '172.30.33.7' }],
+			eth1: [{ address: '172.30.33.8' }],
+		} as any
+		expect(() =>
+			resolveListenAddress(parseCidr('172.30.32.0/23'), multi),
+		).toThrow(/multiple local addresses/)
+	})
```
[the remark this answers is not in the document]
AlCalzone: False positive — same as above: strictNullChecks is off in this repo, so no assertion is needed (and lint rejects one as unnecessary).

test/lib/trustedListener.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+function tcpConfig(allowedCidr: string): TrustedListenerConfig {
+	return {
+		kind: 'tcp',
+		host: '127.0.0.1',
+		// Ephemeral port, resolved via server.address() after listen
+		port: 0,
+		allowedIps: [parseCidr(allowedCidr)],
+	}
+}
```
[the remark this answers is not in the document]
AlCalzone: False positive — strictNullChecks is off in this repo; the code typechecks as-is and eslint removes the ! as unnecessary.

test/lib/trustedAppIntegration.test.ts
```diff
@@ -0,0 +1,138 @@
...
+		mainServer = createServer(app)
+		await new Promise<void>((resolve) =>
+			mainServer.listen(0, '127.0.0.1', resolve),
+		)
+		trustedServer = (await startTrustedListener(app, {
+			kind: 'tcp',
+			host: '127.0.0.1',
+			port: 0,
+			allowedIps: [parseCidr('127.0.0.0/8')],
+		}))!
```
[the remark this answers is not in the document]
AlCalzone: False positive — same: strictNullChecks is off, so the assertion is unnecessary and rejected by lint.

api/lib/trustedListener.ts
AlCalzone: Fixed in f43580b2 — the socket is now chmod-ed to 0660 (owner+group) right after bind, so the ambient umask can no longer leave it world-writable. Docs note the host must share a user/group with the intended client. Left for you to decide separately whether the host should own socket creation instead.

package-lock.json
robertsLando, line 1: i think this is should be removed from diff

robertsLando (dismissed):

robertsLando (approved):

Issue #4722, opened 2026-07-10
feat: backend refactor


robertsLando:
Requirements: 
- turn on "strict: true" in tsconfig
- split big classes in smaller ones
- add more tests coverage
- export all home assistant management stuff to a dedicated file, maybe like a plugin (may need to extend plugin's capabilities) 
- keep 100% API compatibility (http, websocket)

AlCalzone:
Plan:
# Plan: Backend refactor for issue #4722

## Problem and approach

The backend is concentrated in three very large, weakly tested units:

- `api/lib/ZwaveClient.ts`: 8,973 lines and approximately 253 methods/accessors
- `api/lib/Gateway.ts`: 2,834 lines, including almost all legacy Home Assistant MQTT discovery
- `api/app.ts`: 2,310 lines, 35 HTTP routes, and all Socket.IO request handlers

Enabling TypeScript strict mode currently exposes approximately 596 diagnostics, 86.9% of them in those three files. Backend coverage is approximately 7% and there are no HTTP, Socket.IO, `ZwaveClient`, MQTT-client, or end-to-end Home Assistant contract tests.

The work will therefore be delivered as a linear series of focused dependent pull requests. The first layers characterize every external contract, later layers introduce typed seams and extract responsibilities, and the final layer enables strict mode only after the complete backend passes strict type checking. Every layer must preserve behavior for all existing consumers.

## Confirmed decisions

- Cover both Home Assistant integrations:
  - Legacy MQTT discovery and management currently spread across `Gateway`, `MqttClient`, `ZwaveClient`, and `app.ts`
  - The `@zwave-js/server` lifecycle currently embedded in `ZwaveClient`
- Implement Home Assistant as a built-in lifecycle-managed subsystem with plugin-ready ports. Extending the custom plugin framework and converting Home Assistant into a loadable plugin are follow-up work.
- No API consumer, including the frontend, may need a change anywhere in this stack.
- Preserve existing behavior, including accidental quirks, while recording those quirks for separate follow-up fixes.
- Ratchet backend-wide coverage upward from the verified baseline. Require every new or extracted module to reach at least 90% line and function coverage and 85% branch coverage.
- Use dependent PRs, one session/branch per PR layer.
- If GitHub Stacked PRs lands before this work is started, us it; otherwise use ordinary PRs whose base is the preceding branch.

## Target architecture

```text
api/
  app.ts                               # Composition and middleware order only
  runtime/
[...]

PR #4723, opened 2026-07-10
test(api): characterize HTTP compatibility

AlCalzone:
Adds HTTP contract coverage before the backend is decomposed.

- Locks the 35 routes’ authentication, response shapes, status/header/stream behavior, and side effects.
- Keeps production routing unchanged behind a minimal app-construction seam.
- Review focus: contract fidelity and assertions that target application behavior rather than harness mechanics.

Part of #4722. Stack root on `backend-refactor`.


AlCalzone:
@robertsLando the goal of this series of PRs was to preserve the observable behavior, but I think we should address at least some of the quirks. What do you think?

> /api/store/backup streams ZIP data as Content-Type: application/json

@AlCalzone This just seems wrong. I'd say: fix 
@robertsLando 👍🏼 

> /api/importConfig and /api/snippet can throw when gw itself is absent.
> Empty non-multipart /api/store/upload requests throw before the intended missing-file guard.
> PUT /api/password throws when no session user exists.

@AlCalzone Throwing is never good. I'd say: fix  
@robertsLando 👍🏼

> The sixth failed login returns HTTP 200 with the rate-limit error envelope, not 429.

@AlCalzone I'd say: fix. 
@robertsLando This is by design how the frontend works, it never uses real http status codes but only success true/false in the response. We can easily fix this BTW

> Most failures use HTTP 200 { success: false, message } envelopes.

@AlCalzone Not sure. Probably fix for correctness, given that we're touching this now anyways.
@robertsLando  like said above, yes

> Unknown GET /health/:client values fall through to generic health output.

@AlCalzone Also seems unintended. Fix?
@robertsLando yes

api/app.ts
```diff
@@ -217,6 +217,22 @@ let pluginsRouter: Router
...
+	Driver.enumerateSerialPorts.bind(Driver)
+// Tracks whether `enumerateSerialPorts` currently points at the real
+// production collaborator (true) or a test-injected fake (false). Only
+// read by the `__testHooks` observability seam below - production never
+// consults it - so a test harness's teardown can prove
+// `setEnumerateSerialPorts(undefined)` actually restored the production
+// default instead of leaving the previous test's fake installed, without
+// ever invoking the real enumerator (which would trigger serial/mDNS I/O).
+let enumerateSerialPortsIsProductionDefault = true
+
```
AlCalzone: Can we not move enumerateSerialPorts into another module and mock it during tests (vi.mock), so we don't have to add a flag (enumerateSerialPortsIsProductionDefault) just for tests?

api/app.ts
```diff
@@ -2307,4 +2331,74 @@ for (const signal of ['SIGINT', 'SIGTERM']) {
...
+// (status codes, envelopes, collaborator calls) without touching hardware,
+// so this seam lets them set the same module-level state directly with
+// typed fakes instead. Nothing in the production entrypoint
+// (`api/bin/www.ts`) imports or calls this - it only exists for
+// `test/lib/http/*`.
+export const __testHooks = {
+	setGateway(value: Gateway | undefined) {
+		gw = value
+	},
+	setZniffer(value: ZnifferManager | undefined) {
```
AlCalzone:
```suggestion
	setZnifferManager(value: ZnifferManager | undefined) {
```

vitest.config.ts
```diff
@@ -35,6 +35,25 @@ export default defineConfig({
...
+			// backend tests once, in the full combined run, is enough to both
+			// enforce the thresholds below AND produce the file-level line
+			// data Coveralls needs for `api/**`.
+			thresholds: {
+				'api/**': {
+					statements: 19,
+					branches: 14,
+					functions: 23,
+					lines: 19,
+				},
```
AlCalzone: I don't like this. These numbers look arbitrary and are a pain to maintain later. If we need to enforce coverage thresholds, we should do so through the Coveralls action and have it fail CI when a threshold is not reached.

vitest.config.server.ts
```diff
@@ -0,0 +1,53 @@
...
+			reporter: ['text', 'lcov'],
+			reportsDirectory: './coverage-server',
+			include: ['api/**'],
+			exclude: ['**/*.test.*', 'test/**'],
+			thresholds: {
+				statements: 19,
+				branches: 14,
+				functions: 23,
+				lines: 19,
+			},
```
AlCalzone: See comment in vitest.config.ts

vitest.config.server.ts
```diff
@@ -0,0 +1,53 @@
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+import { defineConfig } from 'vitest/config'
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
+
+/**
+ * Backend-only Vitest config, used by `npm run coverage:server`.
```
AlCalzone: Is there no way to have a single config for both and limit which tests run through globs?

test/lib/http/harness.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+/** Loads the same `jsonStore` singleton instance `api/app.ts` uses. */
+export async function loadJsonStore(): Promise<JsonStoreModule> {
+	ensureTestEnv()
+	if (!jsonStoreModulePromise) {
+		jsonStoreModulePromise = (async () => {
+			const [jsonStoreMod, storeMod] = await Promise.all([
+				import('../../../api/lib/jsonStore.ts'),
+				import('../../../api/config/store.ts'),
```
AlCalzone: Can we avoid all these `../../../...` paths by using package.json `imports`?

test/lib/http/harness.ts
```diff
@@ -0,0 +1,204 @@
...
+		init: (config: unknown) => Promise<unknown>
+		get: (model: { file: string }) => unknown
+		put: (model: { file: string }, data: unknown) => Promise<unknown>
+	}
+	store: Record<string, { file: string; default: unknown }>
+}
+
+interface GatewayModule {
+	closeWatchers: () => void
+}
```
AlCalzone: Why do we have to define these interfaces here? Does `typeof import(...)` not give the exact type without the risk of drifting?

api/app.ts
```diff
@@ -2307,4 +2331,74 @@ for (const signal of ['SIGINT', 'SIGTERM']) {
...
+// `gw`, `zniffer`, `pluginsRouter` and `restarting` are normally populated by
+// `startServer`/`startGateway`, which construct a real `Gateway` backed by a
+// real `ZWaveClient`/`MqttClient` (i.e. real hardware/network I/O). HTTP
+// characterization tests need to drive the routes' observable contract
+// (status codes, envelopes, collaborator calls) without touching hardware,
+// so this seam lets them set the same module-level state directly with
+// typed fakes instead. Nothing in the production entrypoint
+// (`api/bin/www.ts`) imports or calls this - it only exists for
+// `test/lib/http/*`.
+export const __testHooks = {
```
AlCalzone: Could we avoid this entire hook setup by having `app.ts` return a factory function instead if defining module level variables?

test/lib/http/harness.ts
```diff
@@ -0,0 +1,204 @@
...
+ *     own isolated module graph, but hygiene here also protects against
+ *     future pool/isolation changes).
+ */
+import { createServer, type Server as HttpServer } from 'node:http'
+import type { Express } from 'express'
+import supertest from 'supertest'
+import { cleanupTestEnv, ensureTestEnv } from './env.ts'
+import type { FakeGateway } from './fakes.ts'
+
+/** Mirrors `Driver.enumerateSerialPorts`'s signature from `zwave-js`. */
```
AlCalzone: Then why are we not using that upstream type?

test/lib/http/harnessLifecycle.test.ts
```diff
@@ -0,0 +1,82 @@
+import { describe, it, expect } from 'vitest'
+import { existsSync } from 'node:fs'
+import { createHttpHarness } from './harness.ts'
+import { getTestStoreDir } from './env.ts'
+
+/**
+ * Regression coverage for `HttpHarness.close()`'s teardown, isolated in its
```
AlCalzone: Yo dawg, I heard you like testing, so we added tests for your tests...

Really?!

test/lib/http/env.ts
```diff
@@ -0,0 +1,174 @@
...
+ * leftover from another test file in the same worker, ...), the app under
+ * test would silently behave differently than the deterministic contract
+ * these suites characterize - e.g. `sslDisabled` flipping, session cookies
+ * becoming `Secure`, serial port enumeration being skipped, or `tz`/`locale`
+ * echoing an ambient value instead of `undefined`.
+ *
+ * This module MUST be imported (and `ensureTestEnv()` called) before any
+ * dynamic `import()` of `api/app.ts` or its config, so every HTTP test file
+ * gets its own throwaway store directory AND a normalized set of app-facing
+ * env vars, instead of inheriting ambient process state.
```
AlCalzone: Specifying requirements in comments is weak. I think we can also avoid this by turning app.ts into a factory.

test/lib/http/envDotenvIsolation.test.ts
```diff
@@ -0,0 +1,125 @@
+import { describe, it, expect, beforeAll, afterAll } from 'vitest'
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import path from 'node:path'
+import { createHttpHarness, type HttpHarness } from './harness.ts'
+
+/**
+ * Regression for `test/lib/http/env.ts`'s `dotenv`-repopulation fix.
```
AlCalzone: Oh great. Another test for the test suite...

test/lib/http/envIsolation.test.ts
```diff
@@ -0,0 +1,146 @@
+import { describe, it, expect, beforeAll, afterAll } from 'vitest'
+import { createHttpHarness, type HttpHarness } from './harness.ts'
+import { createFakeGateway } from './fakes.ts'
+import { seedUser } from './authHelpers.ts'
+
+/**
+ * Regression coverage for `test/lib/http/env.ts`'s ambient-env-var
```
AlCalzone: Lets get rid of all tests that test the tests. This is ridiculous.

test/lib/http/settings.test.ts
```diff
@@ -0,0 +1,309 @@
...
+		it('returns an empty list without throwing when the enumerator resolves none', async () => {
+			harness.testHooks.setEnumerateSerialPorts(() => Promise.resolve([]))
+
+			const res = await harness.request.get('/api/serial-ports')
+
+			expect(res.status).toBe(200)
+			expect(res.body).toEqual({ success: true, serial_ports: [] })
+		})
+
+		it('preserved quirk: an enumerator rejection is caught and reported as a failed-but-200 envelope with an empty list', async () => {
```
AlCalzone:
```suggestion
		it('an enumerator rejection is caught and reported as a failed-but-200 envelope with an empty list', async () => {
```

test/lib/http/settings.test.ts
```diff
@@ -0,0 +1,309 @@
...
+			expect(res.status).toBe(200)
+			expect(res.body).toEqual({
+				success: false,
+				message:
+					'Gateway is restarting, wait a moment before doing another request',
+			})
+		})
+
+		it('treats an empty body as a forced restart request, echoing the stored settings back', async () => {
+			const res = await harness.request.post('/api/settings').send({})
```
AlCalzone, line 93: @robertsLando is this behavior intentional? Seems weird.

test/lib/http/auth.test.ts
```diff
@@ -0,0 +1,269 @@
...
+			await setSettings(harness, { gateway: { authEnabled: true } })
+			const res = await harness.request.get('/api/auth-enabled')
+			expect(res.status).toBe(200)
+			expect(res.body).toEqual({ success: true, data: true })
+		})
+	})
+
+	describe('POST /api/authenticate', () => {
+		it('logs in with correct username/password, omits passwordHash, sets a session cookie', async () => {
+			await seedUser(harness, 'alice', 'correct horse battery staple')
```
AlCalzone, line 34: Great password choice.

test/lib/http/auth.test.ts
```diff
@@ -0,0 +1,269 @@
...
+			const res = await harness.request.post('/api/authenticate').send({
+				username: 'bob',
+				password: 'wrong-password',
+			})
+
+			// Preserved quirk: authentication failures resolve with HTTP 200,
+			// not 401/403 - the caller must inspect `success`.
+			expect(res.status).toBe(200)
+			expect(res.body).toEqual({
+				success: false,
```
AlCalzone: @robertsLando This just feels wrong. We should look at fixing this behavior in a followup PR.

AlCalzone (commented):
Review pass 1 - more to come.

api/app.ts
AlCalzone: Fixed in c4f9180e — serial enumeration now lives in a mockable module, with no runtime test flag.

api/app.ts
AlCalzone: Fixed in c4f9180e — renamed to `setZnifferManager`.

vitest.config.ts
AlCalzone: Fixed in 8ad491bf — removed the numeric Vitest thresholds; Coveralls remains the coverage gate.

vitest.config.server.ts
AlCalzone: Fixed in 8ad491bf — removed the backend-only numeric thresholds too.

vitest.config.server.ts
AlCalzone: Fixed in 8ad491bf — `coverage:server` now uses the root config with CLI test and coverage globs.

test/lib/http/harness.ts
AlCalzone: Fixed in 55e64a43 — added `#api/*` and `#test/*` package imports and removed the deep paths.

test/lib/http/harness.ts
AlCalzone: Fixed in 55e64a43 — module shapes now come from `typeof` namespace imports.

api/app.ts
AlCalzone: Fixed in c4f9180e — `createApp()` now owns per-instance state and replaces the module-level test hooks.

test/lib/http/harness.ts
AlCalzone: Fixed in c4f9180e — the serial module uses `typeof Driver.enumerateSerialPorts`; the tests infer the mocked function type.

test/lib/http/harnessLifecycle.test.ts
AlCalzone: Removed in 4b30aa2f.

test/lib/http/env.ts
AlCalzone: Fixed in c4f9180e — app state now comes from `createApp()`; the test setup comments were also reduced in 1f5f9cf8.

test/lib/http/envDotenvIsolation.test.ts
AlCalzone: Removed in 4b30aa2f.

test/lib/http/envIsolation.test.ts
AlCalzone: Removed in 4b30aa2f.

test/lib/http/settings.test.ts
AlCalzone: Fixed in 55e64a43 — applied the suggested test name.

test/lib/http/auth.test.ts
AlCalzone: Thanks!

test/lib/http/debug.test.ts
```diff
@@ -0,0 +1,175 @@
...
+					)
+				})
+
+			expect(res.status).toBe(200)
+			expect(res.headers['content-type']).toBe('application/zip')
+			expect(res.headers['content-disposition']).toMatch(
+				/^attachment; filename="zwave-debug-.+\.zip"$/,
+			)
+			expect((res.body as Buffer).subarray(0, 4).toString('hex')).toBe(
+				'504b0304',
```
AlCalzone, line 149: This could use a comment explaining why we expect this.

test/lib/http/routeContract.test.ts
AlCalzone, line 1: This entire test file feels pointless...

test/lib/http/routeContract.test.ts
```diff
@@ -0,0 +1,304 @@
...
+		for (const name of names) {
+			counts.set(name, (counts.get(name) ?? 0) + 1)
+		}
+		const duplicated = [...counts.entries()].filter(
+			([, count]) => count > 1,
+		)
+		expect(duplicated).toEqual([])
+
+		harness.resetState()
+	})
```
AlCalzone: except for these two tests

test/lib/http/store.test.ts
```diff
@@ -0,0 +1,419 @@
...
+						response.on('end', () =>
+							callback(null, Buffer.concat(chunks)),
+						)
+					})
+
+				expect(res.status).toBe(200)
+				expect(res.headers['content-type']).toMatch(/application\/json/)
+				expect(
+					(res.body as Buffer).subarray(0, 4).toString('hex'),
+				).toBe('504b0304')
```
AlCalzone, line 255: Why this expectation? Comment!

test/lib/http/store.test.ts
```diff
@@ -0,0 +1,419 @@
...
+				const res = await harness.request.post('/api/store/upload')
+
+				expect(res.status).toBe(200)
+				expect(res.body).toEqual({
+					success: false,
+					message:
+						"Cannot read properties of undefined (reading '0')",
+				})
+			},
+		)
```
AlCalzone: No this is not a "quirk" to preserve, this is a legitimate bug to fix.

test/lib/http/store.test.ts
```diff
@@ -0,0 +1,419 @@
...
+				const res = await harness.request.get('/api/snippet')
+
+				expect(res.status).toBe(200)
+				expect(res.body).toEqual({
+					success: false,
+					message:
+						"Cannot read properties of undefined (reading 'zwave')",
+				})
+			},
+		)
```
AlCalzone: Crashes are not "quirks" to preserve. Fix this.

test/lib/http/health.test.ts
```diff
@@ -0,0 +1,115 @@
...
+		it(
+			'preserved quirk: an invalid client name sends the 500 error body ' +
+				'then falls through into a second res.send() for the same request',
+			async () => {
+				const res = await harness.request.get('/health/not-a-client')
+
+				expect(res.status).toBe(500)
+				expect(res.text).toBe("Requested client doesn 't exist")
+			},
+		)
```
AlCalzone: This seems wrong. Fix this. Also fix the typo in `doesn 't`.

test/lib/http/harness.ts
```diff
@@ -0,0 +1,117 @@
...
+	})
+
+	function resetState() {
+		instance.setGateway(undefined)
+		instance.setZnifferManager(undefined)
+		instance.setPluginsRouter(undefined)
+		instance.setRestarting(false)
+		enumerateSerialPorts.mockReset()
+		enumerateSerialPorts.mockResolvedValue([])
+	}
```
AlCalzone: Do we really need this? The entire purpose of making `app.ts` a factory was that we can throw away the old state and create it fresh, without needing all these hooks.

api/app.ts
```diff
@@ -116,2195 +123,2316 @@ const multerUpload = multer({
...
-			defaultSnippets.push({ name, content })
+		// Register process handlers only when the app starts
+		process.on('uncaughtException', (reason) => {
+			const stack = (reason as any).stack || ''
+			logger.error(
+				`Unhandled Rejection, reason: ${reason}${stack ? `\n${stack}` : ''}`,
+			)
+		})
+		for (const signal of ['SIGINT', 'SIGTERM']) {
+			process.once(signal as NodeJS.Signals, gracefuShutdown)
```
AlCalzone: **[Major] · Correctness + Operability**

Process/signal handlers register at the *end* of `startServer`, after `await startGateway`; in the base they were module-scope (registered at import, before startup).

**Why:** the whole startup window — cert load, socket setup, snippet load, and the multi-second Z-Wave/MQTT bring-up (the most crash-prone phase) — now runs with no handlers. A SIGTERM mid-boot (`docker stop`) skips `gracefuShutdown` and leaves the serial port uncleanly held; a startup exception hits Node's default exit instead of being logged. Contradicts the behavior-preserving claim.

**Fix:** register this block at the top of `startServer`, before the awaits — not in `createApp()`, since the eager `defaultInstance` would re-introduce import-time registration. Flagged independently by both the correctness and operability passes.

test/lib/http/routeContract.test.ts
```diff
@@ -0,0 +1,304 @@
...
+})
+
+const EXPECTED_REGISTERED_ROUTES: Array<{
+	method: 'get' | 'post' | 'put' | 'delete'
+	path: string
+}> = [
+	{ method: 'get', path: '/api/auth-enabled' },
+	{ method: 'post', path: '/api/authenticate' },
```
AlCalzone, line 10: **[Major] · Tests**

31 routes are guarded by `isAuthenticated`, but only `/api/logout` and `/api/auth-enabled` are exercised with `authEnabled: true` + no session, and this drift detector inspects `{method, path}` only — never the middleware chain.

**Why:** the stated next step is splitting app.ts into sub-routers; a refactor that drops a guard off any of the ~28 uncovered routes passes all 134 tests — the exact auth-source regression this suite claims to lock.

**Fix:** parametrize an `authEnabled`+no-session → generic-error check across the guarded routes (mirroring how route existence is parametrized here), or explicitly scope it to Layer 2 with a tracking note.

api/app.ts
```diff
@@ -116,2195 +123,2316 @@ const multerUpload = multer({
...
+export interface AppInstance {
+	app: Express
+	startServer: (port: number | string, host?: string) => Promise<HttpServer>
+	setGateway(value: Gateway | undefined): void
+	setZnifferManager(value: ZnifferManager | undefined): void
+	setPluginsRouter(value: Router | undefined): void
+	setRestarting(value: boolean): void
+	isRestarting(): boolean
+	loadSnippets(): Promise<void>
+}
```
AlCalzone, line 132: **[Major] · Design / API**

The exported `AppInstance` fuses the production surface (`app`, `startServer`, `loadSnippets`) with four test-only mutators (`setGateway`/`setZnifferManager`/`setPluginsRouter`/`setRestarting`); nothing marks them internal and no TODO points at the Layer-2 removal.

**Why:** they read as legitimate API and become permanent cruft — the precise risk this PR set out to avoid by replacing `__testHooks`.

**Fix:** nest under a single `_test` property or `@internal` JSDoc, with a comment referencing the #4722 seam removal.

api/app.ts
```diff
@@ -116,2195 +123,2316 @@ const multerUpload = multer({
...
+		startZniffer(settings.zniffer)
+		await debugManager.init() // Clean up any old debug temp files
+		await startGateway(settings)
 
-		if (await isSnippet(filePath)) {
-			const content = await readFile(filePath, 'utf8')
-			const name = path.basename(filePath, '.js')
-			defaultSnippets.push({ name, content })
+		// Register process handlers only when the app starts
+		process.on('uncaughtException', (reason) => {
```
AlCalzone: **[Minor] · Correctness + Operability**

`uncaughtException` is registered with `.on` (not `.once`) and no dedup, and top-level `removeAllListeners('SIGINT')` was dropped; a second `startServer` call would stack a permanent extra handler.

**Why:** only `bin/www.ts` calls it once today, so nothing accumulates — but the single-call assumption is now implicit.

**Fix:** register once, or `removeListener` before re-adding, and make the assumption explicit. The dropped `removeAllListeners` is itself fine — no bundled dep registers a SIGINT handler.

api/app.ts
```diff
@@ -116,2195 +123,2316 @@ const multerUpload = multer({
...
-	const stack = (reason as any).stack || ''
-	logger.error(
-		`Unhandled Rejection, reason: ${reason}${stack ? `\n${stack}` : ''}`,
-	)
-})
-
-for (const signal of ['SIGINT', 'SIGTERM']) {
-	process.once(signal as NodeJS.Signals, gracefuShutdown)
-}
+const defaultInstance = createApp()
```
AlCalzone: **[Minor] · Performance + Design**

Eager `const defaultInstance = createApp()` builds a throwaway Express app + `SocketManager` + 3 rate limiters per test file (~9 wasted builds) purely to satisfy the legacy default export.

**Why:** negligible wall-clock, but pure waste plus repeated version log lines on every import; timers are `unref`'d so it isn't a leak.

**Fix:** make the default export lazy, or move the eager call into `bin/www.ts`.

test/lib/http/routeContract.test.ts
```diff
@@ -0,0 +1,304 @@
...
+	{ method: 'post', path: '/api/store-multi' },
+	{ method: 'get', path: '/api/store/backup' },
+	{ method: 'post', path: '/api/store/upload' },
+	{ method: 'get', path: '/api/snippet' },
+
+	{ method: 'get', path: '/api/debug/status' },
+	{ method: 'post', path: '/api/debug/start' },
+	{ method: 'post', path: '/api/debug/stop' },
+	{ method: 'post', path: '/api/debug/cancel' },
+]
```
AlCalzone: **[Minor] · Design + Readability**

`ROUTES` and `EXPECTED_REGISTERED_ROUTES` (line 119+) are two hand-maintained 35-entry arrays encoding the same route set; the literal `35` is repeated in titles/`toHaveLength(35)`.

**Why:** an independent inventory vs the app is the point of a drift test, but a second near-identical copy beside it is a maintenance trap the drift test can't catch; the `simplify` commit also deleted the comment explaining why `ROUTES` is kept independent.

**Fix:** derive one from the other (or assert their lengths match), use `EXPECTED_REGISTERED_ROUTES.length` instead of `35`, and restore the one-line rationale.

AlCalzone (commented):
## Deep review — HTTP compatibility characterization

**Verdict:** Ship with minor changes. The `createApp()` refactor is the right seam and is verified behavior-preserving for the single production instance (route/middleware registration, instance-scoped closures, `serialPorts` bind, backcompat for `bin/www.ts`, `loadSnippets` idempotency, and the route-drift walk all check out). One Major is a genuine regression worth fixing before merge; the other two are worth addressing or explicitly deferring.

**Top 3 risks**
1. **Process/signal handlers register after the startup window** (app.ts:348) — a crash or `docker stop` during the multi-second controller bring-up skips graceful shutdown. Found independently by the correctness and operability passes; contradicts the "behavior-preserving" claim. Trivial fix (move the block above the awaits).
2. **The safety net locks response shapes but not the middleware chain** (routeContract.test.ts) — only 2 of 31 `isAuthenticated`-guarded routes are tested authenticated-but-unauthorized, and drift detection ignores middleware, so the planned sub-router split could silently drop an auth guard and stay green.
3. **Test-only setters are exposed as permanent public API** (app.ts:127) with no `@internal`/TODO marker.

**Strengths**
- Route-drift detection correctly recurses mounted sub-routers and reconstructs mount prefixes (verified by its own unit tests).
- Real per-file isolation (env/store/serial-port), no `.only`/`.skip`, no leaked handles (limiter timers `unref`'d).
- `loadSnippets()` idempotency and `Gateway.watchers.clear()` are genuine latent-bug fixes, each covered.
- Backcompat clean; `typeof`-derived harness types are robust, not fragile.

**Themes**
- Middleware/lifecycle is characterized less tightly than routes/envelopes — Majors #1 and #2 are the same shape.
- The final `simplify` commit (1f5f9cf8) over-corrected, deleting several load-bearing why-comments (rate-limit `6`-bound, harness typing/`vi.hoisted`, `ROUTES` independence).
- No Blockers; no secrets leaked (`TEST_SESSION_SECRET` is a self-labeling dummy); `supertest` is correctly dev-only.

**Preserved pre-existing quirks** (untouched by this PR, noted for the record): rate-limit exhaustion returns HTTP 200 not 429; `cors({origin:true, credentials:true})` with no `sameSite` on the session cookie; limiters key on `req.ip`.

**Coverage:** all 8 specialists ran (correctness, security, DRY, design, performance, operability, tests, readability). No changed file left uncovered; `package-lock.json` skipped (generated).

test/lib/http/auth.test.ts
```diff
@@ -0,0 +1,248 @@
...
+import { seedUser, signUserToken, setSettings } from './authHelpers.ts'
+
+describe('HTTP contract: auth & password', () => {
+	let harness: HttpHarness
+
+	beforeAll(async () => {
+		harness = await createHttpHarness()
+	})
+
+	afterAll(async () => {
```
AlCalzone: **[Minor] · DRY**

The `beforeAll → createHttpHarness()` / `afterAll → close()` / `afterEach → resetState()` block is copy-pasted across all 9 contract files.

**Why:** any change to setup/teardown policy needs 9 synchronized edits.

**Fix:** a `useHttpHarness()` accessor in harness.ts that registers the hooks and returns a getter.

test/lib/http/authRateLimit.test.ts
```diff
@@ -0,0 +1,40 @@
...
+	afterAll(async () => {
+		await harness.close()
+	})
+
+	it('replies with the HTTP-200 rate-limit envelope once the login budget is exhausted', async () => {
+		let lastBody: unknown
+		let lastStatus: number | undefined
+		for (let attempt = 1; attempt <= 6; attempt++) {
+			const res = await harness.request
+				.post('/api/authenticate')
```
AlCalzone, line 14: **[Minor] · Readability**

The `<= 6`/`<= 5` bounds lost their explanatory comment in the `simplify` commit (1f5f9cf8).

**Why:** the `6` is tied to the limiter's `max: 5` (a successful login resets the counter, so exhausting it takes 6 consecutive failures) and isn't restatable from the code — exactly the repo's own "worth a why" case.

**Fix:** restore one line, e.g. `// max: 5 lets 5 failures through; the 6th trips the limiter`.

test/lib/http/configurationTemplates.test.ts
```diff
@@ -0,0 +1,297 @@
...
+				data: [{ id: 't1', name: 'Template 1' }],
+			})
+		})
+
+		it('fails with a generic error when no gateway is attached', async () => {
+			const res = await harness.request.get(
+				'/api/configuration-templates',
+			)
+
+			expect(res.status).toBe(200)
```
AlCalzone, line 33: **[Minor] · Readability**

Asserts the raw `"Cannot read properties of undefined (reading 'zwave')"` with no framing (same at importExport.test.ts:39-50), while store.test.ts labels the identical case a "preserved quirk".

**Why:** unlabeled, an unguarded-optional-chain TypeError in an assertion reads as an accidental bug rather than a deliberate characterization — the opposite of what the sibling tests signal.

**Fix:** align the title/comment with store.test.ts's "preserved quirk" pattern at both sites.

test/lib/http/harness.ts
```diff
@@ -0,0 +1,117 @@
...
+type AppInstance = ReturnType<AppModule['createApp']>
+type JsonStoreModule = typeof JsonStoreModuleNamespace
+type StoreConfigModule = typeof StoreConfigModuleNamespace
+type GatewayModule = typeof GatewayModuleNamespace
+
+type TestHooks = Omit<AppInstance, 'setGateway'> & {
+	setGateway(value: FakeGateway | undefined): void
+}
+
+const { enumerateSerialPorts } = vi.hoisted(() => ({
```
AlCalzone: **[Minor] · Readability**

The `simplify` commit deleted the why-comments for narrowing `setGateway` to `FakeGateway` (Gateway's private fields defeat structural typing) and for routing `enumerateSerialPorts` through `vi.hoisted` (vitest hoisting order).

**Why:** both are genuine non-obvious facts the surrounding code can't show; their removal wasn't a restatement cleanup.

**Fix:** restore a one-line why for each.

test/lib/http/authRateLimit.test.ts
```diff
@@ -0,0 +1,40 @@
...
+				.send({ username: 'nobody', password: 'nope' })
+			lastBody = res.body
+			lastStatus = res.status
+
+			if (attempt <= 5) {
```
AlCalzone, line 19: **[Minor] · Tests**

Only `loginLimiter` (max 5/hr) is characterized; `storeLimiter` (100/15min) and `apisLimiter` (500/hr), each with a distinct handler message, are never exercised.

**Why:** a refactor that merges, renumbers, or drops one of the other two limiters has no test to catch it.

**Fix:** assert wiring for each (e.g. an injectable lowered `max`), or add one boundary test per limiter.

test/lib/http/configurationTemplates.test.ts
```diff
@@ -0,0 +1,297 @@
...
+			expect(res.body).toEqual({
+				success: false,
+				message: 'nodeId is required',
+			})
+			expect(gw.zwave.applyConfigurationTemplate).not.toHaveBeenCalled()
+		})
+
+		it('applies the template to the node, coercing force to a strict boolean', async () => {
```
AlCalzone: **[Minor] · Tests**

The "coercing force to a strict boolean" case only omits `force` (asserting `!!undefined === false`), so it locks the default, not the coercion.

**Why:** it never sends a truthy non-boolean, so `!!force` coercing a value to `true` is left unobserved.

**Fix:** add a case with `force: 'yes'`/`1` asserting the collaborator receives literal `true`.

package.json
```diff
@@ -1,6 +1,10 @@
 {
   "name": "zwave-js-ui",
   "type": "module",
+  "imports": {
```
AlCalzone: **[Minor] · DRY + Design**

The new `#api/*` subpath import coexists with the older relative convention (`test/lib/*.test.ts` use `../../api/...`) and vitest's own `@/`/`@server/` aliases — three resolution mechanisms for the same thing.

**Why:** fragments the repo's import story; readers and tooling must know all three.

**Fix:** subpath imports are the more portable choice — fine to standardize on them, but migrate the existing files or note the intent to converge.

api/app.ts
```diff
@@ -116,2195 +123,2316 @@ const multerUpload = multer({
...
+				res.json({
+					success: true,
+					message: 'Password updated',
+					user: userData,
+				})
+			} catch (error) {
+				res.json({
+					success: false,
+					message: 'Error while updating passwords',
+					error: error.message,
```
AlCalzone, line 1213: **[Minor] · Security**

`PUT /api/password` returns raw `error.message` to the client (a `TypeError` from `user.username` when no session user exists), and auth.test.ts:170 now cements it.

**Why:** internal exception strings leak into the response body, and the test guards this as expected-forever.

**Fix:** out of scope to fix the handler here, but don't assert on the internal message text — log the detail and return a generic string in Layer 2.

package.json
```diff
@@ -1,6 +1,10 @@
...
+    "#api/*": "./api/*",
+    "#test/*": "./test/*"
```
AlCalzone: **[Nit] · DRY**

`"#test/*": "./test/*"` is declared but referenced nowhere (the harness and helpers import each other via relative `./env.ts`, `./fakes.ts`).

**Why:** dead alias config.

**Fix:** drop it, or actually use it.

test/lib/http/store.test.ts
```diff
@@ -0,0 +1,419 @@
...
+
+		it('rejects a path that escapes the store dir, without reading anything', async () => {
+			const res = await harness.request
+				.get('/api/store')
+				.query({ path: '../../etc/passwd' })
+
+			expect(res.status).toBe(200)
+			expect(res.body).toEqual({
+				success: false,
+				message: 'Path not allowed',
```
AlCalzone, line 67: **[Nit] · Readability**

The 8-line "buffer response as bytes because supertest's JSON parser would mangle it" block is copy-pasted ~5× across store.test.ts, routeContract.test.ts, and debug.test.ts.

**Why:** each occurrence buries the actual assertion under identical boilerplate; a fix needs synchronized edits.

**Fix:** extract a shared `bufferResponse(req)` helper.

test/lib/http/env.ts
```diff
@@ -0,0 +1,80 @@
...
+import { tmpdir } from 'node:os'
+import path from 'node:path'
+
+vi.mock('dotenv', () => ({
+	config: () => ({ parsed: {} }),
+}))
+
+let storeDir: string | undefined
+
+const APP_ENV_VARS = [
```
AlCalzone: **[FYI] · DRY**

`APP_ENV_VARS` hand-mirrors the ~30 `process.env.*` names the app reads.

**Why:** it will silently drift as new env vars are added, weakening the isolation guarantee.

**Fix:** a pointer comment to the source list, or a shared constant.

api/app.ts
AlCalzone: See also my previous comment. We need a better solution for this, ideally without needing these hooks at all.

api/app.ts
AlCalzone: worth cleaning up.

api/app.ts
AlCalzone: Yes. No default/throwaway instance. Create it only when needed.

test/lib/http/configurationTemplates.test.ts
AlCalzone: see above. worth fixing, not a quirk.

package.json
AlCalzone: deferred to later work

test/lib/http/env.ts
AlCalzone: can we avoid this altogether?

test/lib/http/debug.test.ts
AlCalzone: Addressed in 8b519624. The assertion now explains that the ZIP signature proves the response body is a real archive.

test/lib/http/routeContract.test.ts
AlCalzone: Addressed in 5a35fd60. The shallow request-per-route checks were removed; this file now contains only the independent registration inventory and recursive router traversal regression tests.

test/lib/http/store.test.ts
AlCalzone: Addressed in 8b519624. The setup and ZIP signature assertions now explain why persisted input and the expected bytes are required.

api/app.ts
AlCalzone: Addressed in e1b74bdc. The test-only mutators and eager singleton were removed; createApp now receives initial state and exposes only app, startServer, and loadSnippets.

test/lib/http/authRateLimit.test.ts
AlCalzone: Addressed in 338fd108. The comment now states why six attempts are required for a max of five.

test/lib/http/configurationTemplates.test.ts
AlCalzone: Addressed in 117804e3. Missing gateway state now produces clean route errors instead of leaking TypeErrors, and the tests assert those responses.

test/lib/http/authRateLimit.test.ts
AlCalzone: Addressed in 8b519624. The suite now exercises the exact 100-request store limiter and 500-request API limiter boundaries and their distinct envelopes.

api/app.ts
AlCalzone: Addressed in 8b519624. The test no longer locks the raw internal exception message; generic password error handling remains available for Layer 2.

test/lib/http/store.test.ts
AlCalzone: Addressed in 8b519624. Binary response collection is centralized in the shared bufferResponse helper.

test/lib/http/routeContract.test.ts
AlCalzone: Addressed in 5a35fd60. Only the independent route inventory and recursive traversal tests remain.

test/lib/http/store.test.ts
AlCalzone: Fixed in 117804e3. A request without multipart files now returns the clean No file uploaded envelope instead of throwing.

test/lib/http/store.test.ts
AlCalzone: Fixed in 117804e3. Missing gateway state is now guarded for snippets, imports, and configuration templates rather than producing a TypeError.

test/lib/http/health.test.ts
AlCalzone: Fixed in 117804e3. Invalid health clients return once with the corrected message instead of falling through to a second response.

test/lib/http/harness.ts
AlCalzone: Addressed in e1b74bdc and c7c240bf. The mutator hooks were removed; each test now creates a fresh app instance from injected initial state.

api/app.ts
AlCalzone: Fixed in e1b74bdc. startServer now installs its handlers before the first startup await.

api/app.ts
AlCalzone: Fixed in e1b74bdc. Repeated startup removes only this instance’s prior callbacks before registering replacements, without touching unrelated listeners.

api/app.ts
AlCalzone: Fixed in e1b74bdc. The eager default instance was removed, and bin/www.ts now creates the sole production instance when needed.

test/lib/http/auth.test.ts
AlCalzone: Addressed in c7c240bf and 8b519624. All contract suites now use the shared useHttpHarness lifecycle accessor.

test/lib/http/harness.ts
AlCalzone: Addressed in e1b74bdc and c7c240bf. The gateway narrowing disappeared with the mutator hook, and the remaining vi.hoisted requirement is documented beside the mock.

test/lib/http/configurationTemplates.test.ts
AlCalzone: Addressed in 8b519624. A truthy non-boolean force value now verifies that the collaborator receives literal true.

package.json
AlCalzone: Fixed in a997b496. The unused #test/* import alias was removed.

test/lib/http/env.ts
AlCalzone: Addressed in c7c240bf. The mirrored environment-variable list was removed; tests mock the config module before its dotenv load instead.

api/app.ts
```diff
@@ -116,2195 +123,2311 @@ const multerUpload = multer({
...
+}
 
-const loginLimiter = rateLimit({
-	windowMs: 60 * 60 * 1000, // keep in memory for 1 hour
-	max: 5, // start blocking after 5 requests
-	handler: function (req, res) {
-		res.json({ success: false, message: 'Max requests limit reached' })
-	},
-})
+export function createApp(options: CreateAppOptions = {}): AppInstance {
```
AlCalzone, line 156: I'd prefer if these test-related options would only be evaluated while testing, not in production.

AlCalzone: Fixed in 5c8625b7. Injected state now lives under a nested test option that createApp reads only when NODE_ENV is test, with regression coverage proving production does not evaluate it.

robertsLando:
@AlCalzone could we drop node 20 from ci and add node 24?

api/app.ts
```diff
@@ -116,2195 +123,2313 @@ const multerUpload = multer({
...
+	// ### UTILS
+
+	/**
+	 * Start http/https server and all the manager
+	 */
+	async function startServer(port: number | string, host?: string) {
+		let server: HttpServer
+
+		// Register before any awaits so the whole startup window is covered
+		process.removeListener('uncaughtException', handleUncaughtException)
```
robertsLando, line 278: **[Major] · Correctness + Operability**

Signal and `uncaughtException` handlers register here, inside `startServer()` — but `www.ts` only calls `startServer()` after `await jsonStore.init()`. In the base they were module-scope, live from the moment `app.ts` was imported. That opens a handler-free startup window.

**Why:** `docker/Dockerfile:100` is exec-form `CMD ["node", "server/bin/www"]` with no tini, and there's no `init: true` in the compose file — so node is **PID 1**. PID 1 gets no default signal disposition; Node only installs an OS handler once a JS listener exists. During this window SIGTERM is silently discarded: `docker stop` waits out the full 10s grace period, then SIGKILLs. Normally the window is a few ms of JSON reads, but `jsonStore._getFile()` awaits `readFile` per store file — on a stale NFS mount or a failing SD card (the Raspberry Pi userbase) the process parks there and cannot be stopped at all. Same story for systemd `TimeoutStopSec` on the packaged binary.

What caps this at Major rather than Blocker: nothing is *held* during the window. The reads are read-only, and the serial port + MQTT are opened in `startGateway()`, which runs after registration — so no stuck stick and no torn config. The cost is an unkillable or 10s-stalled shutdown.

**Fix:** install the process handlers from the composition root (`www.ts`, before `jsonStore.init()`) rather than inside `startServer()`. Keeping them out of `startServer` also avoids the ~9 test apps stacking listeners.

test/lib/http/routeContract.test.ts
```diff
@@ -0,0 +1,217 @@
...
+): T[] {
+	return [...routes].sort((a, b) =>
+		`${a.method} ${a.path}`.localeCompare(`${b.method} ${b.path}`),
+	)
+}
+
+describe('HTTP contract: complete Express route inventory (drift detection)', () => {
+	const getHarness = useHttpHarness()
+
+	it('registers exactly the independently expected 35 routes - no more, no fewer', async () => {
```
robertsLando, line 122: **[Major] · Tests + Security** (flagged independently by both passes)

The drift detector pins `{method, path}` only — it never inspects the middleware chain. And behaviorally, of the ~30 routes wrapped in `isAuthenticated`, `authEnabled: true` is set in exactly **one** file (`auth.test.ts`), and only `GET /api/logout` is exercised in the deny direction. Every other protected route is driven with auth at its default-disabled state, where `isAuthenticated` short-circuits on `!isAuthEnabled()` before the session check even runs.

(`PUT /api/password`'s "no session user" test looks like it covers this but doesn't — it also runs with auth disabled, so it asserts the handler's null-session fallback, not the guard firing.)

**Why:** the stated next step is splitting `app.ts` into sub-routers. Today a refactor that drops `isAuthenticated` from `/api/store`, `/api/settings`, `/api/importConfig` or `/api/debug/*` **passes all 262 tests** — the exact auth regression this suite exists to prevent. This is the highest-value thing a characterization suite of an authenticated API can lock, and right now it's the one thing it doesn't.

**Fix:** the route list already exists here. Make it table-driven: for each protected route, set `authEnabled: true`, send with no session/token, and assert the `{success:false, code:3}` envelope. Optionally also have `getActualRegisteredRoutes` collect `layer.route.stack.map(h => h.name)` so the expected handler chain (`apisLimiter,isAuthenticated,…`) is pinned per route — that would cover the rate limiters too, which are currently sampled on only 3 of 35 routes.

package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "zwave-js-ui",
   "type": "module",
+  "imports": {
+    "#api/*": "./api/*"
```
robertsLando: **[Major] · Design / API**

`"#api/*": "./api/*"` maps to **source**, but the shipped artifact has no `api/`: `files` is `dist, server, snippets, …`, `docker/Dockerfile` copies only `dist`/`server`/`snippets`, and `pkg.scripts` is `server/**/*`. `tsc` emits to `server/`, and `rewriteRelativeImportExtensions` only rewrites *relative* specifiers.

**Why:** nothing breaks today because `test/lib/http/*` is the sole consumer. But the manifest now advertises a subpath that dangles at runtime, and the first follow-up PR in the stack that writes `#api/lib/x.ts` inside `api/**` will emit it verbatim into `server/**` and die with `ERR_MODULE_NOT_FOUND` in Docker and pkg — **while `npm test` stays green**. That's a failure mode that skips CI and lands in a release. It's also now a third module-resolution mechanism alongside relative paths and the `@/` / `@server/` vite aliases.

**Fix:** tests are the only consumer — put the alias in `vitest.config.ts` `resolve.alias` (next to the existing `@server`) and drop the manifest field. If it must stay a subpath import, make it conditional (`{"types": "./api/*", "default": "./server/*"}`) and ship `api/` in `files`.

robertsLando (commented):
## Deep review — `test(api): characterize HTTP compatibility`

**Verdict: needs work, but close.** The approach is right and most of this is genuinely good. Two things are worth fixing before it lands, because this is the *root* of the stack and both flaws get inherited by everything built on top of it.

### Top risks

1. **`docker stop` can now be silently ignored during startup.** Signal + `uncaughtException` handlers moved from module scope into `startServer()`, which `www.ts` only calls after `await jsonStore.init()`. `docker/Dockerfile:100` is exec-form `CMD ["node", "server/bin/www"]` with no tini, and there is no `init: true` in the compose file — so **node is PID 1**, which gets no default signal disposition. During the new window there is no listener, so SIGTERM is discarded, Docker waits the full 10s grace period, then SIGKILLs. Usually that window is milliseconds, but `jsonStore` awaits `readFile` per store file: on a stale NFS mount or a dying SD card (the Pi userbase) the process parks there and cannot be stopped at all. Capped at Major because nothing is *held* during the window — the reads are read-only, and the serial port + MQTT open in `startGateway()`, after registration. See the inline comment on `api/app.ts`.

2. **The auth guard — the single highest-value thing this suite exists to lock — is not characterized.** ~30 routes are wrapped in `isAuthenticated`, but `authEnabled: true` is set in exactly one file and only `GET /api/logout` is exercised in the deny direction. A refactor that drops `isAuthenticated` from `/api/store`, `/api/settings` or `/api/debug/start` passes all 262 tests today — the exact regression class this PR was written to prevent. See the inline comment on `routeContract.test.ts`.

3. **The `#api/*` subpath import is a landmine for the next PR in the stack** — green tests, broken Docker/pkg. See the inline comment on `package.json`.

### Strengths

- **The route table is byte-identical before/after** — all 47 registrations keep the same `loginLimiter`/`apisLimiter`/`storeLimiter` + `isAuthenticated` chain. No route silently lost a guard.
- `createApp()` is the correct seam: it converts module-load side effects into an explicit construction step, which is what makes route extraction mechanical later.
- The tests drive a real `http.Server` over supertest through the real middleware chain, not mocks of the harness. All 35 routes have at least one behavioral assertion, and the streaming endpoints properly assert Content-Type, Content-Disposition and ZIP magic bytes.
- The harness is properly factored (`useHttpHarness()` in 9/9 suites, no copy-paste).
- The `NODE_ENV === 'test'` seam has no bypass: prod calls `createApp()` with no argument, so `options.test` is unreachable regardless of the env var.

### A finding that did *not* survive verification

Dropping `process.removeAllListeners('SIGINT')` looked like a Blocker — the theory being that foreign SIGINT handlers from `zwave-js`/`serialport`/`mqtt` would now survive and preempt `gracefuShutdown`. I tested it: importing all three registers **zero** SIGINT/SIGTERM/uncaughtException listeners. The removal is safe. Noting it so it doesn't get re-litigated.

### Housekeeping

- **The red Node-20 CI job is a Coveralls `504`, not a test failure** — all 262 tests passed. It just needs a re-run.
- On the ZIP-served-as-`application/json` quirk on `/api/store/backup` and the other quirks raised in the thread: agreed they should be fixed, but in a follow-up, not here. This PR's job is to *freeze* behavior; mixing fixes into it undermines that. Related: the PR body claims "keeps production routing unchanged", which is true of routing but elides real response changes (the nine new `Z-Wave client not inited` guards, and the `/health/:client` payload rewrite). Worth enumerating them in the body.

### Coverage

Specialists run: Correctness, Security, Design/API, Tests, Operability, DRY, plus a design pre-flight (verdict: PROCEED). Clean on: Performance, Readability.

api/app.ts
```diff
@@ -116,2195 +123,2313 @@ const multerUpload = multer({
 }).array('upload', 1) // Field name and max count
 
 const FileStore = sessionStore(session)
-const app = express()
-const logger = loggers.module('App')
 
-const verifyJWT = promisify(jwt.verify.bind(jwt))
+export interface AppInstance {
+	app: Express
```
robertsLando, line 127: **[Major] · Design / API**

`AppInstance` exposes `{app, startServer, loadSnippets}` but no teardown — yet `createApp()` constructs a `SocketManager`, plugins, `backupManager`, `debugManager`, and adds a `loggers.logStream.on('data')` listener that is never removed.

**Why:** the harness is the proof. `test/lib/http/harness.ts` has to reach *around* the API — importing `closeWatchers()` from `Gateway.ts` and hand-deleting `jsonStore.store` keys — because the factory doesn't own its own cleanup. A factory whose entire purpose is producing N independent instances should be able to dispose of one. As the stack grows and more state moves behind this seam, every new resource will need another bespoke reach-around in the harness.

**Fix:** add `close(): Promise<void>` to `AppInstance` (essentially `gracefuShutdown` minus `process.exit`) and have the harness call it.

api/app.ts
```diff
@@ -116,2195 +123,2313 @@ const multerUpload = multer({
...
 
-			settings.zwave.enableStatistics = enableStatistics
-			settings.zwave.disclaimerVersion = 1
+				if (debugManager.isSessionActive()) {
+					await debugManager.cancelSession()
+				}
 
-			await jsonStore.put(store.settings, settings)
+				// Close gateway and restart
+				await gw.close()
```
robertsLando, line 1592: **[Minor] · Correctness**

`/api/restart` still calls `await gw.close()` (and `backupManager.init(gw.zwave)` just below) unguarded, while ~8 sibling endpoints in this PR gained `if (!gw?.zwave) throw Error('Z-Wave client not inited')`. Same for `debugManager.startSession(gw.zwave)` on `/api/debug/start` and the socket handlers.

**Why:** the guarding is half-done, so the new invariant isn't actually enforced — and `strict: false` in tsconfig means changing `gw` to `Gateway | undefined` buys zero compiler help. An operator hitting restart before the driver came up gets `Cannot read properties of undefined (reading 'close')` instead of the clear message the other routes now produce.

**Fix:** guard all of them or none — and if none, say so, since a partially-applied invariant is worse than an absent one.

api/app.ts
```diff
@@ -116,2195 +123,2313 @@ const multerUpload = multer({
...
-	logger.error(
-		`Unhandled Rejection, reason: ${reason}${stack ? `\n${stack}` : ''}`,
-	)
-})
+	function handleUncaughtException(reason: unknown) {
+		const stack = reason instanceof Error ? reason.stack : undefined
+		const message =
+			reason instanceof Error ? reason.message : inspect(reason)
+		logger.error(
+			`Unhandled Rejection, reason: ${message}${stack ? `\n${stack}` : ''}`,
```
robertsLando: **[Minor] · Operability**

This logs `Unhandled Rejection, reason: …`, but the handler is registered for **`uncaughtException`**. Grepping `api/`, there is no `unhandledRejection` handler at all.

**Why:** a genuine synchronous throw gets reported at 3am as a rejection, sending the reader down the wrong path. (Node ≥15 does route unhandled rejections through `uncaughtException` when a hook exists, so both land here — which is exactly why the label needs to come from the actual event.) Pre-existing mislabel, but this PR now owns and rewrites the function.

**Fix:** register both events explicitly and label from the one that fired.

api/lib/serialPorts.ts
```diff
@@ -0,0 +1,4 @@
+import { Driver } from 'zwave-js'
```
robertsLando: **[Minor] · DRY / Codebase fit**

A production module that exists solely to give `vi.mock()` a seam, with no comment saying so.

**Why:** the repo's existing mockability convention is **injection**, not module mocking — `StorageHelper` takes its fs deps as a constructor arg and `test/lib/jsonStore.test.ts` injects `vi.fn()`s; there is no other `vi.mock()` module mock anywhere in `test/`. This PR *itself* already adds a DI seam (`createApp({ test: { gateway } })`). So there are now two competing testability conventions for the same problem, and one of them leaks into production code. Practical risk: the next dev inlines `Driver.enumerateSerialPorts` back and silently breaks the mock.

**Fix:** pass the enumerator through the options object like `gateway` and delete the module — or keep it and state in a header comment that this is the new convention.

test/lib/http/store.test.ts
```diff
@@ -0,0 +1,423 @@
...
+				// PK\x03\x04 is the ZIP local-file-header signature, proving the body is a real archive
+				expect(
+					(res.body as Buffer).subarray(0, 4).toString('hex'),
+				).toBe('504b0304')
+			},
+		)
+	})
+
+	describe('POST /api/store/upload', () => {
+		it('rejects with "No file uploaded" for a request with no multipart body at all', async () => {
```
robertsLando, line 261: **[Minor] · Security / Tests**

The upload/restore block covers `folder: '../../etc'` but never a ZIP containing an **escaping symlink**.

**Why:** `assertNoEscapingSymlinks` on the staged restore dir is the only store defense with zero coverage. The decomposition could silently drop the `mkdtemp` → assert → merge sequence and nothing would fail.

**Fix:** add a fixture ZIP with a `../../` symlink entry; assert the restore is rejected and that nothing lands outside `storeDir`.

AlCalzone:
> could we drop node 20 from ci and add node 24?

Makes sense - I guess we'll be targeting the v12 release with this anyways.

test/lib/http/store.test.ts
AlCalzone: Double check that this is not already handled by existing tests introduced by https://github.com/zwave-js/zwave-js-ui/pull/4678 or https://github.com/zwave-js/zwave-js-ui/pull/4685. If yes, consider merging those tests into here.

api/lib/serialPorts.ts
AlCalzone: Good point. I introduced the module to give a clean mocking seam, but using the DI hooks is more consistent.

package.json
AlCalzone: Good catch - originally I planned to have more subpath imports, but having one only for tests does not make sense.

api/app.ts
AlCalzone: handleUncaughtException needs access to the logger instance though, which is bound to the app instance. Check if we can reorder the handler attachments to earlier, so we avoid the failure mode outlined here.

AlCalzone: Fixed in 4a2d08d0. The app instance exposes process-handler installation, and www.ts invokes it before jsonStore.init() while retaining access to the instance logger.

test/lib/http/routeContract.test.ts
AlCalzone: Fixed in 0160db37. An independent hard-coded inventory now verifies the exact unauthenticated denial envelope for all 30 guarded routes.

package.json
AlCalzone: Fixed in c9450f51. The package import map is removed; #api resolves only in Vitest and the ESLint TypeScript config.

api/app.ts
AlCalzone: Fixed in 4a2d08d0 and 0160db37. AppInstance.close() now disposes instance-owned lifecycle resources idempotently, and the harness calls it during teardown while retaining separate cleanup for module-global store/watchers.

api/app.ts
AlCalzone: Fixed in 4a2d08d0. Restart and debug-start now reject absent Z-Wave state with the clean error envelope before any collaborator side effects; socket handling remains outside this HTTP-layer PR.

api/app.ts
AlCalzone: Fixed in 4a2d08d0 and covered in 0160db37. uncaughtException and unhandledRejection now have distinct handlers and event-accurate labels.

api/lib/serialPorts.ts
AlCalzone: Fixed in 4a2d08d0 and 0160db37. Serial enumeration now uses the gated createApp test DI hook, and the production-only mock module was removed.

test/lib/http/store.test.ts
AlCalzone: Already covered in the stack. PR #4678 directly tests escaping symlink detection, and follow-up PR #4727 adds the full restore-ZIP rejection and outside-store cleanup contract in commit 348bf532, so this PR does not duplicate it.

robertsLando (approved):

test/lib/http/auth.test.ts
robertsLando: yep as stated above

test/lib/http/settings.test.ts
robertsLando: it is, we can create a dedicated api for this

PR #4724, opened 2026-07-10
test(socket): characterize websocket compatibility

AlCalzone:
Locks the current Socket.IO and dynamic Z-Wave API wire behavior before backend handlers and classes are split.

- Review the real HTTP/Socket.IO harness isolation and test-only `createApp()` hooks.
- Verify the seven inbound and 24 outbound event contracts, ACK envelopes, authentication fallback, room routing, producer payloads, and timing.
- Confirm the complete `allowedApis`/`callApi()` compatibility surface, including underscore-prefixed APIs and preserved wire quirks, without consumer changes.

Stacked on #4723. Part of #4722.


test/lib/socket/callApi.test.ts
```diff
@@ -0,0 +1,279 @@
...
+				'_sceneGetValues',
+				'_addSceneValue',
+				'_removeSceneValue',
+				'_activateScene',
+				'_createGroup',
+				'_updateGroup',
+				'_deleteGroup',
+				'_getGroups',
+			])
+		})
```
AlCalzone: We should avoid having hardcoded lists like these in tests.

test/lib/socket/contracts.test.ts
```diff
@@ -0,0 +1,54 @@
...
+				expect(seenBy.has(evt)).toBe(false)
+				seenBy.set(evt, channel)
+			}
+		}
+	})
+
+	it("ALL_CHANNELS is exactly channelMap's own keys, in declaration order", () => {
+		expect(ALL_CHANNELS).toStrictEqual(Object.keys(channelMap))
+	})
+
```
AlCalzone: I assume the definition of `ALL_CHANNELS` is literally `Object.keys(channelMap)` - this is a useless test.

test/lib/socket/contracts.test.ts
```diff
@@ -0,0 +1,54 @@
...
+	it('eventToChannel is exactly the reverse index of channelMap', () => {
+		const expectedReverse: Record<string, string> = {}
+		for (const [channel, events] of Object.entries(channelMap)) {
+			for (const evt of events) {
+				expectedReverse[evt] = channel
+			}
+		}
+		expect(eventToChannel).toStrictEqual(expectedReverse)
+	})
```
AlCalzone: Likewise. Don't test definitions - test behavior!

test/lib/socket/contracts.test.ts
AlCalzone, line 1: Actually, lets drop this entire file.

test/lib/socket/env.ts
AlCalzone, line 1: ...should we maybe have one shared `test/lib/shared/env.ts` file or similar then?

test/lib/socket/fakes.ts
AlCalzone, line 1: Is this just redefining some of the fakes the http modules use? Why not have one shared file then?

test/lib/socket/harness.ts
AlCalzone, line 1: Lets investigate if we can use a shared harness for HTTP and socket tests.

test/lib/socket/inboundApis.test.ts
```diff
@@ -0,0 +1,399 @@
...
+				args: ['Party mode', 42],
+			})
+			expect(gateway.zwave.callApi).toHaveBeenCalledWith(
+				'_createScene',
+				'Party mode',
+				42,
+			)
+		})
+
+		it('routes through the REAL ZwaveClient.callApi() dispatcher (not a mocked gw.zwave.callApi) for a real allowed method, echoing its real success/result/args (ZwaveClient.ts:6032-6070, app.ts:708-726)', async () => {
```
AlCalzone: NEVER refer to code lines in a test case name. Those are going to change faster than we can update the tests.

test/lib/socket/inboundApis.test.ts
```diff
@@ -0,0 +1,399 @@
...
+			const result = await emit(client, 'MQTT_API', {
+				api: 'notARealAction',
+				args: [],
+			})
+			expect(result).toStrictEqual({
+				success: false,
+				message: 'Unknown MQTT api undefined',
+				api: 'notARealAction',
+			})
+		})
```
AlCalzone: Yeah that is a bug, not a quirk. Fix it.

test/lib/socket/subscriptions.test.ts
```diff
@@ -0,0 +1,160 @@
...
+	return client
+}
+
+// Every socket auto-joins a room named after its own id as a Socket.IO built-in, unrelated to SUBSCRIBE/channelMap
+// Round-trips a marker event through client's own private auto-joined room to deterministically flush anything already in flight
+function barrier(harness: SocketHarness, client: any): Promise<void> {
+	const arrived = waitForEvent(client, '__TEST_BARRIER__')
+	harness.io.to(client.id).emit('__TEST_BARRIER__')
+	return arrived.then(() => undefined)
+}
```
AlCalzone: Is this documented behavior we can rely on? Seems fragile if not.

test/lib/shared/authHelpers.ts
AlCalzone, line 1: Looks like we're duplicating helpers from http tests again. Lets look into unifying both implementations

test/lib/socket/outboundProducers.test.ts
```diff
@@ -0,0 +1,667 @@
...
+	}
+
+	// Like waitForEvent, but captures every argument for events that pass more than one, e.g. NODE_UPDATED's (node, isPartial)
+	function waitForArgs(client: any, event: string): Promise<unknown[]> {
+		return new Promise((resolve) =>
+			client.once(event, (...args: unknown[]) => resolve(args)),
+		)
+	}
+
+	describe('sendToSocket mechanics (shared by ~21 of the 24 events)', () => {
```
AlCalzone: Do not refer to function numbers etc., not in comments, not in test names. These are just going to be outdated quickly.

test/lib/socket/outboundProducers.test.ts
```diff
@@ -0,0 +1,667 @@
...
+			const zwave = realZwave(harness)
+			;(zwave as any)._driver = {
+				controller: { nodes: { get: () => undefined } },
+			}
+			const client = await connectedSubscriber(harness, 'values')
+			const box: unknown[] = []
+			client.on('VALUE_UPDATED', (data: unknown) => box.push(data))
+
+			zwave.emitValueChanged(
+				{ nodeId: 2, commandClass: 37, property: 'x' } as any,
```
AlCalzone: Refer to the actual `CommandClasses` enum instead of using hardcoded CC IDs.

test/lib/socket/outboundProducers.test.ts
```diff
@@ -0,0 +1,667 @@
...
+
+		it('_updateControllerStatus() is a no-op (no emit) when the status has not actually changed', async () => {
+			const harness = await getHarness({ gateway: benignGateway() })
+			const zwave = realZwave(harness)
+			const client = await connectedSubscriber(harness, 'controller')
+			const box: unknown[] = []
+			client.on('CONTROLLER_CMD', (data: unknown) => box.push(data))
+
+			const first = waitForEvent(client, 'CONTROLLER_CMD')
+			;(zwave as any)._updateControllerStatus('Idle')
```
AlCalzone: This `;(zwave as any)` is a really annoying pattern. Can't we do better?

test/lib/socket/outboundProducers.test.ts
```diff
@@ -0,0 +1,667 @@
...
+		it('OTW_FIRMWARE_UPDATE: {result:{success,status}} via the real _onOTWFirmwareUpdateFinished() (ZwaveClient.ts:6446-6459)', async () => {
+			const harness = await getHarness({ gateway: benignGateway() })
+			const zwave = realZwave(harness)
+			const client = await connectedSubscriber(harness, 'firmware')
+			const received = waitForEvent(client, 'OTW_FIRMWARE_UPDATE')
+
+			;(zwave as any)._onOTWFirmwareUpdateFinished({
+				success: true,
+				// 255 is zwave-js's real OTWFirmwareUpdateStatus.OK member, run through getEnumMemberName() so the wire payload asserts the string name
+				status: 255,
```
AlCalzone: Use the actual enums, not their numeric counterparts.

test/lib/socket/outboundProducers.test.ts
```diff
@@ -0,0 +1,667 @@
...
+				result: { success: true },
+			})
+
+			expect(await received).toEqual({
+				node: { id: 3 },
+				result: { success: true },
+			})
+		})
+
+		it('REBUILD_ROUTES_PROGRESS: array-of-tuples shape (ZwaveClient.ts:5201/5209/6795/5936) - DOCUMENTED BOUNDARY, needs a real zwave-js rebuild-routes controller call', async () => {
```
AlCalzone: Whats with the `DOCUMENTED BOUNDARY` stuff. This belongs into comments, not in the test name.

test/lib/socket/outboundProducers.test.ts
```diff
@@ -0,0 +1,667 @@
...
+			// Unlike ZwaveClient.sendToSocket, which defers via process.nextTick, this call site routes immediately and synchronously from within onStateChange() itself
+			expect(toSpy).toHaveBeenCalledWith('znifferState')
+			toSpy.mockRestore()
+
+			expect(await received).toEqual(zniffer.status())
+		})
+
+		describe('ZNIFFER_FRAME: real producer, driven through the real internal Zniffer instance (ZnifferManager.ts:106-125)', () => {
+			// Zniffer's constructor only validates port's type and does no real serial I/O until .init() is called, so only Zniffer.prototype.init is stubbed while ZnifferManager's real 'frame'/'corrupted frame' listeners stay live for this test to drive with a real .emit() call
+			let initSpy: ReturnType<typeof vi.spyOn>
```
AlCalzone: Isn't there an actual type name we can use instead of `ReturnType<typeof vi.spyOn>`?

test/lib/socket/callApi.test.ts
AlCalzone: Fixed in bbe7ecd6 — removed the allowlist definition/structure tests and kept only observable callApi behavior.

test/lib/socket/contracts.test.ts
AlCalzone: Fixed in f14b935a — deleted contracts.test.ts instead of testing ALL_CHANNELS against its own definition.

test/lib/socket/contracts.test.ts
AlCalzone: Fixed in f14b935a — removed the reverse-index definition test with the rest of contracts.test.ts.

test/lib/socket/contracts.test.ts
AlCalzone: Fixed in f14b935a — deleted the entire definition-consistency test file.

test/lib/socket/env.ts
AlCalzone: Fixed in c7b25d7e — env isolation now lives in test/lib/shared/env.ts and is used by both transports.

test/lib/socket/fakes.ts
AlCalzone: Fixed in d747ad2b — common fakes moved to test/lib/shared/fakes.ts; socket-only extensions remain transport-specific.

test/lib/socket/harness.ts
AlCalzone: Fixed in 0153a584 — HTTP and Socket.IO harnesses are now thin adapters over the shared app/store lifecycle.

test/lib/socket/inboundApis.test.ts
AlCalzone: Fixed in 5c0f7efa — removed source file and line coordinates from test names and comments.

test/lib/socket/inboundApis.test.ts
AlCalzone: Fixed in 916cd58d — unknown MQTT actions now report data.api, and the wire test asserts the requested action name.

test/lib/socket/subscriptions.test.ts
AlCalzone: Confirmed as documented Socket.IO behavior: sockets automatically join their ID room, and message ordering is guaranteed. The consolidated barrier in 27237d8e preserves that mechanism: https://socket.io/docs/v4/rooms/ and https://socket.io/docs/v4/faq/#does-socket-io-guarantee-message-order

test/lib/shared/authHelpers.ts
AlCalzone: Fixed in 75f279d6 — auth helpers now live in test/lib/shared/authHelpers.ts with a shared structural storage type.

test/lib/socket/outboundProducers.test.ts
AlCalzone: Fixed in 5c0f7efa — removed producer/function counts and source coordinates from names and comments.

test/lib/socket/outboundProducers.test.ts
AlCalzone: Fixed in 0d26dc57 — value fixtures now use the exported CommandClasses enum.

test/lib/socket/outboundProducers.test.ts
AlCalzone: Fixed in 7d2381f9 — removed the internals adapter and private-method-only tests; retained producer coverage now uses public triggers with localized typed fixture setup.

test/lib/socket/outboundProducers.test.ts
AlCalzone: Fixed in 9f1ffd50 and 0d26dc57 — fixtures now use the exported firmware-status, data-rate, and command-class enums.

test/lib/socket/outboundProducers.test.ts
AlCalzone: Fixed in 25988b2d — removed DOCUMENTED BOUNDARY from test names and kept only concise setup rationale where needed.

test/lib/socket/outboundProducers.test.ts
AlCalzone: Fixed in acc90f24 — Zniffer spies now use Vitest MockInstance with the real method signatures.

robertsLando (commented):
## deep-review — PR #4724 `test(socket): characterize websocket compatibility`

**Verdict:** Ship with minor changes. The strategy is textbook-sound and the execution quality is high. Reviewed at `776dbd5` (byte-identical tree to the earlier `d176785`; the branch was rebased with no content change). All 17 prior self-review threads are already resolved in this head.

**Top 3 things worth a look:**
1. A real production bugfix (`data.apiName` → `data.api`, #4740) is folded into a PR whose stated job is to *lock current behavior*, and its characterization test asserts the **post-fix** string — see the inline note on `api/app.ts`.
2. The PR body claims *24 outbound event contracts* are characterized, but 8 of the 24 have no real-producer coverage (`CONTROLLER_CMD`, `CONNECTED`, `NODE_FOUND`, `NODE_ADDED`, `VALUE_REMOVED`, `METADATA_UPDATED`, `NODE_EVENT`, `OTW_FIRMWARE_UPDATE`).
3. The production test-seam in `createApp()` is a reasonable-but-debatable design call worth a conscious decision before the split lands on it.

**Strengths:**
- Correct strategy, correctly executed: real HTTP/Socket.IO server + injected fakes, reusing the *same* private `setupSocket()`/`setupInterceptor()` production calls rather than forking a parallel mock wiring — the only way the snapshot stays faithful through the refactor.
- The shared-harness extraction (`test/lib/shared/harness.ts`) is genuinely complete DRY (verified across fakes, authHelpers, env, and lifecycle), not partial.
- Tests assert real payloads and honestly document warts rather than tautologies or aspirational behavior. No brittle line-number test names, no `as any` seams, no hardcoded CC IDs — the real enums are imported.

**Themes:** (1) *Characterization purity* — the PR mixes a real bugfix into a behavior-lock PR and labels two genuine bugs as `quirk:`. (2) *Claimed-vs-actual coverage* — the "24 outbound contracts" claim overstates what is actually pinned.

**Coverage:** correctness, design/API, DRY, and tests specialists + prior-thread reconciliation. No standalone security/perf/operability pass — the only production change is a `NODE_ENV`-gated 24-line test seam (verified it cannot fire in prod), so there is no meaningful new attack or hot-path surface.

api/app.ts
```diff
@@ -805,7 +813,7 @@ export function createApp(options: CreateAppOptions = {}): AppInstance {
 							res = gw.removeNodeRetained(data.args[0])
 							break
 						default:
-							err = `Unknown MQTT api ${data.apiName}`
+							err = `Unknown MQTT api ${data.api}`
```
robertsLando, line 812: **[Major] · Correctness / Scope**

This `data.apiName` → `data.api` change is a *correct* production bugfix (the MQTT handler reads `data.api` at 803/808/827; `data.apiName` never existed, so the default branch always emitted `Unknown MQTT api undefined`). The concern is scope: it's a production behavior change inside a PR whose stated job is to lock current behavior, and the paired regression test asserts the **post-fix** string.

**Why:** it defeats the behavior-lock guarantee the refactor stack depends on (the "before" snapshot is no longer faithful), and a prod change rides in under lighter test-PR scrutiny.

**Fix:** split the one-line fix + its regression test into a dedicated #4740 bugfix PR, or drop the "pure characterization" framing from the PR description so the intent is explicit.

test/lib/socket/outboundProducers.test.ts
```diff
@@ -0,0 +1,527 @@
...
+import type ZWaveClientType from '#api/lib/ZwaveClient.ts'
+import type { ZwaveConfig, ZUINode, ZUIValueId } from '#api/lib/ZwaveClient.ts'
+import type ZnifferManagerType from '#api/lib/ZnifferManager.ts'
+import { eventToChannel } from '#api/lib/SocketEvents.ts'
+import { buffer2hex } from '#api/lib/utils.ts'
+import { useSocketHarness, type SocketHarness } from './harness.ts'
+import { createFakeGateway } from './fakes.ts'
+import { connectedSubscriber, waitForEvent, waitForArgs } from './helpers.ts'
+
+describe('Socket contract: outbound producers', () => {
```
robertsLando, line 34: **[Major] · Tests / Coverage**

The PR body claims *24 outbound event contracts* are characterized, but 8 of the 24 `socketEvents` have zero real-producer coverage: `CONTROLLER_CMD`, `CONNECTED`, `NODE_FOUND`, `NODE_ADDED`, `VALUE_REMOVED`, `METADATA_UPDATED`, `NODE_EVENT`, `OTW_FIRMWARE_UPDATE`. Their appearances in `subscriptions.test.ts` are synthetic `io.to(channel).emit(...)` room-routing payloads, not the real producer call sites — they don't pin the real event/payload contract.

**Why:** a third of the claimed surface is unlocked; the refactor could silently break `NODE_ADDED`/`NODE_EVENT`/`METADATA_UPDATED` emission with no test failing.

**Fix:** add real-producer tests mirroring the `NODE_UPDATED`/`VALUE_UPDATED` pattern, or correct the coverage claim in the PR body.

api/app.ts
```diff
@@ -2527,6 +2535,14 @@ export function createApp(options: CreateAppOptions = {}): AppInstance {
 		logFatalError(formatFatalErrorLog('unhandledRejection', reason))
 	}
 
+	// Test-only: wires the real SocketManager (+ auth middleware + every inbound handler), and optionally the log interceptor, over a caller-supplied ephemeral server - the construction-time equivalent of what startServer() does in production
+	if (testOptions?.server) {
```
robertsLando: **[Major — debatable, your call] · Design**

The test seam grows `CreateAppOptions.test` with 3 infrastructure fields (`socketManager`, `server`, `interceptor`) plus this construction-time `setupSocket()`/`setupInterceptor()` side effect. A constructor that conditionally binds a socket server leaks orchestration concern into `createApp()`, and `test.server` is a vague name for "attach sockets here."

**Fix (one option):** a narrow public `attachSocket(server)` on `AppInstance` — called by both `startServer()` and the harness — plus an `io` getter would delete all 3 infra fields and this branch, leaving only fakes under `test`.

**Counter-argument (why it may be fine as-is):** it's safe (gated on `NODE_ENV==='test'`, verified it can't fire in prod), it deliberately avoids forking production wiring (the whole point of characterization), and the upcoming split restructures `createApp()` anyway. Flagging for a conscious decision, not necessarily a change.

test/lib/socket/auth.test.ts
```diff
@@ -0,0 +1,86 @@
...
+import { describe, it, expect, afterEach } from 'vitest'
+import { useSocketHarness } from './harness.ts'
+import { createFakeGateway } from './fakes.ts'
+import { seedUser, setSettings, signUserToken } from '../shared/authHelpers.ts'
+import { emit } from './helpers.ts'
+
+describe('Socket contract: auth middleware', () => {
+	const getHarness = useSocketHarness()
+
+	afterEach(async () => {
```
robertsLando, line 12: **[Minor] · Correctness**

This suite `afterEach` calls `await getHarness()`, relying on Vitest's default `sequence.hooks: 'stack'` (later-registered hook runs first, so `useHarnessLifecycle`'s teardown runs *after* this one). Under `'list'` ordering the harness teardown runs first and sets `current = undefined`, so this hook then spins up a **fresh, never-closed** app + server + io instance per test.

**Why:** a leaked listening server + open io per test, only latent because the ordering happens to hold today.

**Fix:** reset settings via `shared.jsonStore` directly instead of re-acquiring the harness during teardown.

test/lib/socket/harness.ts
```diff
@@ -0,0 +1,159 @@
...
+			pluginsRouter: options.pluginsRouter,
+			restarting: options.restarting,
+			socketManager,
+			server,
+			interceptor: options.interceptor,
+		},
+	})
+	await instance.loadSnippets()
+
+	server.on('request', instance.app)
```
robertsLando: **[Minor] · Design**

Attach order is reversed vs production: here socket.io binds at construction (before `server.on('request', app)` and `listen`), whereas production does `createHttpServer(app)` first, `setupSocket()` second. It only works because the harness forces `transports: ['websocket']` — a polling handshake would route differently.

**Why:** this suite exists to lock wire behavior, but the ordering doesn't mirror production request routing.

**Fix:** attach the app before sockets, or document the websocket-transport dependency as an explicit constraint of the locked contract.

test/lib/socket/inboundApis.test.ts
```diff
@@ -0,0 +1,414 @@
...
+			expect(result).toStrictEqual({
+				success: true,
+				message: 'Success HASS api call',
+				api: 'rediscoverNode',
+			})
+			expect('result' in result).toBe(false)
+		})
+
+		// See #4740 for the tracked follow-up on this asymmetry
+		it('quirk: an unknown apiName silently "succeeds" (switch has no default case, res/err stay undefined)', async () => {
```
robertsLando: **[Minor] · Tests**

Three real production bugs are uniformly prefixed `quirk:` — this one (HASS_API switch has no default → silent false-success), `loadCaptureFromBuffer` called without `await` (line 306), and UNSUBSCRIBE not special-casing `"all"` (line 401). Two are unambiguous bugs; one is an accepted asymmetry. All three link to #4740, so none are silently blessed.

**Why:** the uniform "quirk" label undersells which cases need a real production fix vs. which are accepted behavior.

**Fix:** label the two genuine bugs `BUG, locked pending #4740` to separate them from the UNSUBSCRIBE asymmetry.

test/lib/socket/harness.ts
```diff
@@ -0,0 +1,159 @@
...
+	await new Promise<void>((resolve) => {
+		server.listen(0, '127.0.0.1', () => resolve())
```
robertsLando: **[Nit] · DRY**

The `server.listen(0, '127.0.0.1', …)` promise wrapper here and the verbatim `// Gives each test a fresh createApp() instance + server…` comment are duplicated with `test/lib/http/harness.ts`. Optional: a tiny `listen(server)` helper in `shared/harness.ts` could absorb the 3-line wrapper. Not load-bearing.

test/lib/socket/outboundProducers.test.ts
```diff
@@ -0,0 +1,527 @@
...
+
+			const client = await connectedSubscriber(harness, 'nodes')
+			const grantReceived = waitForEvent(client, 'GRANT_SECURITY_CLASSES')
+			const dskReceived = waitForEvent(client, 'VALIDATE_DSK')
+			const abortReceived = waitForEvent(client, 'INCLUSION_ABORTED')
+
+			// setUserCallbacks() is the public hookup point a real driver uses to reach these handlers.
+			// Invoking the same bound functions it hands off drives the identical code path.
+			zwave.setUserCallbacks()
+			expect(inclusionUserCallbacks).toBeDefined()
```
robertsLando: **[Nit] · Tests**

Lone `toBeDefined()` truthy guard amid otherwise concrete `toStrictEqual`/`toEqual` assertions. Fine as a guard before invoking its methods; noting only for consistency.

api/app.ts
AlCalzone: Sounds good. Lets do this.

test/lib/socket/auth.test.ts
AlCalzone: Can we instead make sequence.hook = stack explicit with a comment why it's needed?

test/lib/socket/harness.ts
AlCalzone: I like DRY

test/lib/socket/auth.test.ts
robertsLando: Yeah, that's the cleaner fix — go with it.

`sequence.hooks` isn't set anywhere today, so everything already runs on Vitest's default `'stack'` (LIFO). Making it explicit is behaviorally a no-op; it just pins the invariant the teardown relies on. Under `stack` the suite `afterEach` runs *before* `useHarnessLifecycle`'s teardown, so `getHarness()` returns the still-live instance and it gets closed normally — the leak only appears under `'list'` (or a per-file `sequence.hooks` override), where teardown nulls `current` first and the re-acquire spins up a fresh, never-closed server+io.

One reason to prefer the config pin over my per-suite suggestion: the pattern isn't socket-only — `test/lib/http/auth.test.ts` has the identical `afterEach → getHarness() → setSettings(...)` shape, so it's the same latent dependency there. Pinning `sequence.hooks: 'stack'` once in `vitest.config.ts` documents it for every suite in one place, rather than rewriting each hook.

Suggested comment for the config, so the "why" is obvious:

```ts
// Suite-level afterEach hooks re-acquire the harness via getHarness() to reset
// per-test settings and MUST run before useHarnessLifecycle's own teardown,
// which disposes the createApp() instance. Vitest's default LIFO 'stack' order
// guarantees this; pin it so a future default change / per-file override can't
// silently leak a listening server + io per test.
sequence: { hooks: 'stack' },
```

api/app.ts
AlCalzone: Fixed in 828ba550 — `AppInstance` now owns typed `attachSocket(server)` and `io`; the socket manager/server/interceptor test options are removed. Attachment is one-shot and rejected after close.

test/lib/socket/harness.ts
AlCalzone: Fixed in 828ba550 — the harness now creates the HTTP server with Express first, then calls the same `attachSocket()` path used by production before listening.

test/lib/socket/harness.ts
AlCalzone: Fixed in 828ba550 — both transport adapters now reuse `listenOnEphemeralPort()` from the shared harness lifecycle.

test/lib/socket/inboundApis.test.ts
AlCalzone: Fixed in 6fc3eb76 — the HASS false-success and unawaited capture cases are labeled as bugs pending #4740; the unsubscribe case remains the compatibility asymmetry.

test/lib/socket/auth.test.ts
AlCalzone: Fixed in 6fc3eb76 — Vitest now explicitly uses `sequence.hooks: stack`, with the lifecycle reason documented beside the setting.

test/lib/socket/outboundProducers.test.ts
AlCalzone: Fixed in 6fc3eb76 — the callback handoff now asserts the complete three-function shape before invoking it.

robertsLando (approved):

PR #4725, opened 2026-07-10
test(hass): characterize home assistant integrations

AlCalzone:
Locks both Home Assistant integrations before they are moved.

- Covers legacy MQTT discovery, custom-device loading, persistence, and `@zwave-js/server` lifecycle behavior.
- Preserves existing topic, payload, cache, and startup-order quirks.
- Review focus: whether fixtures describe production contracts rather than fake or helper behavior.

Part of #4722. Stacked on #4724.


api/lib/Gateway.ts
```diff
@@ -88,47 +88,72 @@ const watch = (filename: string, fn: () => void) => {
...
 	try {
-		if (fs.existsSync(customDevicesJsPath)) {
-			loaded = customDevicesJsPath
-			devices = require(CUSTOM_DEVICES)
-		} else if (fs.existsSync(customDevicesJsonPath)) {
-			loaded = customDevicesJsonPath
-			devices = JSON.parse(fs.readFileSync(loaded).toString())
+		if (fs.existsSync(jsPath)) {
+			loaded = jsPath
+			devices = require(basePath)
```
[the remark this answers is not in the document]
AlCalzone: Does this work even in esm?

test/lib/socket/inboundApis.test.ts
```diff
@@ -396,19 +486,5 @@ describe('Socket contract: inbound ACK APIs', () => {
 			)
 			expect(result).toStrictEqual({ channels: ['values'] })
```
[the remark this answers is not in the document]
AlCalzone: @robertsLando correct me if I'm wrong but this smells like a bug we should fix.

test/lib/socket/inboundApis.test.ts
```diff
@@ -225,38 +226,157 @@ describe('Socket contract: inbound ACK APIs', () => {
...
 		})
 
-		// See #4740 for the tracked follow-up on this asymmetry
-		it('BUG (#4740): an unknown apiName silently succeeds', async () => {
-			const harness = await getHarness({ gateway: createFakeGateway() })
+		it('reports success:false with the thrown error message when the action throws', async () => {
+			const gateway = createFakeGateway({
+				disableDiscovery: vi.fn(() => {
+					throw new Error('hass boom')
+				}),
```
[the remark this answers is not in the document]
AlCalzone: nah, we should fix this.

test/lib/socket/inboundApis.test.ts
```diff
@@ -301,36 +421,6 @@ describe('Socket contract: inbound ACK APIs', () => {
...
-				Buffer.from([1, 2, 3]),
-			)
-			// Proves the missing await in production: result serializes to an empty object over the wire rather than the resolved value
-			expect(result.success).toBe(true)
-			expect(result.result).toStrictEqual({})
-			expect(result.result).not.toBe('parsed-capture')
-		})
 	})
 
 	describe('SUBSCRIBE', () => {
```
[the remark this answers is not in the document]
AlCalzone: also, this is a bug that should be fixed. We should not be locking in such behavior.

test/lib/hass/discoverySensors.test.ts
```diff
@@ -0,0 +1,389 @@
...
+				property: 'state',
+				propertyName: 'state',
+				propertyKey,
+			}),
+		)
+		expect(device.type).toBe('binary_sensor')
+		// 'event_' + AlarmSensorType[key], lowercased in the common tail
+		const expectedObjectId = (
+			'event_' + AlarmSensorType[propertyKey]
+		).toLocaleLowerCase()
```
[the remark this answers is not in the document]
AlCalzone: yes, `...Locale...` does not make sense here.

test/lib/hass/gatewayHarness.ts
```diff
@@ -0,0 +1,164 @@
...
+	mqttConfig?: Partial<MqttConfig>
+	zwave?: Partial<FakeGatewayZwave>
+}
+
+/**
+ * Builds a real `Gateway` + real `MqttClient` (backed by the mocked `mqtt`
+ * broker) with defaults modeling a production HASS deployment: `NAMED` type,
+ * `hassDiscovery: true`, prefix `homeassistant`.
+ */
+export async function createGatewayHarness(
```
AlCalzone, line 60: **[Major] · Design / convention**

The established backend-test convention is a single top-of-file accessor — `const getHarness = useHttpHarness()`, backed by `useHarnessLifecycle` in `shared/harness.ts` — that registers `afterEach(close)`/`afterAll(cleanup)` internally, so a test file *can't* forget teardown. This suite instead exports a bare `createGatewayHarness()` + a separate `cleanupGatewayHarnessEnv()`, and 5 files re-declare the identical `let harness; beforeEach(create); afterEach(close); afterAll(cleanup)` trio: `discoveryClimate.test.ts:42`, `discoveryScenesConfig.test.ts:33`, `discoverySensors.test.ts:34`, `discoverySwitches.test.ts:35`, `mqttLifecycle.test.ts:42`.

**Why:** this is exactly the duplication `useHarnessLifecycle` exists to remove. A future file that copies the pattern can silently drop `afterAll` (leaking the `mkdtemp` STORE_DIR, never `rmSync`'d) or `afterEach` (leaking the MqttClient/broker, Gateway `cancelJobs` timers, and `fs.watch` watchers). The real-Gateway-per-test requirement doesn't justify it — construction is orthogonal to whether create/teardown is wrapped in an accessor.

**Fix:** a `useGatewayHarness()` here mirroring the sibling accessors. One wrinkle: `replaceHarness` (`discoveryScenesConfig.test.ts:47`, `mqttLifecycle.test.ts:65`) closes and rebuilds mid-test, so it can't reuse `useHarnessLifecycle`'s `current ??=` memoization verbatim — the new accessor must track and reassign `current`.

AlCalzone (commented):
## Deep review — test(hass) characterization suite

**Verdict:** Ship with minor changes. No blockers. The approach — locking current HASS behavior before the #4722 refactor — is sound, and both production changes (`app.ts` password guard, `Gateway.ts` `loadCustomDevicesCatalog` extraction) are correct and behavior-preserving.

**Top 3 risks:**
1. Lifecycle wiring is copy-pasted across 5 test files instead of a shared `useGatewayHarness()` accessor — a teardown-leak footgun and the main convention gap (thread on `gatewayHarness.ts`).
2. A cluster of assertions lock a *fixture default* or use a weak oracle rather than real production output, slightly diluting the characterization guarantee the PR exists to provide (threads on `mqttLifecycle`, `server`, `discoverySwitches`).
3. Minor teardown inconsistencies: leak-on-throw in `cleanup`, unbounded `mqttBrokers` growth, and two names for one file cleanup.

**Strengths:** real `Gateway` + real `MqttClient` with only the `mqtt` package faked, so captured topic/payload/qos/retain is genuine production output at the `broker.publish` seam — nothing synthesized by the harness. `discoverySwitches` locks the full wire payload and re-derives the status topic independently; `persistence` asserts on-disk bytes + home-id scoping + sync-projection-before-`nextTick` ordering; `cleanup` proves close ordering via `invocationCallOrder`. `env.ts`'s extension and the `EventEmitter`-backed `createFakeGatewayZwave` are both justified divergences from the shared fakes.

**Theme:** nearly every finding traces to one root — the suite re-implements lifecycle/teardown ad hoc rather than consuming the `useHarnessLifecycle` pattern `http`+`socket` share. A single `useGatewayHarness()` accessor resolves the Major and several Minors together.

### Nits / FYI (not filed as threads)
- **Nit** `customDevices.test.ts:141` — `Object.keys(catalog).length > 1` weak oracle; the `custom-default` check on the next line carries the test.
- **Nit** `mqttLifecycle.test.ts:205` — `expect(d2.discovery_payload).toBeDefined()` is always true.
- **Nit** `auth.test.ts:170` — asserts `.success`/`.message` individually; sibling password tests use full `toEqual({...})`.
- **Nit** `inboundApis.test.ts` delete test — `expect('result' in result).toBe(false)` is redundant beside the preceding `toStrictEqual`.
- **FYI** `mqttMock.ts:85,213` — `forceDisconnected()`'s comment ("shared broker constructed once per file") is inaccurate: each `beforeEach` mints a fresh broker, so the call is redundant. Relatedly, `mqttBrokers` grows unbounded per file since only `cleanup.test.ts` calls `resetMqttBrokers()` (`latestBroker()` stays correct — no functional bug).
- **FYI** `server.test.ts` / `tick()` — async flow advanced by a fixed count of `setImmediate` flushes couples to the fake's scheduling rather than an awaited signal.
- **FYI** `Gateway.ts` — negligible: the catalog is built and counted *before* the sha dedup, so an unchanged watch-reload does one throwaway `Object.assign`. Not a behavior change.
- **FYI** hass files use verbose JSDoc blocks on file/function/fields; `shared`/`http`/`socket` use terse inline `//` — a family consistency gap.

**Coverage:** conventions/DRY, test-quality (production-contract-vs-fake), and source + modified-test correctness. Verified clean: the `/api/password` guard (no bypass or user-enumeration, only reachable in auth-disabled mode), the `Gateway.ts` refactor (behavior-preserving, verified point-by-point), and all 7 new `inboundApis` HASS ACK assertions (match the real envelope). The 5 Copilot findings and the author's own quirk-vs-bug threads were intentionally excluded.

test/lib/hass/gatewayHarness.ts
```diff
@@ -0,0 +1,164 @@
...
+		broker,
+		publishedDiscoveries,
+		lastDiscovery() {
+			const all = publishedDiscoveries()
+			if (all.length === 0) {
+				throw new Error('No discovery packet published yet')
+			}
+			return all[all.length - 1]
+		},
+		resetPublishes() {
```
AlCalzone, line 120: **[Minor] · API**

`resetPublishes()` and `resetState()` are byte-identical (`broker.published.length = 0`), yet `resetState`'s docstring claims it clears "the gateway's de-dup maps" — which it never touches. Tests call both names interchangeably.

**Why:** implies a semantic distinction that doesn't exist; the doc will mislead a maintainer into thinking `resetState` does more.

**Fix:** keep one, or make `resetState` actually reset what its name and doc promise.

test/lib/hass/gatewayHarness.ts
```diff
@@ -0,0 +1,164 @@
...
+		},
+	}
+}
+
+/**
+ * Removes the shared `STORE_DIR` and restores the app env vars. Call once per
+ * file (final `afterAll`) after every harness has closed, so a multi-harness
+ * file doesn't delete the directory `api/config/app.ts` captured.
+ */
+export function cleanupGatewayHarnessEnv(): void {
```
AlCalzone: **[Minor] · Convention**

Two public names for one file teardown: `customDevices`/`persistence`/`server` call `cleanupTestEnv()` directly, while the harness-based files call this `cleanupGatewayHarnessEnv()` — a one-line pass-through to it. Separately, the 5 discovery `afterAll`s omit the explicit `closeWatchers()` file-end safety net that `cleanup.test.ts:37` and `customDevices.test.ts:56` *do* call.

**Why:** obscures which cleanup is canonical, and the watcher safety net is applied inconsistently across the suite.

**Fix:** collapse to one named teardown and apply `closeWatchers()` uniformly — a shared `useGatewayHarness()` accessor (see the Major) would do both.

test/lib/hass/cleanup.test.ts
```diff
@@ -0,0 +1,88 @@
...
+})
+
+afterEach(() => {
+	// Never let a broker leak into the next test's latestBroker()
+	resetMqttBrokers()
+})
+
+describe('Gateway teardown releases the broker', () => {
+	it('closing releases the broker and is safe to call twice', async () => {
+		const harness = await createGatewayHarness()
```
AlCalzone: **[Minor] · Tests**

The harness is created and closed inline inside each `it`, with no `afterEach(close)`. If an assertion throws before `await harness.close()`, that test's real broker (`.end()`) and Gateway timers never tear down — only the module-global watchers are caught by `afterAll(closeWatchers)`.

**Why:** leak-on-throw window, and it's a third teardown style distinct from the other files.

**Fix:** defensible for a suite whose subject *is* `close()`, but a shared `useGatewayHarness()` accessor would close unconditionally in `afterEach` and remove the window.

test/lib/hass/mqttLifecycle.test.ts
```diff
@@ -0,0 +1,510 @@
...
+				(p) => p.topic === SWITCH_DISCOVERY_TOPIC,
+			),
+		).toBe(true)
+
+		// Inbound is dropped while offline: a real broker never pushes to a
+		// disconnected client
+		harness.resetPublishes()
+		harness.broker.deliver('homeassistant/status', 'online')
+		await tick()
+		expect(harness.broker.published).toHaveLength(0)
```
AlCalzone: **[Minor] · Tests**

"before connecting… inbound dropped" (and `:175` "going offline stops routing") proves `broker.deliver`'s own `connected`+subscription gate (`mqttMock.ts:184`), not any `MqttClient` guard.

**Why:** the drop happens inside the mock, so this assertion can't detect a regression in production inbound routing.

**Fix:** `:495-509` shows the right pattern — `deliverRaw` bypasses the mock gate and exercises the real `parts[1] !== this._clientID` guard. Mirror that wherever the intent is to characterize MqttClient's own defenses.

test/lib/hass/mqttLifecycle.test.ts
```diff
@@ -0,0 +1,510 @@
...
+		expect(harness.zwave.writeBroadcast).toHaveBeenCalledTimes(1)
+		const [vId, value] = harness.zwave.writeBroadcast.mock.calls[0]
+		expect(value).toBe(true)
+		expect(vId.commandClass).toBe(CommandClasses['Binary Switch'])
+
+		// Feedback echoed to the same topic without /set
+		const echoed = harness.broker.published.find(
+			(p) => p.topic === `zwave/_CLIENTS/${cid}/broadcast`,
+		)
+		expect(echoed).toBeDefined()
```
AlCalzone, line 409: **[Minor] · Tests**

The broadcast "echoes feedback" test asserts only `expect(echoed).toBeDefined()` — topic existence, never the echoed payload/value the title implies.

**Why:** weak oracle; a wrong echo payload would still pass.

**Fix:** assert the echoed body/value, not just that a publish landed on the topic.

test/lib/hass/mqttLifecycle.test.ts
```diff
@@ -0,0 +1,510 @@
...
+		)
+		await tick()
+
+		expect(harness.zwave.callApi).toHaveBeenCalledWith('getNodes')
+
+		const ackTopic = `zwave/_CLIENTS/${cid}/api/getNodes`
+		const ack = harness.broker.published.find((p) => p.topic === ackTopic)
+		expect(ack).toBeDefined()
+		const parsed = JSON.parse(ack.payload)
+		expect(parsed.success).toBe(true)
```
AlCalzone: **[Minor] · Tests**

`expect(parsed.success).toBe(true)` asserts the `createFakeGatewayZwave.callApi` default `{success:true}` (`fixtures.ts:248`), not envelope logic — `_onApiRequest` only adds `origin` to whatever `callApi` returns, so this cannot fail regardless of production behavior.

**Why:** the assertion characterizes the fixture, not the ACK-envelope construction this test claims to lock.

**Fix:** the sibling assertions here (routed as `getNodes`, ACK topic without `/set`, `origin === payload`) are the genuine ones — either drop the `success` line or drive a fake returning `success:false` to prove the envelope maps it through.

test/lib/hass/discoverySwitches.test.ts
```diff
@@ -0,0 +1,442 @@
...
+		})
+
+		const device = discoverValueOnNode(harness.gw, node, key)
+		expect(device).toBeDefined()
+		expect(device.type).toBe('switch')
+		expect(device.object_id).toBe('switch')
+
+		const published = harness.lastDiscovery()
+		expect(published.topic).toBe('homeassistant/switch/Dev/switch/config')
+		expect(published.options).toEqual({ qos: 0, retain: false })
```
AlCalzone, line 100: **[Minor] · Tests**

Every discovery test asserts `{qos:0, retain:false}`, but `defaultMqttConfig` already sets `qos:0` and `MqttClient.publish` merges config defaults, so dropping `publishDiscovery`'s explicit `qos:0` (`Gateway.ts:776`) wouldn't fail any of them.

**Why:** the suite doesn't actually pin discovery's qos independent of config. `retain` is properly isolated (`mqttLifecycle.test.ts:200` flips `retainedDiscovery`); qos isn't.

**Fix:** add one case with a non-zero `mqttConfig.qos` and assert discovery still publishes `qos:0`.

test/lib/hass/server.test.ts
```diff
@@ -0,0 +1,377 @@
...
+		await zwave.close(true)
+	})
+})
+
+describe('reported server version', () => {
+	it('exposes the upstream @zwave-js/server version', () => {
+		const zwave = new ZWaveClient(
+			{ serverEnabled: true } as any,
+			createRecordingSocket() as any,
+		)
```
AlCalzone, line 394: **[Minor] · Tests**

`serverVersion` is asserted against `hoisted.SERVER_VERSION`, the value the `@zwave-js/server` mock itself exports.

**Why:** the oracle is the mock's own constant, so it confirms the pass-through wiring exists but proves nothing about the value.

**Fix:** acceptable as a wiring check; consider asserting the source/shape rather than mock-constant-equals-mock-constant.

test/lib/http/auth.test.ts
```diff
@@ -1,11 +1,11 @@
-import { describe, it, expect, afterEach } from 'vitest'
+import { describe, it, expect, beforeEach } from 'vitest'
 import { useHttpHarness } from './harness.ts'
 import { seedUser, signUserToken, setSettings } from '../shared/authHelpers.ts'
 
 describe('HTTP contract: auth & password', () => {
 	const getHarness = useHttpHarness()
 
-	afterEach(async () => {
+	beforeEach(async () => {
```
AlCalzone: **[Minor] · Tests**

Moving setup from `afterEach` to `beforeEach` is a latent footgun: `beforeEach` calls `getHarness()` with no options and the harness memoizes via `current ??=`, so the default app is force-instantiated before any test body runs. A future test calling `getHarness({ gateway })` would silently get the default instance with its options ignored.

**Why:** no current impact (no auth test passes options), but it's a maintenance trap that contradicts the per-test-options design.

**Fix:** seed via the test's own first `getHarness(...)` call, or have `beforeEach` not pre-instantiate the harness.

test/lib/socket/inboundApis.test.ts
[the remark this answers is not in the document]
robertsLando: yes sincerly dunno why it behaves like this now but it makes no sense, we should align subscribe and unsubscrivbe behaviour, an helper that manage channels would be good here for DRY

api/lib/Gateway.ts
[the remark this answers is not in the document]
AlCalzone: Fixed in bacd03e6. The `.js` branch now evicts `requireCustomDevices.cache[requireCustomDevices.resolve(basePath)]` before re-requiring, so a rewritten catalog is picked up on reload (new regression test rewrites the file, reloads, and asserts the changed object_id + new sha). Re: ESM — this loader is deliberately CommonJS via `createRequire(import.meta.url)`; its `.cache` is `Module._cache`, so eviction works both in native-ESM dev (where ambient `require` is undefined) and in the esbuild CJS release (where `import.meta.url` is shimmed to `file:${__filename}`). Verified in dev, build, and bundle plus a standalone runtime repro.

test/lib/socket/inboundApis.test.ts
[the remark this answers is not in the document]
AlCalzone: No bug-as-contract coverage was restored here: the `channels:['all']` UNSUBSCRIBE asymmetry is a genuine bug (not a quirk worth locking in a characterization test), so a targeted fix belongs outside this HASS review rather than a test that pins the buggy behavior.

test/lib/socket/inboundApis.test.ts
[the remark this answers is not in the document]
AlCalzone: Fixed in a5c7359d by changing production rather than pinning the silent-success bug: the HASS `api` switch now throws on an unknown `apiName` (mirroring the ZNIFFER default), which the handler turns into a `success:false` ack. Added a real socket route test asserting that public failure ack instead of the internal error.

test/lib/socket/inboundApis.test.ts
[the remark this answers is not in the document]
AlCalzone: Fixed in a5c7359d: `zniffer.loadCaptureFromBuffer(buffer)` is now awaited so the resolved value reaches the ack (previously the un-awaited Promise serialized to `{}`). The inboundApis socket test drives a resolved value and asserts it on the wire ack.

test/lib/hass/discoverySensors.test.ts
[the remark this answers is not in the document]
AlCalzone: Fixed in cbc2c98e: the object_id oracle now uses locale-independent `toLowerCase()` so it can't drift under locale casing rules (e.g. Turkish) on CI.

test/lib/hass/gatewayHarness.ts
AlCalzone: Fixed in e500dfb3. Added `useGatewayHarness()` following the `useHarnessLifecycle` convention (both now delegate to a shared `useManagedCurrent` primitive in `shared/harness.ts`) with internal `afterEach(close)`/`afterAll(cleanup)`, and replaced all 5 duplicated `beforeEach/afterEach/afterAll` trios so a file can't forget teardown.

test/lib/hass/gatewayHarness.ts
AlCalzone: Fixed in e500dfb3: removed the misleading `resetState` alias and its inaccurate docstring; there is now a single `resetPublishes()`.

test/lib/hass/gatewayHarness.ts
AlCalzone: Fixed in e500dfb3: dropped the `cleanupGatewayHarnessEnv` pass-through, and the managed harness `afterAll` now uniformly runs `closeWatchers()` + `cleanupTestEnv()` for every harness-based file, so the watcher safety net is applied consistently.

test/lib/hass/cleanup.test.ts
AlCalzone: Fixed in e500dfb3: `cleanup.test.ts` now uses the managed `useGatewayHarness()` accessor with an unconditional `afterEach(close)`, closing the leak-on-throw window.

test/lib/hass/mqttLifecycle.test.ts
AlCalzone: Fixed in cbc2c98e: dropped the assertions that only exercised the mock's own `connected`+subscription gate. The pre-connect test now only claims outgoing discovery still publishes, and the offline case is reframed to assert the real `_onOffline`/`_onConnect` reconnect re-announce.

test/lib/hass/mqttLifecycle.test.ts
AlCalzone: Fixed in cbc2c98e: the broadcast test now asserts `JSON.parse(echoed.payload)` deep-equals the sent payload, so a wrong echo value fails.

test/lib/hass/mqttLifecycle.test.ts
AlCalzone: Fixed in cbc2c98e: the test now drives a non-default `callApi` result (`{success:false, message, result}`) and asserts the ack payload deep-equals that result plus the added `origin`, with `{qos:0, retain:false}` options — so it exercises the envelope construction, not the fixture default.

test/lib/hass/discoverySwitches.test.ts
AlCalzone: Fixed in cbc2c98e: added a discovery test that sets `mqttConfig.qos:2` and asserts the published discovery is still `qos:0`, so dropping `publishDiscovery`'s explicit `qos:0` now fails (verified non-vacuous). `retain` remains independently asserted.

test/lib/hass/server.test.ts
AlCalzone: This is intentional and a false positive as a bug: it's a deliberate pass-through wiring assertion — we're locking that `getInfo` surfaces the `serverVersion` exported by `@zwave-js/server`, not the value itself. Renamed the test and added a comment in 70970407 to make that purpose explicit.

test/lib/http/auth.test.ts
AlCalzone: Fixed in 70970407: replaced the eager `beforeEach(getHarness)` with a lazy, option-aware `getHarness(options?)` wrapper that seeds settings once per harness instance and lets the first call supply options, so a future `getHarness({gateway})` is honored instead of silently getting the default app. Isolation/order-independence preserved.

robertsLando (approved):

Issue #2937, opened 2023-02-01
Error while fetching store files: EPERM


jriker1:
### Checklist

- [X]  I am not using Home Assistant. Or: a developer has told me to come here.
- [X] I have checked the troubleshooting section and my problem is not described there.
- [X] I have read the changelog and my problem is not mentioned there.

### Deploy method

PKG executable

### Z-Wave JS UI version

8.8.0

### ZwaveJS version

10.4.0

### Describe the bug

Not sure if it was when archiving but have Z-WaveJS UI running on Windows 2022 with Hyper-V role.  There is a log from teh 29th in the log folder as of 2/1 and ther eis also a tar.gz file for that day.  I can't access that log file or change ownership.  When I go to the Store tab I get:

Error while fetching store files: EPERM: operation not permitted, lstat 'E:\Z-WaveJSUI\store\logs\zwavejs_2023-01-29.log'

I have since deleted that log file but the system is still complaining about it when I open up the Z-WaveJS UI > Store location again fresh in a browser.

### To Reproduce

Launch Z-WaveJS UI  and click on Store.

### Expected behavior

It shows a tree of files/folders.

robertsLando:
I'm not a windows expert but how did you start zui? It could be you runned it with admin privileges and that created that file, try to right click on the file to check its ownership and fix it. Also be sure you are not running zui as administrator

jriker1:
I am running the application as a domain admin so anything being done with that executable would be with that account.  It's running with NMMS and set to run as m domain user.  That file is a log file that ZUI creates and looks like when it archives it at night to a tar.gz file it does something to the original file and it fails to be removed.  When I try and view the permissions on that file it says there is no owner.  Happened again when I tried to look at the logs thru the GUI today.

robertsLando:
I have submitted a fix for this, at least it should not throw, the problem is that file seems is still present but when I try to get stats that error is thrown, no clue why BTW, maybe the log rotation delete doesn't completely delete the file and a dead reference remains in memory, no clue as cannot repro this on linux system

jriker1:
Odd, next day it's still complaining about the 2023-02-01.log file.  Figured it would start complaining about 2023-02-02 but looks like there are no new log files being generated.  Below is what I see at 7:53AM on 2/3/2023 CST.
![image](https://user-images.githubusercontent.com/1495536/216620512-04df850d-6f96-4cc8-ba35-e28401ebe564.png)

[...]

PR #4726, opened 2026-07-11
refactor(types): establish strict backend boundaries

AlCalzone:
Introduces strict-ready shared boundaries without changing external behavior.

- Adds typed error, JSON, settings, store, and authentication boundaries used by later extractions.
- Keeps malformed legacy inputs and session serialization behavior compatible.
- Review focus: error coercion, nullable persisted data, and guards that must not change failure semantics.

Part of #4722. Stacked on #4725.


test/lib/errors.test.ts
```diff
@@ -0,0 +1,191 @@
+import { describe, it, expect } from 'vitest'
+import {
+	isError,
+	hasMessage,
+	hasErrorCode,
+	getErrorMessage,
+	toError,
+} from '../../api/lib/errors.ts'
+import type { ErrnoException } from '../../api/lib/errors.ts'
```
AlCalzone: Previous PRs have changed the `../../api` imports to use an alias (testing only). We should do the same here.

AlCalzone (commented):
## Review summary

**Verdict:** Ship with minor changes. The approach is sound and behavior-preservation is real — throw→throw preserved in `parseSecurityKeys`, `passwordHash` stripped on client-facing paths, and error coercion keeps winston's stack capture. No blockers.

**Top 3 risks:**
1. `errors.ts` *duplicates* `ErrnoException` (and re-implements `getErrorMessage`) instead of consolidating — production code still uses the `utils.ts` versions, so the "centralized error boundary" isn't the one anyone uses.
2. `json.ts` ships with zero production consumers — dead in this PR.
3. Two genuine runtime changes ride under a "types" label (`parseSecurityKeys` typed throws; logger transport-generation fix) — both correct, but worth pointing reviewers at given the "no external behavior change" framing.

**Theme — boundary primitives added but not wired in:** both `errors.ts` and `json.ts` duplicate existing symbols (`utils.ErrnoException`, `@zwave-js/shared`'s `getErrorMessage`/`JSONObject`) and are consumed only by their own tests. If the intent is one strict boundary, these should *replace* the old symbols, not sit beside them. Acceptable as pre-landed primitives if the consuming stack layer is imminent.

**Strengths:** tests exercise real behavior (real `express-session` + `session-file-store` round-trip; genuine Proxy/circular/hostile-value coverage), and the constructor-site casts are honest, localized bridges locked by the new boundary tests.

**Also (pre-existing, carried through — a future observability pass, not this PR):**
- `ZnifferManager.init()` (~line 140) retries forever logging only `info('Retrying in 5s...')` with no error context.
- Socket JWT verify failures (app.ts ~274-283) collapse to a generic `Error` without logging the reason (expired vs malformed vs bad signature).
- No test asserts `setupAll()` closes `.close()` across *every* tracked logger generation, so a regression reintroducing the #2937 handle leak would pass CI.

**Carry-over:** the prior review thread on `test/lib/errors.test.ts:9` (use the `#api/...` test alias instead of the relative `../../api` path) is still open.

**Coverage:** ran Correctness, Security, Design/API, DRY, Tests, Operability, Readability. Performance not run — no files in its lane (types refactor, no query/hot-path/cache surface). `package-lock.json` not reviewed (generated; sole change is the `@types/session-file-store` devDep).

api/lib/errors.ts
```diff
@@ -0,0 +1,90 @@
+// Narrow helpers for values caught from a `catch` clause, avoiding `as any` scattered across the codebase
+
+/** A Node.js-style error carrying an errno/code, e.g. from fs/net */
+export interface ErrnoException extends Error {
```
AlCalzone: **[Major] · Design / DRY / Readability**

New `ErrnoException` duplicates the one at `utils.ts:41` (and drops its `stack?`), while every production caller (`app.ts:388,2242,2492`) still imports `utils.ErrnoException` — the new copy is referenced only by `errors.test.ts`. `isError`/`toError`/`hasMessage` likewise have no production call site.

**Why:** a PR whose stated purpose is a centralized error boundary ships two divergent same-named types plus test-only exports, so a later stack extraction can import the wrong one.

**Fix:** one source of truth — define here and re-export from `utils` (or vice-versa) — and either wire the unused guards into a real site or defer them.

test/lib/hass/persistence.test.ts
```diff
@@ -435,6 +435,14 @@ describe('home-id scoping of persisted nodes', () => {
 		// Assert the rejection contract without pinning the message text.
 		await expect(zwave.getStoreNodes()).rejects.toThrow()
 	})
+
+	it('rejects a malformed null node record during Home Assistant device removal (see #4736)', async () => {
```
AlCalzone: **[Major] · Tests**

The only malformed-node test asserts `storeDevices()` *crashes* with an unhandled `TypeError` on `{9: null}`, pinning the crash as the contract rather than testing recovery.

**Why:** one corrupt entry in `nodes.json` (partial write, manual edit, migration bug) takes down HA device-sync for that node with no graceful path.

**Note:** the crash is pre-existing (old code threw at the same point) and fixing it is out of scope for this behavior-preserving refactor. **Fix:** file a follow-up for skip/repair-on-malformed; not a blocker here.

api/lib/json.ts
```diff
@@ -0,0 +1,65 @@
+// Precise types for values that have gone through JSON.parse, i.e. what jsonStore reads from/writes to disk, instead of `any`
```
AlCalzone: **[Minor] · Design / DRY**

This whole module (`JsonValue`/`JsonObject`/`isJsonValue`/`isJsonObject`) has no production consumer — only `json.test.ts` imports it, and `jsonStore` was refactored to generics + `unknown`, not to `JsonValue`. A loose `JSONObject = Record<string, any>` already exists in `@zwave-js/shared`.

**Why:** runtime validation shipped that nothing calls (speculative generality with a 158-line test attached).

**Fix:** wire the guards into the `jsonStore` read/write boundary they were written for, or defer to the PR that consumes them.

api/lib/errors.ts
```diff
@@ -0,0 +1,90 @@
...
+function tryOrUndefined<T>(fn: () => T): T | undefined {
+	try {
+		return fn()
+	} catch {
+		return undefined
+	}
+}
+
+// Never throws, even for hostile values whose toJSON/toString/Symbol.toPrimitive or Proxy traps themselves throw
+export function getErrorMessage(value: unknown): string {
```
AlCalzone, line 33: **[Minor] · DRY**

`getErrorMessage` re-implements `@zwave-js/shared`'s exported `getErrorMessage(e, includeStack?)` (the package is already imported at `utils.ts:12`). The extra hostile-value/`JSON.stringify` hardening makes divergence defensible.

**Why:** duplicated primitive from an already-present dependency invites future "just use shared's" churn.

**Fix:** add a one-line comment on why the shared helper is insufficient.

api/lib/utils.ts
```diff
@@ -25,13 +27,16 @@ export interface Snippet {
 	content: string
 }
 
-export type DeepPartial<T> = {
-	[P in keyof T]?: T[P] extends Array<infer U>
+// Bottoms out at non-object types (primitives, functions, class instances) instead of recursing into their prototype members, since `keyof` on those produces a nonsensical mapped type
```
AlCalzone: **[Minor] · Design / Readability**

The DeepPartial comment claims it bottoms out at "class instances," but `Date`/`Map`/`Set`/class instances fall through to the same object branch (`DeepPartial<Date>` accepts `{}`, i.e. every method becomes optional).

**Why:** the comment gives a wrong mental model for exactly the non-obvious case it exists to clarify.

**Fix:** drop "class instances" from the parenthetical, or note they intentionally go through the object branch like plain objects.

api/app.ts
```diff
@@ -82,15 +84,32 @@ import {
...
 } from './lib/importConfig.ts'
+import { getErrorMessage } from './lib/errors.ts'
 
 const createCertificate = promisify(generate)
 
 declare module 'express-session' {
 	export interface SessionData {
-		user?: User
+		// Session files may contain the full user record, including passwordHash; see #4739
+		user?: User | PublicUser
```
AlCalzone, line 94: **[Minor] · Security / Tests**

This `User | PublicUser` session type is a fiction over the runtime object: `req.session.user` is assigned the full `User` (with `passwordHash`) on the JWT (~1099) and `PUT /api/password` (~1252) paths, so the bcrypt hash is written to the on-disk `session-file-store` files.

**Why:** duplicates the secret beyond `users.json`; not client-facing (cookie holds only the session id). Pre-existing, tracked as #4739, but institutionalized here.

**Fix:** strip to `PublicUser` before assigning to `req.session.user` at both sites, as the JSON responses already do.

api/lib/utils.ts
```diff
@@ -401,47 +422,69 @@ export function parseSecurityKeys(
...
-			)
+		if (isKeyOf(key, availableKeys)) {
+			const value = config.securityKeys[key]
+			// Rejects persisted null standard keys (see #4736)
+			if (value === null) {
+				throw new TypeError(
+					`config.securityKeys.${key} is null; remove the key entirely instead of persisting it as null`,
+				)
+			}
+			if (value?.length === 32) {
```
AlCalzone: **[Minor] · Correctness**

`parseSecurityKeys` now tolerates an `undefined` key value (`value?.length === 32` → omit) where the prior code threw a `TypeError`; the new test bakes this in.

**Why:** contradicts the "no external behavior change" claim. Real-world impact is nil — `undefined` is unreachable from persisted JSON or env, and the `null` case still throws.

**Fix:** restore the throw on `undefined`, or document the intentional relaxation.

api/lib/logger.ts
```diff
@@ -106,15 +112,21 @@ export const logStream = new PassThrough()
  * Create the base transports based on settings provided
  */
 export function customTransports(config: LoggerConfig): winston.transport[] {
-	// setup transports only once (see issue #2937)
-	if (transportsList) {
+	const wantsFileTransport = Boolean(config.enabled && config.logToFile)
+	const key = `${config.enabled}:${wantsFileTransport}`
+
+	// Share matching transports within a configuration generation because duplicate rotation handles conflict (#2937)
+	if (transportsList && transportsListKey === key) {
```
AlCalzone: **[Nit] · Correctness**

`transportsListKey` is a last-wins single slot while `transportGenerations` accumulates every array; alternating `enabled`/`logToFile` keys within one generation could recreate a second `DailyRotateFile` on the same path (the #2937 rotation-handle conflict).

**Why:** latent regression of the #2937 invariant. Not currently triggerable — the only production driver (`setupAll`) feeds every logger a uniform key.

**Fix:** key the cache by generation-map, or assert single-key-per-generation.

api/lib/errors.ts
```diff
@@ -0,0 +1,90 @@
...
+	// A hostile Proxy's has/get traps can make the 'in' check and property read themselves throw
+	const messageCarrierMessage = tryOrUndefined(() =>
+		hasMessage(value) ? value.message : undefined,
+	)
+	if (typeof messageCarrierMessage === 'string') {
+		return messageCarrierMessage
+	}
+
+	// Falls through to the String() fallback below for undefined/functions, which JSON.stringify returns as undefined rather than throwing
+	const stringified = tryOrUndefined(() => JSON.stringify(value))
```
AlCalzone, line 55: **[FYI] · Security / Operability**

`getErrorMessage`'s `JSON.stringify(value)` fallback would serialize an arbitrary thrown object wholesale, and `.stack`/`.cause` are dropped.

**Why:** no live leak today (no call site throws a secret-bearing object; `parseSecurityKeys` throws messages with key *names* only), and the catches that log the full error object still feed winston's stack capture — but a future `throw someConfigObject` on these paths would dump it.

**Fix:** none required now; consider a guard or comment if this utility spreads.

api/lib/errors.ts
AlCalzone: why don't we just use the original version from Node.js's standard library?

test/lib/hass/persistence.test.ts
AlCalzone: We should not cement crashes as desired behavior. Rejecting entries should be errors, not crashes

api/lib/json.ts
AlCalzone: It sounds like this would make sense to use in jsonStore, but I'm not sure.

api/lib/errors.ts
AlCalzone: nah its fine.

api/lib/utils.ts
AlCalzone: Is this really necessary? Where is `DeepPartial` used? In zwave-js, it is only used for options objects that have no such problematic members.

api/app.ts
AlCalzone: This should be fixed in #4739 if needed

api/lib/utils.ts
AlCalzone: Lets check if this was intentional.

test/lib/errors.test.ts
AlCalzone: Fixed in 349fd5e0 — Layer 4 test imports now use the existing `#api/*` alias.

api/app.ts
```diff
@@ -503,12 +527,12 @@ export function createApp(options: CreateAppOptions = {}): AppInstance {
 	}
 
 	function setupLogging(settings: {
-		gateway: utils.DeepPartial<GatewayConfig>
+		gateway?: utils.DeepPartial<GatewayConfig>
 	}) {
 		loggers.setupAll(settings ? settings.gateway : null)
```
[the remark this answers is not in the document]
AlCalzone: Fixed in bae35e1c — logging now normalizes an omitted gateway config to `{}` before `setupAll()`.

api/lib/jsonStore.ts
```diff
@@ -91,7 +96,8 @@ export class StorageHelper {
 			})
 
 			for (const model in this.config) {
-				const config: StoreFile = this.config[model]
+				const config: StoreFile<unknown> =
+					this.config[model as StoreKeys]
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8ae8b548 — `backup()` now guards the optional store configuration explicitly.

api/lib/errors.ts
AlCalzone: Fixed in 349fd5e0 — duplicate errno declarations were removed in favor of `NodeJS.ErrnoException`; internal guards are private and the unused conversion helper is gone.

test/lib/hass/persistence.test.ts
AlCalzone: Fixed in 349fd5e0 — the crash-pinning malformed-node characterization was removed. Recovery behavior remains tracked by #4736.

api/lib/json.ts
AlCalzone: Fixed in 349fd5e0 — the unused JSON guard module and its helper-only tests were removed rather than changing `jsonStore` malformed-data behavior.

api/lib/utils.ts
AlCalzone: Fixed in 3522cf14 — the comment now describes the actual callable, array, and object behavior without claiming class-instance support.

api/app.ts
AlCalzone: No Layer 4 change — session serialization remains compatible here; password-field handling is deferred to #4739 as requested.

api/lib/utils.ts
AlCalzone: Fixed in 3522cf14 — explicitly undefined standard keys again fail with `TypeError`, preserving the prior failure contract without pinning an incidental message.

api/lib/logger.ts
AlCalzone: Fixed in bae35e1c — transports are now cached per configuration key for the active generation, preventing duplicate file transports after alternating configurations.

api/lib/errors.ts
AlCalzone: No change needed for this FYI. The helper preserves the existing public message behavior and has no current secret-bearing call site.

robertsLando (approved):

api/lib/utils.ts
```diff
@@ -401,47 +414,74 @@ export function parseSecurityKeys(
...
+			}
+			if (value.length === 32) {
+				securityKeys[key] = Buffer.from(value, 'hex')
+			}
 		}
 	}
 
+	options.securityKeys = securityKeys
+
+	// Defaulted only immediately before this loop, matching the pre-existing ordering where the earlier KEY_LR_* loop is the only path that can hit the missing-map failure
```
AlCalzone: WTF is this comment?

api/lib/ZnifferManager.ts
```diff
@@ -181,10 +183,12 @@ export default class ZnifferManager extends TypedEventEmitter<ZnifferManagerEven
 			.emit(socketEvents.znifferState, this.status())
 	}
 
-	private checkReady() {
+	// Returns `this.zniffer` so callers can narrow it via a local variable instead of re-reading the nullable field
+	private checkReady(): Zniffer {
```
AlCalzone: This no longer purely checks. How about `ensureReady`

api/lib/ZwaveClient.ts
```diff
@@ -2781,16 +2786,19 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 			nodes = storeNodes
 		}
 
+		// Runtime key check can't narrow this union type, so each branch below casts to whichever shape the check actually implies
```
AlCalzone: remove comment. it does not clarify anything

api/lib/ZwaveClient.ts
```diff
@@ -2801,7 +2809,8 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
 				return
 			}
 
-			const nodes = jsonStore.get(store.nodes)
+			// Assumes the current home-ID-keyed shape, true once getStoreNodes has migrated any legacy on-disk shape
```
AlCalzone:
```suggestion
			// Assumes store.nodes is keyed by home ID, which is true once getStoreNodes has migrated any legacy on-disk shape
```

api/lib/utils.ts
AlCalzone: Fixed in 8a3df47e — replaced this with a concise compatibility reason tied to #4736.

api/lib/ZnifferManager.ts
AlCalzone: Fixed in 8a3df47e — renamed `checkReady()` to `ensureReady()` at every call site.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in 8a3df47e — removed the redundant comment.

api/lib/ZwaveClient.ts
AlCalzone: Fixed in 8a3df47e — applied the suggested wording.

PR #4727, opened 2026-07-11
refactor(api): extract runtime and http routers

AlCalzone:
Separates application lifecycle and HTTP routing from `app.ts`.

- Introduces `AppRuntime` and focused routers while keeping the existing endpoints and middleware contracts.
- Makes restart generations and live service ownership explicit.
- Review focus: stale service capture, restart behavior, and teardown ordering.

Part of #4722. Stacked on #4726.


api/runtime/AppRuntime.ts
```diff
@@ -0,0 +1,306 @@
...
+	private readonly deps: AppRuntimeDeps
+
+	constructor(deps: AppRuntimeDeps) {
+		this.deps = deps
+		this.gateway = deps.gateway
+		this.zniffer = deps.zniffer
+		this.restarting = deps.restarting ?? false
+	}
+
+	getGateway(): GatewayPort | undefined {
```
robertsLando: why not a getter instead?

robertsLando: same for others

api/routes/store.ts
```diff
@@ -0,0 +1,395 @@
...
+}
+
+const Storage = diskStorage({
+	async destination(reqD, file, callback) {
+		await utils.ensureDir(tmpDir)
+		callback(null, tmpDir)
+	},
+	filename(reqF, file, callback) {
+		callback(null, file.originalname)
+	},
```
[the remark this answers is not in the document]
AlCalzone: This is already constrained by Multer’s default `preservePath: false`: Busboy applies `path.basename()` before assigning `file.originalname`, so traversal segments never reach disk storage.

api/routes/store.ts
```diff
@@ -0,0 +1,395 @@
...
+				}
+
+				res.json({ success: true })
+			} catch (err) {
+				res.json({ success: false, message: getErrorMessage(err) })
+			}
+
+			if (file && isRestore) {
+				await rm(file.path)
+			}
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 7264ed06. Restore cleanup now uses forced removal and logs cleanup failures instead of rejecting after the response.

api/routes/settings.ts
```diff
@@ -0,0 +1,466 @@
...
+				let settings = req.body
+
+				let shouldRestart = false
+				let shouldRestartGw = false
+				let shouldRestartZniffer = false
+				let canUpdateZwaveOptions = false
+
+				const actualSettings = jsonStore.get(store.settings)
+
+				// TODO: validate settings using calss-validator
```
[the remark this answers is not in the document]
AlCalzone: Fixed in fea78de8.

test/lib/shared/fakes.ts
```diff
@@ -131,3 +146,37 @@ export function createFakeGateway(
...
+		status: vi.fn(() => ({ active: false, frequency: undefined })),
+		start: vi.fn(() => Promise.resolve(undefined)),
+		stop: vi.fn(() => Promise.resolve(undefined)),
+		clear: vi.fn(() => undefined),
+		getFrames: vi.fn(() => []),
+		setFrequency: vi.fn(() => Promise.resolve(undefined)),
+		setLRChannelConfig: vi.fn(() => Promise.resolve(undefined)),
+		saveCaptureToFile: vi.fn(() => Promise.resolve('/tmp/capture.zlf')),
+		// Returns a promise like the real method, so the missing await in production leaves an unresolved promise that serializes to {} over the wire
+		loadCaptureFromBuffer: vi.fn(() => Promise.resolve(undefined)),
```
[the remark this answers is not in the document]
AlCalzone: Fixed in fea78de8. The comment now states only that the fake mirrors the real async API.

api/runtime/AppRuntime.ts
```diff
@@ -0,0 +1,306 @@
...
+				this.deps.getSocketServer(),
+			)
+		}
+
+		// Backup initialization is valid when Z-Wave is disabled.
+		backupManager.init(zwave, this.backupManagerOwner)
+
+		const gw = new Gateway(settings.gateway as GatewayConfig, zwave, mqtt)
+		this.setGateway(gw)
+
```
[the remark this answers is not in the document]
AlCalzone: Fixed in fea78de8. Runtime locals, Gateway ownership, and plugin context now model disabled clients as explicitly optional.

api/runtime/AppRuntime.ts
AlCalzone: Changed in fea78de8. Gateway, Zniffer, plugin router, restart state, and debug ownership now use getters while preserving request-time freshness.

api/routes/settings.ts
```diff
@@ -0,0 +1,466 @@
...
+						? 'Configuration saved. Restart required to apply changes.'
+						: 'Configuration updated successfully',
+					data: settings,
+					shouldRestart,
+				})
+			} catch (error) {
+				runtime.setRestarting(false)
+				logger.error(error)
+				res.json({ success: false, message: getErrorMessage(error) })
+			}
```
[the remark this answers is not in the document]
AlCalzone: Fixed in fea78de8. The settings route no longer releases restart state it does not own, with a repeated-request regression.

test/lib/shared/fakes.ts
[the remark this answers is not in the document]
AlCalzone: this comment is completely unnecessary IMO. lets remove it completely.

api/lib/BackupManager.ts
```diff
@@ -26,7 +26,7 @@ class BackupManager {
 	private config!: BackupSettings
 	private storeJob?: Cron
 	private nvmJob?: Cron
-	private zwaveClient!: ZwaveClient
+	private zwaveClient?: Pick<ZwaveClientPort, 'backupNVMRaw'>
```
AlCalzone: This deserves its own type at the module level.

api/lib/BackupManager.ts
```diff
@@ -111,6 +114,9 @@ class BackupManager {
 		logger.info('Backup NVM started')
 
 		try {
+			if (!this.zwaveClient) {
+				throw new Error('Z-Wave client not initialized')
+			}
```
AlCalzone: We have some other places where we've factored this out to `const zwaveClient = this.ensureReady()`. We should do this consistently.

api/routes/configurationTemplates.ts
```diff
@@ -0,0 +1,238 @@
...
+
+export interface ConfigurationTemplatesRoutesDeps {
+	apisLimiter: RateLimitRequestHandler
+}
+
+/**
+ * Every route below resolves the Z-Wave client via `runtime.requireZwaveClient()`,
+ * which throws a clean typed error when none is attached, caught by each
+ * route's own try/catch
+ */
```
AlCalzone: I don't get why this is a JSDoc comment.

api/routes/health.ts
```diff
@@ -0,0 +1,57 @@
...
+		const gw = runtime.gateway
+		if (gw) {
+			mqtt = gw.mqtt?.getStatus() ?? false
+			zwave = gw.zwave?.getStatus().status ?? false
+		}
+
+		// Disabled mqtt reports a status object, not a boolean, but still counts as healthy (see #469)
+		if (mqtt && typeof mqtt !== 'boolean') {
+			mqtt = mqtt.status || mqtt.config.disabled
+		}
```
AlCalzone: We could solve most of the type confusion by doing this check inside the `if (gw)` branch.
Rename the top level variables to `mqttHealthy` and `zwaveHealthy`, and assign the result of the check. Btw, does `gw.mqtt?.getStatus()` always return an object? In that case:
```
const mqttStatus = gw.mqtt?.getStatus();
mqttHealthy = (mqttStatus?.status || mqttStatus?.config.disabled) ?? false
```
(double check the operator preference on this last line)

api/runtime/AppRuntime.ts
```diff
@@ -0,0 +1,306 @@
...
+	private _ownsDebugSession = false
+
+	// `BackupManager.init()`/`.close()`'s `owner` parameter doesn't exist in
+	// this layer's own `BackupManager.ts` - it's this (base) layer's own,
+	// independently-added ownership token. Scoping it per `AppRuntime`
+	// instance (mirroring `api/app.ts`'s pre-extraction per-`createApp()`
+	// local) keeps one instance's shutdown from tearing down a different,
+	// concurrently-live instance's backup manager state (e.g. two
+	// `AppRuntime`s constructed by separate test suites within the same
+	// process).
```
AlCalzone: Too much bla bla. Keep focused, use simple sentences. We're not in a philosopher's competition who can create the most convoluted sentences.
Apply to all comments you added.

test/lib/shared/fakes.ts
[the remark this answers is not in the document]
AlCalzone: Removed in 788e6c7e.

api/runtime/AppRuntime.ts
```diff
@@ -0,0 +1,306 @@
...
+		}
+		return this._gateway
+	}
+
+	requireZwaveClient(): ZwaveClientPort {
+		if (this._gateway?.zwave === undefined) {
+			throw new Error('Z-Wave client not inited')
+		}
+		return this._gateway.zwave
+	}
```
AlCalzone: The naming of these methods across the different modules is inconsistent. Others call it `ensureZWaveClient`. Pick one.

api/runtime/ports.ts
```diff
@@ -0,0 +1,85 @@
...
+		| 'dumpNode'
+		| 'getNode'
+		| 'nodes'
+		| 'restart'
+		| 'setUserCallbacks'
+		| 'removeUserCallbacks'
+		| 'backupNVMRaw'
+	>,
+	'driver' | 'nodes'
+> & {
```
AlCalzone: Why pick them at all, if you omit them immediately. Can we not do `Omit<ZWaveClient, 'driver' | 'nodes'>`?

AlCalzone: Same for the other types

api/lib/BackupManager.ts
AlCalzone: Fixed in aceaae14. Added the module-level `BackupZWaveClient` type.

api/lib/BackupManager.ts
AlCalzone: Fixed in aceaae14. `backupNvm()` now uses `const zwaveClient = this.ensureZWaveClient()`.

api/runtime/ports.ts
AlCalzone: Simplified in aceaae14. The narrow `Pick` remains intentional for port isolation, but fields replaced by custom port types are no longer picked and immediately omitted.

api/lib/serialPorts.ts
AlCalzone, line 1: I thought we had eliminated this because it was only used for testing. There is now a parameter on the app factory.

api/routes/auth.ts
```diff
@@ -0,0 +1,266 @@
...
+					`User ${attemptedUsername} failed to login from ${req.ip}: wrong credentials`,
+				)
+			}
+
+			res.json(result)
+		} catch (error) {
+			res.json({
+				success: false,
+				message: 'Authentication failed',
+				code: 3,
```
AlCalzone: I suppose the error codes should be a shared constant.

api/app.ts
```diff
@@ -154,11 +67,9 @@ export interface AppInstance {
 
 export interface CreateAppOptions {
 	test?: {
-		gateway?: Gateway
-		zniffer?: ZnifferManager
-		pluginsRouter?: Router
+		gateway?: GatewayPort
+		zniffer?: ZnifferPort
 		restarting?: boolean
-		enumerateSerialPorts?: typeof Driver.enumerateSerialPorts
```
AlCalzone, line 161: Why was this removed? We added this to be able to test this function without needing an extra module

test/lib/http/authRateLimit.test.ts
```diff
@@ -25,39 +30,11 @@ describe('HTTP contract: rate limiting', () => {
...
-		for (let attempt = 1; attempt <= 501; attempt++) {
-			const res = await harness.request.get('/health')
-			lastBody = res.body
-		}
-
-		expect(lastBody).toEqual({
-			success: false,
-			message: 'Max requests limit reached',
-		})
-	})
```
AlCalzone, line 62: where did these test go?

test/lib/http/harness.ts
```diff
@@ -14,13 +16,22 @@ import {
...
+		Promise.resolve<string[]>([]),
+	),
+}))
 
-function createSerialPortsEnumeratorMock(): SerialPortsEnumeratorMock {
-	return vi.fn(() => Promise.resolve<string[]>([]))
-}
+vi.mock('#api/lib/serialPorts.ts', () => ({ enumerateSerialPorts }))
+
+export { enumerateSerialPorts }
```
AlCalzone: The entire point of having an app factory and doing dependency injection was NOT needing a separate module.

test/lib/http/routeContract.test.ts
AlCalzone, line 1: What happened here?

package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "zwave-js-ui",
   "type": "module",
+  "imports": {
+    "#api/*": "./api/*"
+  },
```
AlCalzone: Previous PRs removed this in favor of vitest path mappings. We should stick with that.

AlCalzone (commented):
Make sure we don't remove tests without a good reason.

api/lib/serialPorts.ts
AlCalzone: Fixed in f16a8090. Serial enumeration is injected through createApp again, and the test-only wrapper module is removed.

api/routes/auth.ts
AlCalzone: Fixed in f16a8090. The auth response codes now live in the shared ResponseCodes module.

api/app.ts
AlCalzone: Fixed in f16a8090. createApp now accepts the serial-port enumerator and passes it to the settings routes.

test/lib/http/authRateLimit.test.ts
AlCalzone: Restored in ad6fc132. The store and API limiter envelope tests are back alongside the login limiter test.

test/lib/http/harness.ts
AlCalzone: Fixed in f16a8090 and ad6fc132. The app factory owns the injection seam, and each HTTP harness now gets an isolated enumerator mock.

package.json
AlCalzone: Fixed in f16a8090. The package import map is removed; Vitest and ESLint path mappings remain authoritative.

test/lib/http/routeContract.test.ts
AlCalzone: Restored in ad6fc132. The recursive exact 35-route inventory, duplicate detection, and mounted-router traversal tests are back.

api/routes/auth.ts
AlCalzone: I meant `code: 3`, not the constants.

AlCalzone: Fixed in 4f06d38e. The numeric authentication error code is now shared and used by all three auth failure responses.

AlCalzone:
@robertsLando if you don't find anything else, feel free to merge.

api/runtime/AppRuntime.ts
```diff
@@ -0,0 +1,296 @@
...
+
+	get gateway(): GatewayPort | undefined {
+		return this._gateway
+	}
+
+	private setGateway(value: GatewayPort | undefined): void {
+		this._gateway = value
+	}
+
+	requireGateway(): GatewayPort {
```
robertsLando: naming with methods here are still inconsistet, use ensureXXX everywhere instead of requireXXX

api/app.ts
```diff
@@ -288,22 +190,18 @@ export function createApp(options: CreateAppOptions = {}): AppInstance {
...
-
 	let closed = false
 	let socketAttached = false
-	let ownsDebugSession = false
 	let logStreamInterceptor: ((chunk: Buffer | string) => void) | undefined
 
 	// ### UTILS
 
 	function installProcessHandlers() {
+		if (closed) {
```
robertsLando: could we infer closed from appruntime and socket attached from socketmanager?

robertsLando (changes requested):

api/runtime/AppRuntime.ts
robertsLando: Done in 03f1a19b — renamed `requireGateway`/`requireZniffer` to `ensureGateway`/`ensureZniffer` so the accessors are uniform. Also renamed `ZnifferManager.ensureReady` → `ensureZniffer` for the same reason.

api/app.ts
robertsLando: Done in 03f1a19b — dropped both local flags: `socketAttached` now derives from `socketManager.io`, and `closed` from a new idempotent `AppRuntime.isClosed`. Released debug-session ownership before its await so a concurrent `close()` cannot cancel the same session twice.

PR #4728, opened 2026-07-11
refactor(socket): extract protocol handlers

AlCalzone:
Moves Socket.IO request handling into typed protocol modules.

- Keeps event names, rooms, acknowledgement envelopes, argument order, timing, and dynamic API behavior unchanged.
- Resolves collaborators per call so restarts cannot leave stale handlers.
- Review focus: wire compatibility, first/last-client lifecycle, and production persistence paths.

Part of #4722. Stacked on #4727.


api/socket/hassApi.ts
```diff
@@ -0,0 +1,139 @@
...
+			let res: void
+			let err: string | undefined
+			try {
+				switch (data.apiName) {
+					case 'delete':
+						{
+							const gateway = runtime.ensureGateway()
+							res = Reflect.apply(
+								gateway.publishDiscovery.bind(gateway),
+								undefined,
```
AlCalzone: I thought the 2nd argument is meant to pass the `this` context - why do we need the explicit `.bind(gateway)` in the first argument then?

api/socket/mqttApi.ts
```diff
@@ -0,0 +1,72 @@
...
+			let err: string | undefined
+
+			try {
+				switch (data.api) {
+					case 'updateNodeTopics':
+						{
+							const gateway = runtime.ensureGateway()
+							res = Reflect.apply(
+								gateway.updateNodeTopics.bind(gateway),
+								undefined,
```
AlCalzone: same here

api/socket/types.ts
```diff
@@ -0,0 +1,3 @@
+export type SocketAck<T> = (result: T) => void
+
+export const noop = (): void => {}
```
AlCalzone: This is not a type. Why is it in a file called `types.ts`?

api/socket/mqttApi.ts
AlCalzone: ...and in other files too

api/socket/types.ts
AlCalzone: Also what is the functional reason for having noop as a callback for socket events?

api/socket/hassApi.ts
```diff
@@ -0,0 +1,139 @@
...
+			logger.info(`Hass api call: ${data.apiName}`)
+
+			let res: void
+			let err: string | undefined
+			try {
+				switch (data.apiName) {
+					case 'delete':
+						{
+							const gateway = runtime.ensureGateway()
+							res = Reflect.apply(
```
AlCalzone: **[Major · Design / DRY / Readability]**

`Reflect.apply(gateway.publishDiscovery.bind(gateway), undefined, [...])` appears 11 times across `hassApi.ts` (42, 59, 76, 86, 96, 106, 116), `mqttApi.ts` (39, 49), and `znifferApi.ts` (61, 71). It exists only because `HassApiRequest.device`/`nodeId` are optional and `mqttApi`'s `data.args[0]` is `unknown`, while the callees require concrete types — the base called these directly as `runtime.ensureGateway().publishDiscovery(...)`.

**Why:** the stated win of this PR is a *typed* protocol layer, and every one of these sites opts out of both arity and argument-type checking, so the new request interfaces document a contract nothing enforces and a `GatewayPort`/`ZnifferPort` signature change will compile silently. Three further signals that this isn't load-bearing: the idiom appears nowhere else in the repo; the relevant `no-unsafe-*` rules are `off` in `eslint.config.js`, so it isn't satisfying a lint rule; and `znifferApi.ts` calls 6 of its 8 cases directly and wraps only the two that take arguments — which confirms the mechanism is type-laundering, not policy. Each site also forces a block scope plus a local alias, roughly tripling every switch arm.

**Fix:** call the collaborators directly. Either narrow the request types so the required fields are non-optional, or reject missing fields through the existing `err = ...` path.

This supersedes the two open `.bind(gateway)` threads — the whole `Reflect.apply` should go, not just the redundant bind.

test/lib/socket/harness.ts
```diff
@@ -127,12 +113,7 @@ async function createHarnessInstance(
 		}
 		clients.clear()
 
-		// Best-effort: lets the server finish settling activeSockets/room membership before returning, but a test that already awaited its own disconnect may have nothing left to wait for
-		try {
-			await waitForServerSocketCount(0, 1000)
-		} catch {
-			// ignore - see above
-		}
+		await waitForServerSocketCount(0, 1000)
```
AlCalzone, line 116: **[Major · Correctness]**

Dropping the `try/catch` around this call makes a teardown timeout escape `disconnectAllClients`, and `closeInstance` (line 131) awaits it *before* `await instance.close()` with no `finally`.

**Why:** one slow teardown leaks the HTTP server, `io`, and the gateway/zniffer into every later test in the file, and the leak surfaces as unrelated failures downstream rather than as the real timeout.

**Fix:** keep the assertion strict — enforcing transport teardown is the right call — but guarantee `instance.close()` runs regardless of whether the wait succeeded.

api/socket/subscriptions.ts
```diff
@@ -0,0 +1,69 @@
...
+		async (
+			data: ChannelSubscriptionRequest | undefined,
+			cb: SocketAck<ChannelSubscriptionAck> = noop,
+		) => {
+			const channels = requestedChannels(data)
+
+			const isAll = channels.includes('all')
+			const validChannels = isAll
+				? ALL_CHANNELS
+				: channels.filter((c) => Object.hasOwn(channelMap, c))
```
AlCalzone: **[Major · DRY]**

The `isAll`/`validChannels` resolution here is copied verbatim from lines 36-39. The base had a single `applyChannelSubscription(data, 'join' | 'leave', cb)`.

**Why:** the `'all'` special case and the `Object.hasOwn(channelMap, ...)` rule that gates event fan-out now have two edit sites, so a future `channelMap` rule change has to land in both or they drift.

**Fix:** restore a single resolver plus a shared join/leave applier, leaving the two `socket.on` registrations as thin wrappers.

api/socket/zwaveApi.ts
```diff
@@ -0,0 +1,89 @@
...
+): void {
+	socket.on(
+		inboundEvents.zwave,
+		async (data: ZwaveApiRequest, cb: SocketAck<ZwaveApiAck> = noop) => {
+			const currentGw = runtime.ensureGateway()
+			if (currentGw.zwave) {
+				if (!data.args) data.args = []
+				const callApi = currentGw.zwave.callApi.bind(
+					currentGw.zwave,
+				) as DynamicCallApi
```
AlCalzone, line 75: **[Major · Design / API]**

`callApi.bind(...) as DynamicCallApi` re-declares the call as returning `Promise<ZwaveApiAck>`, discarding the real `CallAPIResult<T>` that the pre-PR code imported from `ZwaveClient.ts` (and that `ZwaveClientPort` already exposes via `'callApi'`).

**Why:** this loses the `success: true | false` discriminated union, so `result` reads as present on failure acks, and it puts the ack shape in two places with a cast bridging them — drift in `callApi`'s result shape then reaches the frontend unchecked.

**Fix:** keep the collaborator's declared return type and intersect `{ api?: string }`, as the original did.

api/socket/hassApi.ts
```diff
@@ -0,0 +1,139 @@
...
+							const zwave = runtime.ensureZWaveClient()
+							res = await Reflect.apply(
+								zwave.storeDevices.bind(zwave),
+								undefined,
+								[data.devices, data.nodeId, data.remove],
+							)
+						}
+						break
+					default:
+						err = `Unknown HASS api ${data.apiName}`
```
AlCalzone: **[Minor · Correctness]**

The unknown-api default assigns `err` directly; the base threw here, which routed it through `logger.error('Error while calling HASS api', ...)`.

**Why:** unknown HASS calls stop appearing in error logs. `znifferApi.ts:89` still throws, so the two sibling handlers now disagree on how they surface the same failure class. The wire shape is unchanged either way.

**Fix:** log at the default branch, or keep the throw for symmetry with `znifferApi`.

api/socket/hassApi.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export interface HassApiRequest {
+	apiName?: string
+	device?: HassDevice
+	nodeId?: number
+	devices?: Record<string, HassDevice>
+	remove?: unknown
+}
+
+export interface HassApiAck {
```
AlCalzone: **[Minor · DRY]**

This interface is structurally identical to `MqttApiAck` (`mqttApi.ts:15`) and `ZnifferApiAck` (`znifferApi.ts:23`), and the `{ success: !err, message: err || 'Success X api call', result, api }` construction is copied three times alongside them.

**Why:** `api/socket/types.ts` was created as the shared home for exactly this and holds only `SocketAck`/`noop`, so the ack envelope contract the frontend depends on is now defined in three places.

**Fix:** one generic `ApiAck<T>` plus an ack builder in `types.ts`, parameterised by the api label.

api/socket/znifferApi.ts
```diff
@@ -0,0 +1,104 @@
...
+	runtime: AppRuntime,
+): void {
+	socket.on(
+		inboundEvents.zniffer,
+		async (
+			data: ZnifferApiRequest,
+			cb: SocketAck<ZnifferApiAck> = noop,
+		) => {
+			const apiName: string = data.apiName
+			logger.info(`Zniffer api call: ${apiName}`)
```
AlCalzone: **[Minor · Security]**

This now interpolates the client-supplied `apiName` into the log line. The base logged `data.api` — a field the UI never sends — so the value was always `undefined`.

**Why:** unsanitized CRLF/ANSI in an attacker-controlled string forges log lines in the log file and in the in-UI log viewer. It is newly reachable precisely *because* the refactor fixed the field name. The `hass` and `mqtt` handlers have the same shape and were already reachable.

**Fix:** log the matched switch case, or cap the length and strip control characters.

api/socket/znifferApi.ts
```diff
@@ -0,0 +1,104 @@
...
+							.ensureZniffer()
+							.loadCaptureFromBuffer(buffer)
+						break
+					}
+					default:
+						throw new Error(`Unknown ZNIFFER api ${apiName}`)
+				}
+			} catch (error) {
+				logger.error('Error while calling ZNIFFER api', error)
+				err = getErrorMessage(error)
```
AlCalzone, line 90: **[Minor · Correctness]**

`getErrorMessage(error)` replaces the base's `error.message`, so a thrown non-`Error` now acks `success: false` where it previously left `err` `undefined` and acked `success: true, message: 'Success ZNIFFER api call'`. Same change in the `hass` and `mqtt` handlers.

**Why:** a genuine improvement, but an undeclared behavior change in a PR whose claim is that acknowledgement envelopes are unchanged, and nothing pins it.

**Fix:** call it out in the description, or add a case that throws a string.

api/socket/zwaveApi.ts
```diff
@@ -0,0 +1,89 @@
...
+			let state: Partial<ZwaveState> & { zniffer?: ZnifferStatus } = {}
+
+			const currentGw = runtime.ensureGateway()
+			if (currentGw.zwave) {
+				state = currentGw.zwave.getState()
+			}
+
+			const currentZniffer = runtime.zniffer
+			if (currentZniffer) {
+				state.zniffer = currentZniffer.status()
```
AlCalzone, line 30: **[Minor · Correctness]**

This mutates the object `getState()` returned; the base spread it into a fresh literal first.

**Why:** safe today only because `ZwaveClient.getState()` builds a new literal. If it ever returns a cached or shared object, the client's state silently grows a `zniffer` key.

**Fix:** build the ack from a spread copy.

test/lib/socket/serviceFreshness.test.ts
```diff
@@ -0,0 +1,237 @@
...
+		const currentHarness = await getHarness({ gateway, zniffer })
+		const client = await connectedClient(currentHarness)
+
+		await expect(
+			emit(client, 'ZNIFFER_API', {
+				apiName: 'getFrames',
+			} satisfies ZnifferApiRequest),
+		).resolves.toMatchObject({ result: ['frame-a'] })
+
+		await setSettings(currentHarness, { zniffer: { enabled: false } })
```
AlCalzone: **[Minor · Tests]**

`{ zniffer: { enabled: false } }` is a truthy object, so `AppRuntime.startZniffer`'s `if (settings)` constructs a fresh manager and this test passes through the healthy path.

**Why:** the actual stale path is untested and still broken — with no `zniffer` key at all, `startZniffer` never clears `_zniffer`, so `ensureZniffer()` hands handlers the closed previous manager instead of throwing. The `success: false` assertion here therefore holds for the wrong reason. The bug itself is outside this diff; the test that claims to cover it is inside.

**Fix:** cover settings with no `zniffer` key, and clear the runtime slot when none is started.

test/lib/socket/serviceFreshness.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+async function connectedClient(
+	currentHarness: SocketHarness,
+): Promise<ClientSocket> {
+	const client = currentHarness.createClient()
+	await currentHarness.connectClient(client)
+	return client
+}
+
+function emit<T>(
```
AlCalzone: **[Minor · DRY]**

`emit` and `connectedClient` (line 12) are redefined locally, but `test/lib/socket/helpers.ts` exports both identically and five sibling socket tests import them.

**Why:** divergence in the harness contract the suite shares.

**Fix:** import from `./helpers.ts`; promote `restart` there too if it's generally useful.

api/socket/zwaveApi.ts
```diff
@@ -0,0 +1,89 @@
...
+	api?: string
+}
+
+type DynamicCallApi = (
+	apiName: string,
+	...args: unknown[]
+) => Promise<ZwaveApiAck>
+
+function wireArguments(args: unknown): unknown[] {
+	return [...(args as Iterable<unknown>)]
```
AlCalzone: **[FYI · Security]**

`[...(args as Iterable<unknown>)]` spreads unvalidated wire input, so `{"api":"...","args":{}}` throws a `TypeError` inside an async listener that Socket.IO never awaits — unhandled rejection, and process exit under Node's default.

Pre-existing and reachability is unchanged (the base spread `...data.args` identically), but this `unknown`-to-`Iterable` cast is the natural place to reject non-arrays now that this is a typed boundary.

api/socket/zwaveApi.ts
```diff
@@ -0,0 +1,89 @@
...
+				const result = await callApi(
+					data.api,
+					...wireArguments(data.args),
+				)
+				result.api = data.api
+				cb(result)
+			} else {
+				cb({
+					success: false,
+					message: 'Zwave client not connected',
```
AlCalzone: **[FYI · Design]**

Preserved verbatim from the base, so wire compatibility holds — noting only that it differs from `ZwaveClient.ts:6064`'s `'Z-Wave client not connected'` for the same condition, and both spellings are live on the wire (`inboundApis.test.ts:76` asserts this one, `serviceFreshness.test.ts:140` the other).

api/socket/types.ts
AlCalzone: **[Nit · Design]**

A module named `types.ts` exporting the runtime value `noop`. Related naming drift across the new siblings: `register...ApiHandler` vs `registerSubscriptionHandlers` vs `registerInitHandler`, against the `register...Routes` convention of the sibling `api/routes/*` stack — and `registerInitHandler`, a cross-cutting zwave+zniffer+debug snapshot, lives in `zwaveApi.ts`. Explicitly optional.

AlCalzone (commented):
## Verdict: Ship with minor changes

No blockers. I ran `vitest run test/lib/socket/` at this head: **7 files, 99 tests, all pass.**

**Top 3 risks**
1. 11 `Reflect.apply` sites launder unvalidated wire fields past the type checker, defeating the PR's own "typed protocol modules" premise.
2. `harness.ts` teardown can skip `instance.close()`, leaking servers into every later test in the file.
3. The extraction un-shared channel-validation logic that the base had factored into one helper.

**Strengths**

The wire contract is verified identical against the merge base — event names, ack field names and order, room filtering, argument order, and the `logger.module('App')` label — and the `src/` consumers (`App.vue`, `HomeAssistant.vue`, `ExpandedNode.vue`, `Zniffer.vue`, `InstancesMixin.js`) are unaffected. Auth ordering is intact: `socketManager.authMiddleware` is assigned before `bindServer`, and `io.use(authMiddleware)` is installed there before `registerSocketApi` attaches its `io.on('connection')`, so no handler is reachable pre-auth. `ZwaveClient.callApi`'s `allowedApis` array-membership check is unchanged, so dynamic dispatch gained no new invocation path. The new freshness test pins the properties a lift-and-shift actually breaks: in-flight call surviving a restart, per-call collaborator resolution, first/last-client `setUserCallbacks` accounting. Two silent bug fixes ride along — `znifferApi` logged `data.api` (always `undefined`) where it now logs `apiName`, and non-`Error` throws used to ack `success: true`.

**Themes**

- **Type-laundering.** `Reflect.apply` and the `DynamicCallApi` cast both exist to make optional or `unknown` wire data satisfy a stricter callee.
- **Extraction that un-shared previously-shared code.** `subscriptions.ts` and the three duplicated ack envelopes both regressed from one definition to several.

### Findings outside the diff

Two findings land in `api/lib/SocketManager.ts`, which this PR doesn't touch, so they can't be line-anchored:

- **[Major · Design]** `SocketManager.ts:88` — `_onConnection` still registers a `socket.on` per inbound event and re-emits into `SocketManagerEventCallbacks`, but the only subscriber anywhere in `api/` or `test/` is `'clients'` (`registerSocketApi.ts:28`). Every `ZWAVE_API`/`HASS_API`/`MQTT_API`/`ZNIFFER_API` packet now runs two listeners, one of which only debug-logs, and `SocketManager`'s public typed-event surface advertises a dispatch that can never ack a request. Two registration paths for the same wire events survive a refactor whose purpose was to consolidate them. **Fix:** delete the inbound half of the typed-event surface here — this is the PR that takes ownership — or leave an explicit marker if a later stack entry claims it.
- **[Minor · Readability]** `SocketManager.ts:87` — the comment still says subscribe/unsubscribe are "handled directly in app.ts"; they live in `api/socket/subscriptions.ts` now.

### Coverage

Specialists run: Correctness, Security, DRY & Codebase Fit, Design/API/BackCompat, Tests, Operability, Readability. Clean on Operability — all four `logger` call sites moved verbatim, same module name, every path still calls `cb(...)`. Not run: Performance (no files in its lane).

One Tests finding (a claimed `toStrictEqual` mismatch at `serviceFreshness.test.ts:136`) was refuted on verification: the not-connected ack comes from `ZwaveClient.callApi`, which sets `args` and returns `'Z-Wave client not connected'`, after which the handler sets `.api`. The suite passing confirms it.

api/socket/zwaveApi.ts
AlCalzone: lets address this

api/socket/zwaveApi.ts
AlCalzone: Z-Wave is the correct spelling

api/socket/hassApi.ts
AlCalzone: Fixed in 88f9caf5 — removed the redundant bind and Reflect.apply in favor of direct typed collaborator calls.

api/socket/mqttApi.ts
AlCalzone: Fixed in 88f9caf5 — removed the MQTT bind/Reflect.apply sites and now call the typed gateway methods directly.

api/socket/types.ts
AlCalzone: Fixed in 88f9caf5 — removed noop, made acknowledgements explicitly optional, and moved the shared runtime ACK helpers into api.ts.

api/socket/hassApi.ts
AlCalzone: Fixed in 88f9caf5 — all 11 Reflect.apply sites are gone; discriminated request types now provide the concrete arguments for direct calls.

test/lib/socket/harness.ts
AlCalzone: Fixed in 0e694fdc — closeInstance now runs instance.close() in a finally block while preserving the disconnect timeout failure.

api/socket/subscriptions.ts
AlCalzone: Fixed in 88f9caf5 — subscribe and unsubscribe are thin registrations over one shared channel resolver/applier.

api/socket/zwaveApi.ts
AlCalzone: Fixed in 88f9caf5 — the ACK now uses CallAPIResult<AllowedApis> intersected with the echoed api field.

api/socket/hassApi.ts
AlCalzone: Fixed in 7cab661a — unknown HASS actions are error-logged with a sanitized operation name while the raw wire error remains unchanged.

api/socket/hassApi.ts
AlCalzone: Fixed in 88f9caf5 — HASS, MQTT, and Zniffer now share ApiAck<T> and createApiAck().

api/socket/znifferApi.ts
AlCalzone: Fixed in 88f9caf5 and 7cab661a — operation names are control-character stripped and length-capped in both info and unknown-action error logs.

api/socket/znifferApi.ts
AlCalzone: Fixed in 88f9caf5 — retained getErrorMessage and added real Socket.IO regressions for string rejection values in HASS and Zniffer.

api/socket/zwaveApi.ts
AlCalzone: Fixed in 88f9caf5 — INITED now spreads getState() into a fresh object; the regression asserts the source object is untouched.

test/lib/socket/serviceFreshness.test.ts
AlCalzone: Fixed in 88f9caf5 — startZniffer(undefined) clears the stale runtime slot, and the restart test now removes the zniffer setting entirely.

test/lib/socket/serviceFreshness.test.ts
AlCalzone: Fixed in 88f9caf5 — serviceFreshness now imports emit and connectedClient from the shared socket helpers.

api/socket/zwaveApi.ts
AlCalzone: Fixed in 88f9caf5 — truthy non-array args now receive a failure ACK instead of throwing from the async listener; falsy values still default to an empty array.

api/socket/zwaveApi.ts
AlCalzone: Fixed in 88f9caf5 — the handler now returns “Z-Wave client not connected,” with the wire test updated accordingly.

AlCalzone:
Review follow-up: I preserved SocketManager’s generic first-argument inbound relay because it is an explicit compatibility contract for this layer. It has no protocol consumers today, does not invoke a second protocol handler, and cannot send a second ACK. The stale subscription-ownership comment was corrected in 88f9caf5.

api/socket/api.ts
```diff
@@ -0,0 +1,35 @@
...
+	}
+}
+
+export function safeOperationName(value: unknown): string {
+	if (typeof value !== 'string') return '<invalid>'
+
+	let sanitized = ''
+	for (const character of value) {
+		const code = character.charCodeAt(0)
+		if ((code >= 32 && code < 127) || code > 159) {
```
AlCalzone, line 31: This could use a short comment mentioning which characters 32, 127 and 159 are.

api/socket/api.ts
```diff
@@ -0,0 +1,35 @@
...
+			sanitized += character
+		}
+	}
+	return sanitized.slice(0, 100)
```
AlCalzone, line 36: and another one why we're slicing to 100

api/socket/registerSocketApi.ts
```diff
@@ -0,0 +1,36 @@
...
+		registerZwaveApiHandler(socket, runtime)
+		registerMqttApiHandler(socket, runtime)
+		registerHassApiHandler(socket, runtime)
+		registerZnifferApiHandler(socket, runtime)
+		registerSubscriptionHandlers(socket)
+	})
+
+	socketManager.on('clients', (event, activeSockets) => {
+		const currentGw = runtime.ensureGateway()
+		if (event === 'connection' && activeSockets.size === 1) {
```
AlCalzone, line 31: Add a comment saying that we set the callbacks for the first client, and remove them after the last one is disconnected.

api/socket/znifferApi.ts
```diff
@@ -0,0 +1,92 @@
...
+	| {
+			apiName:
+				| 'start'
+				| 'stop'
+				| 'clear'
+				| 'getFrames'
+				| 'saveCaptureToFile'
+	  }
+	| { apiName: 'setFrequency'; frequency: number }
+	| { apiName: 'setLRChannelConfig'; channelConfig: number }
```
AlCalzone: I think the channelConfig should be the corresponding zwave-js enum, not `number`

api/socket/hassApi.ts
```diff
@@ -0,0 +1,111 @@
...
+	  }
+	| {
+			apiName: 'rediscoverNode' | 'disableDiscovery'
+			nodeId: number
+	  }
+	| {
+			apiName: 'store'
+			devices: Record<string, HassDevice>
+			nodeId: number
+			remove: unknown
```
AlCalzone: `unknown` seems weird. Check the correct type.

test/lib/socket/serviceFreshness.test.ts
```diff
@@ -0,0 +1,231 @@
...
+			success: boolean
+			result: unknown
+		}>(client, 'ZNIFFER_API', {
+			apiName: 'getFrames',
+		} satisfies ZnifferApiRequest)
+		expect(replacementResult.success).toBe(false)
+		expect(zniffer.getFrames).toHaveBeenCalledOnce()
+	})
+
+	it('accepts a stop request without a buffer', async () => {
```
AlCalzone: this test title is weird. why would we accept a stop request with a buffer anyways?

test/lib/socket/serviceFreshness.test.ts
```diff
@@ -0,0 +1,231 @@
...
+		const client = await connectedClient(currentHarness)
+
+		await emit(client, 'ZNIFFER_API', {
+			apiName: 'stop',
+		} satisfies ZnifferApiRequest)
+
+		expect(zniffer.stop).toHaveBeenCalledOnce()
+	})
+
+	it('accepts a clear request without a buffer', async () => {
```
AlCalzone: dito

test/lib/socket/serviceFreshness.test.ts
```diff
@@ -0,0 +1,231 @@
...
+
+	it('processes a request without an acknowledgement callback', async () => {
+		const gateway = createFakeGateway()
+		const currentHarness = await getHarness({ gateway })
+		const client = await connectedClient(currentHarness)
+
+		client.emit('ZWAVE_API', { api: 'fireAndForget' })
+		// Socket.IO guarantees ordering, so the acknowledged call flushes the earlier request
+		await emit(client, 'ZWAVE_API', { api: 'barrier' })
+
```
AlCalzone: What are these two APIs? I hope they only exist at test time, not at runtime.

api/socket/api.ts
AlCalzone: Fixed in 89f78ffa — the filter now identifies C0, DEL, and C1 control ranges.

api/socket/api.ts
AlCalzone: Fixed in 89f78ffa — the comment now explains that the 100-character cap bounds client-controlled log fields.

api/socket/registerSocketApi.ts
AlCalzone: Fixed in 89f78ffa — documented first-client callback installation and last-client callback removal.

api/socket/znifferApi.ts
AlCalzone: Fixed in 89f78ffa — channelConfig now uses ZnifferLRChannelConfig through the socket request and Zniffer manager.

api/socket/hassApi.ts
AlCalzone: Fixed in 89f78ffa — remove is now boolean, matching the UI and storeDevices semantics.

test/lib/socket/serviceFreshness.test.ts
AlCalzone: Fixed in 89f78ffa — renamed the test to describe stopping the current Zniffer capture.

test/lib/socket/serviceFreshness.test.ts
AlCalzone: Fixed in 89f78ffa — renamed the test to describe clearing captured Zniffer frames.

test/lib/socket/serviceFreshness.test.ts
AlCalzone: Fixed in 89f78ffa — replaced the fake fireAndForget/barrier names with the real _getScenes API; the names existed only in the fake collaborator.

robertsLando (approved):

Issue #3088, opened 2023-05-15
Simon IO 100 Roller Blind is not working as Roller Blind.


davidcalleja:
### Version

Build/Run method:

[X] Docker
[ ] Git Clone + yarn install + yarn run build + yarn start
[ ] Pkg

Z-Wave JS UI version: 8.16.6
Z-Wave JS version: 10.20.0

### Describe the bug

[After add a new device](https://github.com/zwave-js/node-zwave-js/pull/5743) (a Simon IO 100 Roller Blind) is not detecting as a roller blind because in Home Assistant section it says this: 

![Screenshot 2023-05-15 at 11 45 31](https://github.com/zwave-js/zwave-js-ui/assets/5727085/ef4b2936-ce57-4146-a2b5-5f1faba9cf91)

robertsLando:
I misread the issue, I thought it was missing the node values of that device.

@davidcalleja Could you please tell me what controls you are missing from that device? Could you tell me what are the device valueId/command classes that you are missing in MQTT discovery? 

BTW If you are using home assistant I really suggest you to use zwave-js-integration that's the official way to work with zwave-js, in order to use it simply install that on home assistant and in zwave-js-ui enable WS Server in settings:

![2023-05-15_15-46](https://github.com/zwave-js/zwave-js-ui/assets/11502495/f23e1125-4575-48ed-bc99-ed5a3361868b)

davidcalleja:
I am using Home Assistant but not directly, MQTT is the connector between because are not in same device.

This device is another "Simon IO 100 Roller Blind" with different ID. It must be same as this another roller blind I have because is same with different ID.

![Screenshot 2023-05-15 at 15 54 15](https://github.com/zwave-js/zwave-js-ui/assets/5727085/3a25319e-ab86-4049-80c6-3358d8bd2d7e)

robertsLando:
Them are not the same device, one is 615-0-258 the other is 615-0-4, do you see command classes to control the cover position in node tab?

robertsLando:
> I am using Home Assistant but not directly, MQTT is the connector between because are not in same device.

You can use the integration anyway, no need to have HA in the same device, just need HA to be able to point to the ip address to the device that is running Z-Wave JS UI. I really suggest you to give it a try

Anyway I should give a look at application logs, maybe the device is not recognized as a roller blind for some reason

davidcalleja:
Yes but I'm using it not only with HA, I'm sending this events not others ;)

[...]

PR #4729, opened 2026-07-11
refactor(hass): extract discovery domain

AlCalzone:
Extracts the Home Assistant discovery domain from `Gateway`.

- Separates entity generation, custom-device loading, persistence, and watcher ownership.
- Keeps discovery payloads, identifiers, templates, command mappings, and synchronous projection compatible.
- Review focus: cache/watcher lifecycle, stale publication fencing, and exact MQTT discovery output.

Part of #4722. Stacked on #4728.


api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+	private readonly logger: HassLogger
+
+	public constructor(options: DiscoveryGeneratorOptions) {
+		this.config = options.config
+		this.mqtt = options.mqtt
+		this.zwave = options.zwave
+		this.topics = options.topics
+		this.registry = options.registry
+		this.state = options.state
+		this.logger = options.logger
```
robertsLando, line 102: we could think about using a dedicated logger for modules like this

api/lib/Gateway.ts
```diff
@@ -290,16 +193,69 @@ export default class Gateway {
...
+		customDeviceRegistry: HassDeviceRegistryLifecyclePort,
 	) {
 		this.config = config || { type: 1 }
 		// clients
 		this._mqtt = mqtt
 		this._zwave = zwave
+		this.customDeviceRegistry = customDeviceRegistry
+		const getDiscovered = () => this.discovered
+		const getMqtt = () => this._mqtt
+		const getZwave = () => this._zwave
```
robertsLando: not sure I understand why this getters are needed here, looks strange

api/lib/Gateway.ts
```diff
@@ -290,16 +193,69 @@ export default class Gateway {
...
+		this.discoveryGenerator = new DiscoveryGenerator({
+			config: this.config,
+			mqtt: {
```
robertsLando: any reason apart from limit the module capabilities (security) we don't pass the instance down here?

api/lib/ZwaveClient.ts
```diff
@@ -1243,72 +1241,26 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
-				const id = hassDevice.id
-				delete hassDevice.id
-				node.hassDevices[id] = hassDevice
-			}
-
-			this.emitNodeUpdate(node, {
-				hassDevices: node.hassDevices,
-			})
-		}
+		this.hassDeviceStore.updateDevice(hassDevice, nodeId, deleteDevice)
```
robertsLando, line 1247: instead of having methods like this why not make the callers that call this methods directly use the hassDeviceStore instead? Call zwave.hassStore.updateDevice instead making hassStore a getter. I think it makes cleaner that you are calling a method related to hass as calling updateDevice on zwaveclient is misleading. same applies to gateway

api/lib/Gateway.ts
```diff
@@ -457,49 +391,8 @@ export default class Gateway {
...
-									property: 'Up',
-								},
-								false,
-							)
-							.catch(() => {})
-						return null
-					}
-				}
-			}
+			payload = this.discoveryGenerator.transformPayload(payload, valueId)
```
robertsLando, line 376: from a logic prospective not sure if parsepayload really belongs to discovery generator, maybe only part of it

api/hass/DiscoveryGenerator.ts
AlCalzone: Fixed in 807e73cc — DiscoveryGenerator keeps its injected logger port, and Gateway now supplies a dedicated HASS module logger.

api/lib/Gateway.ts
AlCalzone: These closures construct narrow adapters without binding the adapter object as Gateway. They keep collaborator access behind the HASS ports rather than exposing Gateway internals.

api/lib/ZwaveClient.ts
AlCalzone: These methods remain compatibility facades because runtime and socket consumers depend on the ZwaveClient public contract. Exposing hassDeviceStore would leak the implementation and expand this migration beyond the discovery extraction.

api/lib/Gateway.ts
AlCalzone: Passing the full instances would couple discovery back to the concrete MQTT and Z-Wave classes. The narrow capability ports are intentional for ownership boundaries and isolated testing, rather than as a security measure.

api/lib/ZwaveClient.ts
```diff
@@ -913,6 +897,20 @@ class ZwaveClient extends TypedEventEmitter<ZwaveClientEventCallbacks> {
...
 		this.cfg = config
 		this.socket = socket
+		this.hassDeviceStore = new HassDeviceStore({
+			hasNode: (nodeId) => this._nodes.has(nodeId),
+			getNodeDevices: (nodeId) => this._nodes.get(nodeId)?.hassDevices,
+			setNodeDevices: (nodeId, devices) => {
+				const node = this._nodes.get(nodeId)
+				if (node) node.hassDevices = devices
+			},
+			getStoredNode: (nodeId) => this.storeNodes[nodeId],
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 0b47ffb6 — absent per-node persistence records are now initialized through the typed store port, while malformed values remain rejected.

api/lib/Gateway.ts
AlCalzone: parsePayload itself remains in Gateway for generic MQTT coercion. Only the HASS-specific thermostat mapping and cover STOP transformation is delegated to DiscoveryGenerator.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+			this.logger.error(`Failed to load ${loaded}:`, error)
+			return
+		}
+
+		try {
+			const sha = crypto
+				.createHash('sha256')
+				.update(JSON.stringify(devices))
+				.digest('hex')
+			if (this.lastCustomDevicesLoad === sha) return
```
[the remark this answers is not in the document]
AlCalzone: Fixed in b675b991 — the registry now hashes the loaded file bytes, so JavaScript function and undefined-value edits are detected.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+				case CommandClasses.Time:
+				case CommandClasses['Energy Production']:
+				case CommandClasses.Battery: {
+					let sensor: SensorDefinition | null = null
+					let isSensor = true
+					if (commandClass === CommandClasses['Multilevel Sensor']) {
+						// Skip auxiliary values because zwave-js omits their sensor metadata
+						if (!valueId.ccSpecific) return
+						const sensorType = valueId.ccSpecific.sensorType
+						if (typeof sensorType !== 'number') return
```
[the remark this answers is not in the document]
AlCalzone: > entities discovered under the old payload orphan on upgrade

We MUST avoid this during this refactor. However, zwave-js guarantees `ccSpecific.sensorType` to not be undefined, so this might be false positive.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+			this.topics.valueTopic(node, currentColorValue, true),
+		)
+		const endpoint = currentColorValue.endpoint
+		const supportedColors: ColorMode[] = ['rgb']
+		config.discovery_payload.supported_color_modes = supportedColors
+		config.values = []
+		config.discovery_payload.rgb_state_topic = this.mqtt.getTopic(
+			currentColorTopics.topic,
+		)
+		if (currentColorTopics.targetTopic) {
```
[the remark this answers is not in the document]
AlCalzone: Not sure how exactly this flows into Z-Wave JS UI, but I'm pretty sure the targetValue CCValue always exists.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+			}
+
+			const modeId = modes[0]
+			if (modeId) {
+				config.values.push(modeId)
+				const mode = node.values[modeId]
+				config.discovery_payload.mode_state_topic = modeId
+				config.discovery_payload.mode_command_topic = modeId + '/set'
+
+				const availableModes = (mode.states ?? [])
```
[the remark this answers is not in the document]
AlCalzone: Claims like `Thermostat Mode value with no states` should be double-checked against the zwave-js codebase. This particular case cannot exist.

api/hass/ports.ts
```diff
@@ -0,0 +1,142 @@
...
+	readonly homeHex: string | undefined
+	getNode(nodeId: number): unknown
+	getNodes(): Iterable<readonly [number, unknown]>
+	updateDevice(
+		device: HassDevice,
+		nodeId: number,
+		deleteDevice?: boolean,
+	): void
+	emitNodeUpdate(nodeId: number, devices: HassDeviceMap): void
+	writeCoverStop(value: HassValue): Promise<unknown>
```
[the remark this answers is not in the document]
AlCalzone: Double check that this proposed fix does not change existing behavior for the worse.

api/lib/ZwaveClient.ts
```diff
@@ -157,6 +157,14 @@ import { PkgFsBindings } from './PkgFsBindings.ts'
 import { regionSupportsAutoPowerlevel } from './shared.ts'
 import { deviceConfigPriorityDir } from './Constants.ts'
 import { createRequire } from 'node:module'
+import { HassDeviceStore } from '../hass/DeviceStore.ts'
+import type {
+	HassDevice,
+	HassDeviceMap,
+	StoreHassDevicesResult,
+} from '../hass/types.ts'
```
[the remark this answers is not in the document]
AlCalzone: How about we don't deprecate things that are only used inside this project anyways and just do the change?

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+						].includes(specificDeviceClass) ||
+						node.deviceId === '615-0-258'
+					) {
+						config = utils.copy(hassCfg.cover_position)
+						config.discovery_payload.command_topic = setTopic
+						config.discovery_payload.position_topic = getTopic
+						config.discovery_payload.set_position_topic = setTopic
+						config.discovery_payload.position_template =
+							'{{ value_json.value | round(0) }}'
+						config.discovery_payload.position_open = 99
```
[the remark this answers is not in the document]
AlCalzone: do note that this is a HASS convention, Z-Wave actually has different positions for a subset of covers (tilt-only)

api/hass/DiscoveryGenerator.ts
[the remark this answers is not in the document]
AlCalzone: Confirmed against zwave-js: `ThermostatModeCCValues.thermostatMode.meta` always defines `states`, and supported-mode reports only replace that map with a narrower one. This case cannot occur for a real Thermostat Mode value, so no code change is needed.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+
+		const hassId = source ? source.type + '_' + source.object_id : null
+
+		try {
+			if (!hassId || node.hassDevices[hassId]) return
+
+			const hassDevice = utils.copy(source)
+			const payload = hassDevice.discovery_payload
+			const values = hassDevice.values
+			if (!values) throw new TypeError('HASS device has no values')
```
AlCalzone: **[Major] · Correctness**

The `if (!values) throw` guard sits above the `type === 'climate'` branch.

**Why:** Verified against the base revision: the climate path never reads `hassDevice.values`. A `customDevices.js` climate entry without `values` previously still landed in `node.hassDevices[hassId]` and stayed visible and editable in the UI, with publication failing separately inside `publishDiscovery`. It now aborts before the assignment and the device disappears from the node entirely.

**Fix:** Require `values` only in the branch that actually consumes it, matching the pre-move reachability.

api/lib/Gateway.ts
```diff
@@ -253,24 +111,70 @@ interface ValueIdTopic {
...
 	private config: GatewayConfig
-	private _mqtt?: MqttClient
-	private _zwave?: ZwaveClient
-	private topicValues: { [key: string]: ZUIValueId }
-	private discovered: { [key: string]: HassDevice }
-	private topicLevels: number[]
-	private _closed: boolean
+	private _mqtt?: TMqtt
+	private _zwave?: TZwave
+	private topicValues: { [key: string]: ZUIValueId } = {}
```
AlCalzone, line 157: **[Major] · Design**

`discovered` has two owners: `Gateway` declares the field and reassigns it in `start()` but never reads it, while `DiscoveryGenerator` performs all reads and writes through the `state.discovered` getter closure.

**Why:** The refactor's stated goal is ownership, yet the domain's core index stays externally reachable and replaceable. The getter indirection exists only to survive Gateway's reassignment, so any future Gateway-side read re-couples the two. This is also what makes the PR's own stated review focus — stale-publication fencing — hard to verify by inspection.

**Fix:** Move the map into `DiscoveryGenerator`, expose a `reset()`, and drop `HassDiscoveryState`.

api/runtime/AppRuntime.ts
```diff
@@ -32,11 +31,21 @@ export interface ManagedService {
 	close(): Promise<void>
 }
 
+export interface GatewayFactoryPort {
```
AlCalzone: **[Major] · DRY / Codebase fit**

`GatewayFactoryPort` is declared inline here and typed with the concrete `ZWaveClient` / `MqttClient` classes, while every sibling port in this refactor series (`GatewayPort`, `ZnifferPort`, `ZwaveClientPort`, `MqttClientPort`) lives in `api/runtime/ports.ts` as a structural `Pick<>`.

**Why:** It breaks the port-location and port-style convention the series established, and re-imports concrete client classes into the runtime layer that `ports.ts` exists to decouple.

**Fix:** Move it to `api/runtime/ports.ts` and express the params via the existing `ZwaveClientPort` / `MqttClientPort` aliases.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+					}
+
+					if (!sensor) return
+					if (isSensor) config = utils.copy(hassCfg.sensor_generic)
+					if (!config) return
+					config.object_id = utils.joinProps(
+						sensor.sensor,
+						sensor.objectId,
+					)
+					let unit = valueId.unit ?? valueUnit(valueId.value)
```
AlCalzone: **[Minor] · Correctness**

`valueId.unit ?? valueUnit(valueId.value)` replaces the base truthiness chain (`if (valueId.unit) ... else if (valueId.value?.unit)`).

**Why:** Verified against base: `unit === ''` now short-circuits and suppresses `unit_of_measurement` entirely, where base fell through to `valueId.value.unit`. Empty-string units do occur in zwave-js metadata, and the result is a silently missing unit in the HA discovery payload.

**Fix:** Keep `||` semantics so the empty string falls through.

api/lib/Gateway.ts
```diff
@@ -457,49 +392,8 @@ export default class Gateway {
 				payload = valueId.value > 0 ? 0 : 0xff
 			}
```
AlCalzone, line 375: **[Minor] · Correctness**

`transformPayload`'s `null` return is now the sole abort sentinel, so any incoming `null` payload cancels the write.

**Why:** Base returned `null` only inside the `type === 'cover' && property === 'targetValue' && payload === payload_stop` branch. A literal `null` on a `/set` topic for an undiscovered value used to reach `writeValue`; it is now silently dropped instead of producing a driver-level rejection.

**Fix:** Use a distinct sentinel or a `{ handled }` result rather than overloading `null`.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+		)
+		const endpoint = currentColorValue.endpoint
+		const supportedColors: ColorMode[] = ['rgb']
+		config.discovery_payload.supported_color_modes = supportedColors
+		config.values = []
+		config.discovery_payload.rgb_state_topic = this.mqtt.getTopic(
+			currentColorTopics.topic,
+		)
+		if (currentColorTopics.targetTopic) {
+			config.discovery_payload.rgb_command_topic = this.mqtt.getTopic(
```
AlCalzone, line 1429: **[Minor] · Correctness**

`rgb_command_topic`, `brightness_command_topic` and `command_topic` became conditional on `targetTopic` being present, where base always assigned `getTopic(targetTopic, true)`.

**Why:** The emitted discovery payload loses those keys (and the derived `color_temp_command_topic` becomes `undefined`) instead of carrying the previous value, so retained discovery configs change shape for any RGB node whose target topic does not resolve.

**Fix:** Confirm the shape change is intended and pin it with a test asserting the exact payload.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+					break
+				}
+				case CommandClasses['Alarm Sensor']: {
+					if (valueId.property !== 'state') return
+					config = this.getBinarySensorConfig(
+						Constants.deviceClass.sensor_binary.PROBLEM,
+					)
+					const alarmType =
+						typeof valueId.propertyKey === 'number'
+							? AlarmSensorType[valueId.propertyKey]
```
AlCalzone: **[Minor] · Correctness**

`AlarmSensorType[valueId.propertyKey]` is now gated on `typeof propertyKey === 'number'`; base also hit the enum's reverse string mapping.

**Why:** A string `propertyKey` used to append the numeric enum value to `object_id`. Low likelihood, but it silently changes an entity id.

AlCalzone (commented):
## Self-review via `/deep-review`

**Verdict:** Ship with minor changes — no confirmed blockers. The seam is defensible and the discovery contract verified byte-identical.

**Top 3 risks**
1. `storeDevices` became result-returning and one caller still ignores it, turning a thrown import failure into a false success.
2. A `values` guard hoisted above the climate branch drops climate custom-devices that previously registered.
3. `??`-for-`||` substitution during the move changes empty-string unit handling.

**Strengths.** The module-global `allDevices` plus the never-closable module-scope `fs.watch` is now owned behind `start()` / `dispose()`, with the lifecycle closed at both ends (`Gateway.close()` disposes its fork, `AppRuntime.destroy()` disposes the factory). That alone justifies the PR. The discovery contract was verified unchanged: `unique_id`, `identifiers`, `discoveryTopic`, the availability block, RAW templating and the persisted `hassDevices` shape are byte-identical. Every assertion in the deleted `test/lib/Gateway.test.ts` has a replacement in the new per-unit suites. Log lines survived the move verbatim under the same `'Gateway'` module logger.

**Theme.** Nearly every real finding is *incidental modernization during a code move* — `??` swapped for `||`, guards hoisted above branches they did not previously gate, truthiness conditions added to unconditional assignments. A strict move-only discipline (identical operators, guards left in place) would have prevented all of them and made the diff reviewable by inspection. Worth carrying into the remaining #4722 extractions.

### Findings that could not be anchored to a diff line

- **Major · `api/routes/importExport.ts:117`** — the changed `storeDevices` now returns `StoreHassDevicesResult` and this caller discards it. Verified against base: when `storeNodes[nodeId]` is absent, base threw a `TypeError` that the route's `catch` turned into `success: false`. It now returns `invalid-stored-node`, `hassDevices` are never persisted, and the route still replies `"Configuration imported successfully"` — a silent partial import. `api/socket/hassApi.ts` handles the result correctly, so only this caller lags. Fix: make the result non-ignorable at every call site, or split a throwing variant for server-side callers.
- **Minor · `api/lib/Gateway.ts:107`** — `interface ValueIdTopic` survives and is used only as a single cast, duplicating `HassValueTopic`, which is already the declared return type of `valueTopic`. The cast also discards the new port typing.
- **FYI · `api/socket/hassApi.ts:57`** — predates this PR, faithfully moved: the `discover` / `delete` handlers pass a fully client-supplied `HassDevice` to `publishDiscovery`, and `hassDevice.discoveryTopic` is used verbatim as the MQTT topic with client-supplied `discovery_payload` as the body. An authenticated socket client can publish arbitrary payloads to arbitrary topics under `discoveryPrefix`. This PR actually narrows it slightly by adding the `discoveryTopic` truthiness guard the old code lacked.
- **FYI** — there is no epoch/generation fence and no `_closed` check in `publishDiscovery`. Unchanged from base, so not a regression, but it means stale-publication fencing rests on tests rather than on construction.

### Coverage

Eight specialist lanes ran: correctness, security, performance, DRY/codebase-fit, design/API/back-compat, tests, operability, readability. Clean on tests, operability and readability. Review ran against a dedicated worktree at PR head `58a17a11`. No prior unresolved review threads.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+
+	private load(): void {
+		let loaded = ''
+		let source: Buffer
+		let devices: unknown
+
+		try {
+			if (fs.existsSync(this.customDevicesJsPath)) {
+				loaded = this.customDevicesJsPath
+				source = fs.readFileSync(loaded)
```
AlCalzone, line 106: **[Minor] · Security**

Evicting `require.cache` before `require(modulePath)` makes every write to `<store>/customDevices.js` re-execute that file in the app process, where previously the cached module was reused after the first load.

**Why:** `PUT /api/store` (`api/routes/store.ts`, `isAuthenticated` only) lets any logged-in user write arbitrary content under `storeDir`, so the fs.watch → require path is a file-write-to-code-exec chain. The boundary was already crossable once before this PR (a not-yet-required `customDevices.js` executed on the first watch event), so this converts one-shot execution into repeatable execution rather than opening a new hole.

**Fix:** Gate `.js` catalog loading behind an explicit opt-in and prefer the `.json` path by default, or document that store-dir write access equals code execution.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+		})
+	}
+
+	public dispose(): void {
+		this.unsubscribeSource?.()
+		this.unsubscribeSource = undefined
+		for (const watcher of this.watchers.values()) watcher.close()
+		this.watchers.clear()
+		this.listeners.clear()
+		this.started = false
```
AlCalzone, line 172: **[Minor] · Correctness**

`dispose()` clears watchers and listeners and resets `started`, but retains `lastCustomDevicesLoad`.

**Why:** A start → dispose → start cycle whose file content is unchanged skips `notify()`, so `dispose()` is not a true reset. Harmless today only because forks re-project in their own `start()`.

**Fix:** Clear the sha in `dispose()`.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+
+export class CustomDeviceRegistry implements HassDeviceRegistryPort {
+	private readonly baseDevices: HassDeviceCatalogSource
+	private readonly logger: Pick<HassLogger, 'error' | 'info'>
+	private readonly customDevicesPath: string
+	private readonly customDevicesJsPath: string
+	private readonly customDevicesJsonPath: string
+	private readonly watchers = new Map<string, fs.FSWatcher>()
+	private readonly listeners = new Set<() => void>()
+	private readonly source: CustomDeviceRegistry | undefined
```
AlCalzone, line 59: **[Minor] · Design**

One class plays both roles — watching owner and subscribed fork — branching on `source` throughout the constructor, `start()` and `dispose()`.

**Why:** Forks carry unused `load`, `watch` and path fields, and `fork()` on a fork silently chains. The `started` / `lastCustomDevicesLoad` interactions after `dispose()` + re-`create()` have to be traced rather than being evident from the type.

**Fix:** Split the read-only projection into its own type implementing `HassDeviceRegistryLifecyclePort`.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+		this.listeners.clear()
+		this.started = false
+	}
+
+	private copyProjection(): HassDeviceCatalogSource {
+		return copyCatalogProjection(this.allDevices)
+	}
+
+	private subscribe(listener: () => void): () => void {
+		this.listeners.add(listener)
```
AlCalzone, line 180: **[Minor] · DRY / Codebase fit**

Hand-rolled `listeners` / `subscribe` / `notify`.

**Why:** The repo already has `TypedEventEmitter` (`api/lib/EventEmitter.ts`), used by `SocketManager`, `ZnifferManager`, `MqttClient` and `ZwaveClient`. A fifth ad-hoc pub/sub is a pattern deviation.

**Fix:** Extend `TypedEventEmitter<{ change: () => void }>` and drop the bespoke set.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+	public start(): void {
+		if (this.started) return
+
+		this.started = true
+		if (this.source) {
+			this.allDevices = this.source.copyProjection()
+			this.unsubscribeSource = this.source.subscribe(() => {
+				this.allDevices = this.source?.copyProjection() ?? {}
+			})
+			return
```
AlCalzone, line 89: **[Minor] · Performance**

`start()` recomputes `copyProjection()` from the source although the constructor already did the same clone for forked instances.

**Why:** Doubles the per-restart clone cost for no observable benefit. Small in absolute terms — the catalog is bounded and this runs on settings save / restart, not on a hot path.

**Fix:** Skip the redundant assignment when the constructor already seeded `allDevices` from the same source.

api/hass/ports.ts
```diff
@@ -0,0 +1,142 @@
...
+	getStoredNode(nodeId: number): unknown
+	setStoredNode(nodeId: number, node: HassPersistenceNode): void
+	emitNodeUpdate(nodeId: number, devices: HassDeviceMap): void
+	updateStoreNodes(): Promise<void>
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+	return value !== null && typeof value === 'object' && !Array.isArray(value)
+}
+
```
AlCalzone: **[Minor] · Design**

`isHassNode` checks 4 fields but asserts 12, plus `Required<hassDevices | values>`.

**Why:** Its six call sites in `Gateway.ts` turn a guard failure into a silent no-op where the old code ran unconditionally. Confirmed benign for every node `ZwaveClient` actually builds (`ready`, `values`, `hassDevices` are all initialized) — only `fakeNodes.json` entries omitting `ready` would newly skip discovery — but the silence is the hazard.

**Fix:** Log on rejection, or narrow `HassNode` to what the guard actually proves.

api/hass/ports.ts
```diff
@@ -0,0 +1,142 @@
...
+export interface HassDeviceStorePort {
+	hasNode(nodeId: number): boolean
+	getNodeDevices(nodeId: number): HassDeviceMap | undefined
+	setNodeDevices(nodeId: number, devices: HassDeviceMap): void
+	getStoredNode(nodeId: number): unknown
+	setStoredNode(nodeId: number, node: HassPersistenceNode): void
+	emitNodeUpdate(nodeId: number, devices: HassDeviceMap): void
+	updateStoreNodes(): Promise<void>
+}
+
```
AlCalzone, line 98: **[Minor] · DRY / Codebase fit**

A private `isRecord` is redefined here.

**Why:** `utils.isRecord` (`api/lib/utils.ts`) is the repo helper and is already imported by the two neighbouring new files, `DeviceStore.ts` and `DiscoveryGenerator.ts`.

**Fix:** Reuse `utils.isRecord`.

api/hass/ports.ts
```diff
@@ -0,0 +1,142 @@
...
+> &
+	Required<Pick<ZUINode, 'hassDevices' | 'values'>>
+export type HassTopicNode = Pick<
+	ZUINode,
+	'deviceId' | 'id' | 'loc' | 'name' | 'values'
+>
+
+export interface HassValueConfiguration {
+	device_class?: string
+	icon?: string
```
AlCalzone, line 32: **[Minor] · DRY / Codebase fit**

`HassValueConfiguration` and `HassDiscoveryConfig` hand-restate field subsets of `GatewayValue` / `GatewayConfig`, whereas every other alias in this file and in `runtime/ports.ts` derives via `Pick<>`.

**Why:** A hand-restated subset drifts silently when the gateway config gains or renames a field; a `Pick<>` breaks the build instead.

**Fix:** Derive with `Pick<GatewayConfig, …>` / `Pick<GatewayValue, …>`.

api/runtime/GatewayFactory.ts
```diff
@@ -0,0 +1,36 @@
+import { CustomDeviceRegistry } from './CustomDeviceRegistry.ts'
+import type { HassLogger } from './ports.ts'
+import type { HassDeviceCatalog } from './types.ts'
+import Gateway from '../lib/Gateway.ts'
```
AlCalzone, line 4: **[Minor] · Design**

The hass module imports and constructs `Gateway` from `api/lib`, and `AppRuntime` now reaches into hass to obtain a gateway.

**Why:** The direction is inverted — hass is a leaf domain everywhere else, and here the extracted domain owns construction of the god-object it was extracted from. The factory's real job is owning a `CustomDeviceRegistry` lifecycle.

**Fix:** Put the factory next to `GatewayFactoryPort` in `api/runtime`, keeping only the registry in `api/hass`.

api/lib/Gateway.ts
```diff
@@ -1,192 +1,50 @@
...
+	HassDeviceRegistryLifecyclePort,
+	HassNode,
+	HassTopicNode,
+	HassValue,
+	HassValueTopic,
+} from '../hass/ports.ts'
+import { isHassNode } from '../hass/ports.ts'
 
 const logger = module('Gateway')
+const hassLogger = module('HASS')
```
AlCalzone, line 38: **[Minor] · DRY / Codebase fit**

`NODE_PREFIX = 'nodeID_'` stays here while the extracted `HASS_NODE_PREFIX` (`api/hass/types.ts`) holds the same literal. Both are live — topic building vs. discovery naming.

**Why:** Two copies of a wire-visible constant that must stay in sync. Flagged independently by two review lanes.

**Fix:** Import the exported constant.

api/lib/ZwaveClient.ts
AlCalzone, line 165: **[Minor] · Design**

`HassDevice` is re-exported for back-compat with no `@deprecated` tag.

**Why:** `hassApi.ts` imports the type from `ZwaveClient` while importing the result type from `hass/types`, so the intended direction is already ambiguous to callers.

**Fix:** Mark the alias deprecated and migrate the in-repo imports.

api/runtime/GatewayFactory.ts
```diff
@@ -0,0 +1,36 @@
...
+	public create<TZwave extends GatewayZwave, TMqtt extends GatewayMqtt>(
+		config: GatewayConfig,
+		zwave: TZwave | undefined,
+		mqtt: TMqtt | undefined,
+	): Gateway<TZwave, TMqtt> {
+		this.registry.start()
+		return new Gateway(config, zwave, mqtt, this.registry.fork())
+	}
+
+	public dispose(): void {
```
AlCalzone, line 33: **[Nit] · Design**

`dispose()` introduces a fourth teardown verb.

**Why:** Existing api teardown is `close()` (`SocketManager`, `MqttClient`, `ZnifferManager`, `Gateway`, `ManagedService`) or `destroy()`. Optional.

api/hass/DeviceStore.ts
```diff
@@ -0,0 +1,80 @@
+import * as utils from '../lib/utils.ts'
+import type { HassDeviceStorePort, HassPersistenceNode } from './ports.ts'
+import type {
+	HassDevice,
+	HassDeviceMap,
+	StoreHassDevicesResult,
+} from './types.ts'
+
```
AlCalzone, line 8: **[Nit] · Design**

File `DeviceStore.ts` exports `HassDeviceStore`.

**Why:** Sibling modules match filename to class name exactly. Optional.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+				"{{'true' if value_json.value else 'false'}}"
+		}
+		payload.availability_mode = 'all'
+	}
+
+	private getDiscoveryTopic(
+		hassDevice: HassDevice,
+		nodeName: string,
+	): string {
+		return `${hassDevice.type}/${utils.sanitizeTopic(nodeName, true)}/${
```
AlCalzone, line 1319: **[FYI] · Security**

`hassDevice.type` and `hassDevice.object_id` are interpolated into the discovery topic unsanitized — only `nodeName` goes through `utils.sanitizeTopic` — so a custom-device definition or a client-supplied device can inject `/` segments.

**Why:** Predates this PR and is unchanged from `Gateway._getDiscoveryTopic`. Noted for the record, no action expected in this PR.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+										Constants.deviceClass.sensor_binary
+											.MOISTURE,
+									)
+								}
+								discoveredObjectId = valueId.propertyName
+								break
+						}
+						config ||= utils.copy(hassCfg.binary_sensor)
+						this.setBinaryPayloadFromSensor(config, states, off)
+						config.object_id = String(discoveredObjectId)
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — discovery now skips binary notifications that have no discoverable object ID.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+	private subscribe(listener: () => void): () => void {
+		this.listeners.add(listener)
+		return () => this.listeners.delete(listener)
+	}
+
+	private notify(): void {
+		for (const listener of this.listeners) listener()
+	}
+
+	private watch(filename: string): void {
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 1d520b1b — watcher errors are logged, and the failed watcher is closed and removed from registry ownership.

api/runtime/GatewayFactory.ts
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — GatewayFactory now lives in the runtime composition layer, removing the HASS-to-Gateway dependency.

api/hass/ports.ts
```diff
@@ -0,0 +1,142 @@
...
+		topic: string,
+		data: unknown,
+		options?: IClientPublishOptions,
+		prefix?: string,
+	): unknown
+}
+
+export interface HassZwavePort {
+	readonly homeHex: string | undefined
+	getNode(nodeId: number): unknown
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — discovery now consumes concrete typed node maps and initializes owner-derived HASS maps without redundant record validation.

api/hass/ports.ts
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — the port exposes writeValue, and the exact cover STOP mapping to Up=false now remains in HASS transformation code.

api/hass/ports.ts
```diff
@@ -0,0 +1,142 @@
...
+	ccConfigEnableDiscovery?: boolean
+}
+
+export interface HassValueTopic {
+	topic: string
+	valueConf?: HassValueConfiguration
+	targetTopic?: string
+}
+
+export interface HassTopicPort {
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — MQTT, Z-Wave, and logger capabilities now use Pick<> types, and Gateway passes compatible full collaborators directly.

api/lib/Gateway.ts
AlCalzone: Fixed in 8d471fc6 — compatible full MQTT and Z-Wave instances are passed directly; only the Gateway-owned node-update adapter remains.

api/lib/Gateway.ts
AlCalzone: Fixed in 8d471fc6 — the forwarding wrappers were removed in favor of structurally typed full collaborators and one genuine node-update adapter.

test/lib/hass/persistence.test.ts
```diff
@@ -1,464 +1,248 @@
...
-import type { Server as SocketServer } from 'socket.io'
-import { ensureTestEnv, cleanupTestEnv, getTestStoreDir } from './env.ts'
-import {
-	buildNode,
-	createRecordingSocket,
-	type RecordingSocket,
-} from './fixtures.ts'
-import { socketEvents } from '#api/lib/SocketEvents.ts'
-import type ZWaveClientType from '#api/lib/ZwaveClient.ts'
+import { beforeEach, describe, expect, it } from 'vitest'
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 57f6ff23 — thin real ZwaveClient/jsonStore tests now cover home scoping, legacy migration, array conversion, and disk persistence.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+					}
+				}),
+			)
+		} catch {
+			this.watchers.set(
+				filename,
+				fs.watch(path.dirname(filename), (_event, changedFilename) => {
+					if (
+						!changedFilename ||
+						changedFilename === 'customDevices.js' ||
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 1d520b1b — fallback watcher matches now derive both filenames from the registry instance paths.

api/lib/Gateway.ts
```diff
@@ -1,192 +1,50 @@
...
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — runtime composition now supplies the documented HASS logger tag consistently.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+							template,
+						)
+					}
+				}
+			}
+
+			const skipDiscovery =
+				hassDevice.ignoreDiscovery ||
+				(this.config.manualDiscovery && !options.forceUpdate)
+			if (!skipDiscovery && hassDevice.discoveryTopic) {
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — a missing discovery topic now enters the legacy publication error path instead of silently suppressing the publish.

api/hass/types.ts
```diff
@@ -0,0 +1,45 @@
...
+	deleteDevice?: boolean
+	forceUpdate?: boolean
+}
+
+export type StoreHassDevicesResult =
+	| { status: 'stored' }
+	| { status: 'node-not-found' }
+	| { status: 'invalid-stored-node' }
+
+export const HASS_NODE_PREFIX = 'nodeID_'
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — Gateway and discovery now share the exported HASS_NODE_PREFIX constant.

api/hass/ports.ts
```diff
@@ -0,0 +1,142 @@
...
+	hasNode(nodeId: number): boolean
+	getNodeDevices(nodeId: number): HassDeviceMap | undefined
+	setNodeDevices(nodeId: number, devices: HassDeviceMap): void
+	getStoredNode(nodeId: number): unknown
+	setStoredNode(nodeId: number, node: HassPersistenceNode): void
+	emitNodeUpdate(nodeId: number, devices: HassDeviceMap): void
+	updateStoreNodes(): Promise<void>
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — typed node capabilities replaced the duplicate record guard.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+			if (value.commandClass === CommandClasses['Color Switch']) {
+				prioritizedValueIds.unshift(id)
+			} else {
+				prioritizedValueIds.push(id)
+			}
+		}
+		return prioritizedValueIds
+	}
+
+	private getIdWithoutNode(valueId: HassValue): string {
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — Gateway and discovery now share getIdWithoutNode.

api/runtime/AppRuntime.ts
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — GatewayFactoryPort now lives beside the other runtime ports.

api/socket/hassApi.ts
```diff
@@ -82,15 +83,26 @@ export function registerHassApiHandler(
...
+						const storeResult = await runtime
 							.ensureZWaveClient()
 							.storeDevices(
 								data.devices,
 								data.nodeId,
 								data.remove,
 							)
+						res = storeResult
+						if (storeResult.status === 'node-not-found') {
+							err =
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 57f6ff23 — persistence status remains internal, and the Socket.IO store action again emits the legacy success ACK without a result field.

api/lib/ZwaveClient.ts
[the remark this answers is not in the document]
AlCalzone: Fixed in 57f6ff23 — hassApi imports HassDevice and HassDeviceMap directly from hass/types; the compatibility re-export remains unchanged.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+				`Error while discovering device ${hassId} of node ${
+					node.id
+				}: ${getErrorMessage(error)}`,
+				error,
+			)
+		}
+	}
+
+	public discoverClimates(node: HassNode): void {
+		if (!node.deviceClass || node.deviceClass.generic !== 0x08) return
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — thermostat discovery now uses a named generic-device-class constant.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+					)
+					if (
+						[
+							'specific_type_class_a_motor_control',
+							'specific_type_class_b_motor_control',
+							'specific_type_class_c_motor_control',
+							'specific_type_class_motor_multiposition',
+							'specific_type_motor_multiposition',
+						].includes(specificDeviceClass) ||
+						node.deviceId === '615-0-258'
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — the Simon IO Roller Blind device ID is named, with the #3088 motor-class workaround retained beside its use.

api/hass/DiscoveryGenerator.ts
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — HASS open and closed cover positions are named, with the separate Z-Wave tilt-position rationale documented.

api/hass/ports.ts
```diff
@@ -0,0 +1,142 @@
...
+	ZUIDeviceClass,
+	ZUIEndpoint,
+	ZUINode,
+	ZUIValueId,
+	ZUIValueIdState,
+} from '../lib/ZwaveClient.ts'
+import type { PayloadType } from '../lib/shared.ts'
+import type { HassDevice, HassDeviceMap } from './types.ts'
+
+export interface HassLogger {
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 8d471fc6 — HassLogger is now derived from ModuleLogger with Pick<> and accepts full logger instances structurally.

api/hass/DiscoveryGenerator.ts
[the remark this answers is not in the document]
AlCalzone: False positive — valid Multilevel Sensor metadata supplies a numeric sensorType, and the existing characterization covers the supported path.

api/hass/DiscoveryGenerator.ts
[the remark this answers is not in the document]
AlCalzone: False positive — zwave-js creates currentColor and targetColor together for Color Switch values, so the characterized lookup is invariant.

api/hass/DiscoveryGenerator.ts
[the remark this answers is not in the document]
AlCalzone: False positive — zwave-js Thermostat Mode metadata always supplies states; the existing reply and characterization cover this invariant.

test/lib/hass/customDevices.test.ts
```diff
@@ -1,168 +1,78 @@
...
- * loader resolves a `.js` file (preferred) or `.json`, overlays its entries onto
- * the base catalog, and reports the merged catalog, a dedup hash, the
- * custom-device count, and which file was used - or nothing when neither file
- * exists or the JSON is malformed. Tests inject a base path and base catalog so
- * they drive the real loader against throwaway fixtures under an isolated store
- * dir, never the repo store/.
- */
-import { describe, it, expect, beforeAll, afterAll } from 'vitest'
-import { writeFileSync } from 'node:fs'
+import { afterAll, afterEach, beforeAll, expect, it, vi } from 'vitest'
```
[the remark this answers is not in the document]
AlCalzone: False positive — direct registry tests already cover JSON/JS precedence, malformed loading, cache eviction, and source reload behavior.

test/lib/hass/CustomDeviceRegistry.direct.test.ts
```diff
@@ -0,0 +1,270 @@
+import * as fs from 'node:fs'
```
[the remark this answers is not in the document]
AlCalzone: False positive — watcher ownership, rebind, disposal, and now fatal watcher-error cleanup are covered through the public registry lifecycle.

test/lib/hass/DiscoveryGenerator.test.ts
```diff
@@ -0,0 +1,1010 @@
+import { CommandClasses } from '@zwave-js/core'
```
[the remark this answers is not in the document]
AlCalzone: False positive — direct discovery tests already cover present, missing, and incomplete lookup cases through public behavior.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+	private readonly listeners = new Set<() => void>()
+	private readonly source: CustomDeviceRegistry | undefined
+	private allDevices: HassDeviceCatalogSource
+	private lastCustomDevicesLoad: string | null = null
+	private unsubscribeSource: (() => void) | undefined
+	private started = false
+
+	public constructor(options: CustomDeviceRegistryOptions) {
+		this.source = options.source
+		this.baseDevices = copyCatalog(options.devices)
```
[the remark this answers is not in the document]
AlCalzone: False positive — each fork is intentionally seeded from the source projection before subscribing, preventing a startup gap.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+			source: this,
+		})
+	}
+
+	public dispose(): void {
+		this.unsubscribeSource?.()
+		this.unsubscribeSource = undefined
+		for (const watcher of this.watchers.values()) watcher.close()
+		this.watchers.clear()
+		this.listeners.clear()
```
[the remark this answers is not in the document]
AlCalzone: False positive — restarting a disposed source while live forks remain is outside the supported ownership lifecycle.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+			this.watchers.set(
+				filename,
+				fs.watch(filename, (event: string) => {
+					this.load()
+					if (event === 'rename') {
+						this.watchers.get(filename)?.close()
+						this.watch(filename)
+					}
+				}),
+			)
```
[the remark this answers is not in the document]
AlCalzone: False positive — absent custom-device files normally use directory fallback, so that expected path intentionally remains quiet.

api/hass/DeviceStore.ts
```diff
@@ -0,0 +1,80 @@
...
+		nodeId: number,
+		remove: unknown,
+	): Promise<StoreHassDevicesResult> {
+		if (!this.port.hasNode(nodeId)) return { status: 'node-not-found' }
+
+		const storedNode = this.port.getStoredNode(nodeId)
+		let persistenceNode: HassPersistenceNode
+		if (storedNode === undefined) {
+			persistenceNode = {}
+			this.port.setStoredNode(nodeId, persistenceNode)
```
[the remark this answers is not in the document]
AlCalzone: False positive — updateStoreNodes intentionally omits empty persistence records before writing nodes.json.

api/lib/Gateway.ts
```diff
@@ -704,1256 +664,88 @@ export default class Gateway {
...
-			})
 		}
+		this.discoveryGenerator.rediscoverNode(nodeID)
 	}
 
-	/**
-	 * Disable the discovery of all devices of this node
-	 *
-	 */
 	disableDiscovery(nodeId: number): void {
```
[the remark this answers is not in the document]
AlCalzone: False positive — these Gateway discovery methods are deliberate compatibility facades for existing callers during the extraction.

api/hass/DiscoveryGenerator.ts
```diff
@@ -0,0 +1,1512 @@
...
+				hassDevice.mode_map
+			) {
+				return hassDevice.mode_map[payload]
+			}
+		} else if (
+			hassDevice.type === 'cover' &&
+			valueId.property === 'targetValue' &&
+			payload === (hassDevice.discovery_payload.payload_stop ?? 'STOP')
+		) {
+			this.zwave.writeCoverStop(valueId).catch(() => {})
```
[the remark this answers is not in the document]
AlCalzone: Skipped — splitting the broader compatibility side effects is outside this focused extraction layer.

api/hass/CustomDeviceRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+	private readonly source: CustomDeviceRegistry | undefined
+	private allDevices: HassDeviceCatalogSource
+	private lastCustomDevicesLoad: string | null = null
+	private unsubscribeSource: (() => void) | undefined
+	private started = false
+
+	public constructor(options: CustomDeviceRegistryOptions) {
+		this.source = options.source
+		this.baseDevices = copyCatalog(options.devices)
+		this.allDevices = options.source
```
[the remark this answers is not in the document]
AlCalzone: Skipped — the small startup projection copy is intentional isolation and not worth additional lifecycle complexity.

api/runtime/GatewayFactory.ts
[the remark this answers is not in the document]
AlCalzone: Skipped — dispose() matches the registry lifecycle vocabulary already used by its owner; this is a naming preference.

api/runtime/AppRuntime.ts
AlCalzone: Addressed in 8d471fc6 — GatewayFactoryPort now lives with the runtime ports and uses the structurally typed gateway collaborators.

api/hass/ports.ts
AlCalzone: Addressed in 8d471fc6 — the partial runtime guard was removed; concrete typed nodes now initialize the owner-derived HASS maps.

api/hass/ports.ts
AlCalzone: Addressed in 8d471fc6 — the duplicate record guard was removed as part of the typed-node cleanup.

api/runtime/GatewayFactory.ts
AlCalzone: Addressed in 8d471fc6 — GatewayFactory moved from api/hass to the runtime composition layer.

api/lib/Gateway.ts
AlCalzone: Addressed in 8d471fc6 — Gateway and discovery now share HASS_NODE_PREFIX.

api/lib/ZwaveClient.ts
AlCalzone: Addressed in 57f6ff23 — in-repo HASS types now import directly from hass/types; the compatibility re-export intentionally remains non-deprecated.

api/runtime/GatewayFactory.ts
AlCalzone: Addressed by the approved prior verdict — dispose() remains the registry lifecycle term, so no naming-only change was made.

api/hass/CustomDeviceRegistry.ts
AlCalzone: Addressed by the approved prior verdict — the bounded startup projection copy remains for fork isolation.

api/hass/DiscoveryGenerator.ts
AlCalzone: No change needed — zwave-js creates the target RGB value alongside the current value, so this missing-target shape is not reachable at runtime. This is the same invariant covered by the earlier color-topic review thread.

api/hass/DiscoveryGenerator.ts
AlCalzone: Fixed in aefeb760 — the values requirement now applies only to non-climate custom devices, preserving legacy climate registration before publication validation.

api/lib/Gateway.ts
AlCalzone: Fixed in 2a51f52a — DiscoveryGenerator now owns and resets its discovery index; Gateway no longer exposes or replaces the map.

api/hass/DiscoveryGenerator.ts
AlCalzone: Fixed in aefeb760 — empty metadata units now fall through to the value unit with the legacy truthiness semantics.

api/lib/Gateway.ts
AlCalzone: Fixed in aefeb760 and b5fcd461 — cover STOP uses a distinct handled sentinel, while literal null reaches both direct and compatibility broadcast writes.

api/hass/DiscoveryGenerator.ts
AlCalzone: Fixed in aefeb760 — Alarm Sensor lookup now preserves both numeric and string enum keys without unsafe casts.

api/hass/CustomDeviceRegistry.ts
AlCalzone: Skipped for Layer 7 — cache eviction is an explicitly locked compatibility quirk for custom JS reload behavior.

api/hass/CustomDeviceRegistry.ts
AlCalzone: No change needed — restarted forks copy the retained source catalog during start, so retaining the dedupe hash cannot suppress projection.

api/hass/CustomDeviceRegistry.ts
AlCalzone: Skipped — splitting owner and projection roles is a broader redesign outside these focused compatibility fixes.

api/hass/CustomDeviceRegistry.ts
AlCalzone: Skipped — the private synchronous listener Set is narrower than adding an event API and has no duplicated production behavior.

api/hass/ports.ts
AlCalzone: No change needed — deriving these ports from Gateway types would recreate the forbidden HASS-to-Gateway dependency.

api/hass/DeviceStore.ts
AlCalzone: Skipped — the filename suggestion is naming-only and unrelated to Layer 7 behavior.

api/hass/DiscoveryGenerator.ts
AlCalzone: Skipped as noted — this discovery-topic behavior predates the extraction and is unchanged in Layer 7.

AlCalzone:
Second-review unanchored findings:

- Fixed HASS import persistence failures in `63e6c294`; non-stored results now return an explicit failed import response.
- No change to `ValueIdTopic`; Gateway needs the wider `GatewayValue` topic shape, while `HassValueTopic` is intentionally narrower.
- No change to arbitrary authenticated discovery publication or publication fencing; both were marked FYI and are unchanged from the pre-extraction behavior.

robertsLando (approved):

PR #4743, opened 2026-07-13
feat(bot): answer questions from docs and related posts

AlCalzone:
## Summary

- port the consolidated AI support bot from zwave-js PRs [#8937](https://github.com/zwave-js/zwave-js/pull/8937), [#8940](https://github.com/zwave-js/zwave-js/pull/8940), [#8941](https://github.com/zwave-js/zwave-js/pull/8941), and [#8942](https://github.com/zwave-js/zwave-js/pull/8942)
- index the Z-Wave JS UI documentation, issues, and Q&A discussions with GitHub Models embeddings
- answer new issues and Q&A discussions from documentation and suggest related posts in one bot comment
- learn from reactions, suppress similar low-quality answers, and maintain nightly retrieval evaluations and feedback tracking
- migrate the existing CommonJS bot scripts to `.cjs` so they load correctly with this repository's `type: module` configuration

## Z-Wave JS UI adaptations

- use `https://zwave-js.github.io/zwave-js-ui/#` for documentation links
- limit discussion handling to the existing `q-a` category
- use the existing `documentation` label for the feedback digest
- preserve the current UI-vs-driver issue classifier and existing bot comment commands
- post live immediately, matching the final upstream behavior


.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -0,0 +1,573 @@
...
+		})
+		.join("\n");
+
+	const sections = deduped.map((chunk) => `${chunk.file}#${chunk.anchor}`);
+	const single = deduped.length === 1;
+	if (
+		allowAnswer && result.confidence >= ANSWER_CONFIDENCE && result.answer
+	) {
+		return {
+			text: `${result.answer}
```
robertsLando: **[Blocker] · Security**

`result.answer` is raw model output — driven by the attacker-controlled issue/discussion body embedded in the prompt — and it is interpolated straight into the comment body and posted as `zwave-js-bot` with no sanitization.

**Why:** Prompt injection lets any anonymous user make the trusted bot account emit `@mention` mass-pings, phishing links, or remote images that IP-log every viewer of the thread. Notably this PR *already ships* the mitigation: `neutralizeMentions()` in `updateDocsFeedbackIssue.cjs:35`. It is applied to the low-visibility digest issue but not to the high-visibility public comment.

**Fix:** Treat model output as untrusted before publishing it: neutralize mentions, strip raw HTML/image syntax, and constrain link targets to `DOCS_BASE_URL`/github.com. `confidence` is likewise an attacker-steerable control signal.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -0,0 +1,573 @@
...
+		console.log("Failed to parse model response:", e);
+		return;
+	}
+	console.log("Model response:", JSON.stringify(result));
+
+	const related = (result.relatedExcerpts ?? [])
+		.map((i) => ranked[i]?.chunk)
+		.filter(Boolean);
+
+	if (result.confidence < LINKS_CONFIDENCE || related.length === 0) {
```
robertsLando, line 309: **[Major] · Correctness**

A malformed model response bypasses the confidence floor. If `result.confidence` is `undefined`, `undefined < LINKS_CONFIDENCE` evaluates to `false`, so this guard passes and a links comment is posted anyway. Separately, a non-array `relatedExcerpts` throws on `.map` at line 233 — outside the try/catch at 224.

**Why:** The floor is the only thing standing between an off-topic question and a confidently-wrong bot comment on a real user's issue.

**Fix:** Validate the shape and types of the parsed object after `JSON.parse` (line 132) and bail on anything unexpected, rather than trusting the model to return well-formed JSON.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -0,0 +1,573 @@
...
+	rankRelatedPosts,
+} = require("./postsIndex.cjs");
+
+const DOCS_BASE_URL = "https://zwave-js.github.io/zwave-js-ui/#";
+const DOCS_ANSWER_COMMENT_TAG = "<!-- DOCS_ANSWER_COMMENT_TAG -->";
+const DOCS_ANSWER_METADATA_TAG = "DOCS_ANSWER_METADATA";
+const DOCS_ANSWER_METADATA_VERSION = 1;
+
+// Users whose posts should never be answered automatically
+const EXCLUDED_USERS = ["AlCalzone", "zwave-js-bot"];
```
robertsLando: **[Major] · Design / DRY**

`EXCLUDED_USERS = ["AlCalzone", "zwave-js-bot"]` is an unadapted upstream constant. This repo's maintainer set is `AlCalzone, robertsLando, blhoward2` (`authorizedUsers.cjs:3`). The same list is also hardcoded a second time as a YAML `if:` guard in `zwave-js-bot_post.yml:69`.

**Why:** As written, issues opened by **robertsLando** (this repo's primary maintainer) and **blhoward2** will be auto-answered by the docs bot — exactly what the exclusion exists to prevent. And the two copies of the list can drift.

**Fix:** Single-source the exclusion list (derive it from `authorizedUsers.cjs`, or export it from the script and drop the redundant YAML guard) and populate it with *this* repo's maintainers.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -0,0 +1,573 @@
...
+		return;
+	}
+
+	// Load the pre-built embeddings indices. Either may be missing,
+	// each one enables its part of the comment.
+	const docsIndexPath = process.env.DOCS_INDEX_PATH;
+	/** @type {any} */
+	let docsIndex;
+	try {
+		docsIndex = JSON.parse(await fs.readFile(docsIndexPath, "utf8"));
```
robertsLando: **[Major] · Correctness / Design**

The docs index is loaded with a bare `JSON.parse` and its `version`/`model` are never validated — unlike the posts path, where `postsIndex.cjs:92 loadPostsIndex()` correctly rejects a version mismatch. `INDEX_VERSION` also lives in the *builder* (`buildDocsIndex.cjs:13`) rather than in the shared reader.

**Why:** A stale-shaped index is silently accepted at read time. Combined with the cache-key issue on `docs-embeddings.yml:33`, bumping `INDEX_VERSION` or changing `EMBEDDING_MODEL` becomes a **no-op** and the old index keeps being served indefinitely.

**Fix:** Mirror the posts path: move the version constant into `docsIndex.cjs` behind a `loadDocsIndex()` that validates version + model, and have all three call sites (here, `evalDocsAnswers.cjs`) use it.

.github/workflows/docs-embeddings.yml
```diff
@@ -0,0 +1,116 @@
...
+    steps:
+    - name: Checkout master branch
+      uses: actions/checkout@v7
+
+    - name: Restore previous index
+      id: cache
+      uses: actions/cache@v6
+      with:
+        path: .docs-index
+        key: docs-embeddings-v1-${{ hashFiles('docs/**/*.md') }}
```
robertsLando: **[Major] · Correctness / Design**

The cache key hashes only `docs/**/*.md` — not the chunking code, not `EMBEDDING_MODEL` — and an exact hit *skips the build step entirely* (line 40).

**Why:** Changing the chunking logic, `slugify`, `INDEX_VERSION`, or the embedding model never rebuilds the index: no docs edit means no push trigger, an exact key hit, and a skipped build. So the `previous.version === INDEX_VERSION && previous.model === EMBEDDING_MODEL` guard in `buildDocsIndex.cjs:194` never even executes. The only way to force a rebuild is hand-editing the `v1` literal across four separate workflow keys.

**Fix:** Fold a hash of the bot scripts (and the model name) into the cache key so code changes invalidate the index.

robertsLando (commented):
## Deep review — 8 specialist passes (correctness, security, performance, DRY, design, tests, operability, readability)

**Verdict: Needs work** — one Blocker, then this is a good change. The approach is sound and the port is faithful; this is a "fix the Blocker and the port artifacts" review, not a rework.

### Strengths

The `.cjs` migration is complete and correct, and it fixes **live breakage**: root `package.json` is `"type": "module"`, so today's `require(".../bot-scripts/index.js")` in `zwave-js-bot_comment.yml` resolves as ESM — the existing bot commands are almost certainly broken on master right now. All 13 call sites were updated, and `post-release-notes.js` is genuine ESM and correctly left alone.

The Z-Wave JS UI adaptation is real, not a copy-paste: all 18 `expectedFiles` in `docsAnswersEvalCases.json` resolve to actual files under `docs/`, the `q-a` category slug and `documentation` label match this repo, and the deleted workflow's UI-vs-driver classifier is preserved verbatim (same prompt, model, token, trigger types) behind a correct `if: github.event_name == 'issues'` guard. Using the Actions cache as the vector store — no external DB, no new service, no runtime deps — is the right complexity level for a repo-scoped RAG bot.

### Top 3 risks

1. **Raw LLM output is posted as the bot with no sanitization**, on a path any anonymous user can trigger by opening an issue. The mitigation (`neutralizeMentions`) ships in this very PR — it's just applied to the digest, not the public comment.
2. **The docs index can never be invalidated by a code change.** `INDEX_VERSION` is effectively a no-op, and a stale-schema index is served indefinitely.
3. **The failure modes that matter all end in a green check.** Missing index → steps skipped → job succeeds → bot silently stops answering, and nobody finds out.

### Themes

- **Port artifacts** — upstream constants that were correct for `zwave-js` and are wrong here. The `EXCLUDED_USERS` one is worth calling out directly: as written, **@robertsLando's own issues get auto-answered by the bot**.
- **The docs path is the poor cousin of the posts path** — `postsIndex.cjs` has a version-validating loader; `docsIndex.cjs` has none. The same asymmetry drives three separate findings.
- **Silent green** — cache misses, API outages, and quota exhaustion are all indistinguishable from a healthy run.

### Findings not anchored to a diff line

- **No `timeout-minutes:` on any new job.** `embedBatched` throttles 4.5s/request and `modelsRequest` retries 5× with up to 60s backoff, so a degraded Models API leaves nightly jobs spinning toward the 6h default.
- **No README for the subsystem.** 13+ new files and no entry point explaining the `buildDocsIndex → answerFromDocs → collectDocsFeedback → updateDocsFeedbackIssue` pipeline, how to rebuild an index locally, or how to add an eval case. A short `.github/bot-scripts/README.md` would pay for itself the first time someone else touches this.
- **`answerFromDocs.cjs:506-551`** — `main()` inlines ~45 lines of comment-body string composition (nested ternaries, conditional separators, metadata JSON) directly in the orchestration function, obscuring its actual control flow. A `composeComment()` helper would mirror the existing `buildDocsAnswerSection`/`buildRelatedPostsSection` decomposition.
- **Duplicated tracking-issue logic** — `updateDocsFeedbackIssue.cjs` and `updateEvalTrackingIssue.cjs` both find-or-create a marker-identified tracking issue, via two different API layers (raw `fetch` over `githubApi.cjs` vs the injected octokit) and two different idempotency schemes, invoked from the same workflow two different ways (`node` vs `github-script`). Only one of them paginates — which is the bug flagged at `updateEvalTrackingIssue.cjs:27`.

### FYI — checked and *not* held against this PR

- `BOT_TOKEN` on the anonymous `issues: [opened, edited]` path, and the missing `permissions:` block on `check-issue-repo`, are both **pre-existing** — I diffed against the deleted `zwave-js-bot_issue.yml`, which already did exactly this. Not a regression. But since the file is being rewritten anyway, a top-level `permissions: {}` with per-job grants is nearly free.
- Floating action tags (`checkout@v7`, `cache@v6`, …) in workflows holding a PAT are a supply-chain exposure, but that's this repo's existing convention everywhere — out of scope here.

### Coverage

All 8 specialists run, on security-critical model tiering. Cross-lens agreement on the `EXCLUDED_USERS` port artifact (2 specialists), the docs-index invalidation gap (3), and the duplicate-comment risk (3). Every finding below was verified against the PR branch before submission.

.github/workflows/zwave-js-bot_post.yml
```diff
@@ -0,0 +1,165 @@
+name: 'Z-Wave Bot: react to Issues/Discussions'
+
+on:
+  issues:
+    types: [opened, edited]
```
robertsLando, line 5: **[Major] · Correctness / Operability**

The workflow fires on both `opened` and `edited`, and there is no `concurrency:` group anywhere in the file.

**Why:** Users routinely fix a typo seconds after opening an issue. That spawns a second run which races the first through `alreadyAnswered()` (a TOCTOU check) — both pass, both post. The result is **two bot comments on a real user's issue**.

**Fix:** Add a per-post concurrency group. Note this is only half the fix — see the `alreadyAnswered` pagination issue at `answerFromDocs.cjs:55`.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -0,0 +1,573 @@
...
+	return url;
+}
+
+/**
+ * Checks whether the bot already answered this post
+ * @param {{github: Github, context: Context}} param0
+ * @param {any} post
+ * @param {boolean} isDiscussion
+ */
+async function alreadyAnswered({ github, context }, post, isDiscussion) {
```
robertsLando, line 65: **[Major] · Correctness**

`alreadyAnswered` scans only a single page of comments (`comments(first: 50)` for discussions, one `per_page: 100` page for issues) with no pagination.

**Why:** Once a post exceeds the cap the guard silently returns `false` and the bot posts a **second** answer. The `edited` trigger makes this reachable on any long-running thread.

**Fix:** Rather than scanning a bounded page, query directly for a bot-authored comment carrying the metadata tag — or paginate.

.github/workflows/zwave-js-bot_post.yml
```diff
@@ -0,0 +1,165 @@
...
+          const aiResponse = process.env.AI_RESPONSE;
+          console.log('Raw AI response:', JSON.stringify(aiResponse));
+
+          const bot = require(`${process.env.GITHUB_WORKSPACE}/.github/bot-scripts/index.cjs`);
+          return bot.classifyIssueFeedback({github, context}, aiResponse);
+
+  # Answer questions that are covered by the documentation and/or
+  # suggest similar existing posts. Discussions are restricted to the
+  # Q&A category by answerFromDocs.cjs itself.
+  answer-from-docs:
```
robertsLando, line 76: **[Major] · Operability**

There is no runtime kill switch on `answer-from-docs`; the only guard is a hardcoded user list.

**Why:** If the model starts emitting garbage at 3am, the mitigations are (a) edit + merge YAML, or (b) disable the whole workflow in the Actions UI — which now *also* kills the wrong-repo notifier and the posts-index updater, both folded into this file when `zwave-js-bot_issue.yml` was deleted. Neither is fast.

**Fix:** Gate the job on a repo variable (e.g. `if: vars.DOCS_BOT_ENABLED != 'false'`) that a maintainer can flip in seconds without a merge.

.github/workflows/docs-embeddings.yml
```diff
@@ -0,0 +1,116 @@
...
+    - name: Restore docs index
+      id: restore-index
+      uses: actions/cache/restore@v6
+      with:
+        path: .docs-index
+        key: docs-embeddings-v1-${{ hashFiles('docs/**/*.md') }}
+
+    - name: Evaluate retrieval quality
+      id: eval
+      if: steps.restore-index.outputs.cache-hit == 'true'
```
robertsLando: **[Major] · Operability**

The eval and tracking-issue steps are both gated on `cache-hit == 'true'`. A cache miss or eviction means both are skipped — and the job reports **success**.

**Why:** A missing index is indistinguishable from a healthy run. The same shape appears at `zwave-js-bot_post.yml:108` (`cache-matched-key != ''`), where the bot silently stops answering behind a green check. The tracking issue only fires on a *retrieval-quality* regression, never on infra absence.

**Fix:** "Index expected but absent" needs to be a failure/alert path, not a silent skip.

.github/workflows/posts-embeddings.yml
```diff
@@ -0,0 +1,92 @@
...
+    - name: Restore posts index
+      id: restore-index
+      uses: actions/cache/restore@v6
+      with:
+        path: .posts-index
+        key: posts-embeddings-v1-${{ github.run_id }}
+
+    - name: Evaluate retrieval quality
+      id: eval
+      if: steps.restore-index.outputs.cache-hit == 'true'
```
robertsLando: **[Major] · Operability**

Same silent-green shape as `docs-embeddings.yml:66` — eval and tracking-issue steps gated on `cache-hit == 'true'`, so a cache miss yields a passing job that did nothing.

**Why:** The posts index can quietly go stale or vanish with no signal.

**Fix:** Same fix: treat a missing-but-expected index as a failure.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -0,0 +1,573 @@
...
+		return;
+	}
+
+	// Ask the model whether the docs answer the question
+	/** @type {{confidence: number, answer: string | null, relatedExcerpts: number[]}} */
+	let result;
+	try {
+		result = await judgeAnswer(question, ranked, token);
+	} catch (e) {
+		console.log("Failed to parse model response:", e);
```
robertsLando: **[Major] · Operability**

The `try` at line 224 wraps the entire `judgeAnswer()` call, but the handler reports `"Failed to parse model response"` and returns.

**Why:** A Models 429-after-5-retries or a 5xx is swallowed, mislabeled as a parse error, and the job exits **green**. A quota exhaustion or API outage looks identical to a normal "confidence too low, not answering" — so there is no signal anywhere that the bot has gone dark.

**Fix:** Narrow the catch to the `JSON.parse` itself and let genuine API errors fail the step.

.github/bot-scripts/collectDocsFeedback.cjs
```diff
@@ -0,0 +1,446 @@
...
+	// The suppression list makes answerFromDocs.cjs demote responses
+	// to questions similar to these. The indices are built with the same
+	// EMBEDDING_MODEL, so the runtime similarity comparison is valid.
+	// Related-posts-only comments say nothing about the docs answer
+	// quality, so they don't suppress anything.
+	const downvoted = records.filter(
+		(r) => r.score <= SUPPRESS_SCORE && r.style !== "posts",
+	);
+	const embeddings = downvoted.length > 0
+		? await embed(downvoted.map((r) => r.question), token)
```
robertsLando: **[Major] · Performance**

The suppression-list embedding is a single unbatched `embed()` call over *all* downvoted questions, bypassing the `embedBatched` helper that exists precisely to respect the request token limit (and which the sibling builders already use).

**Why:** At up to ~6000 chars (~1500 tokens) per question, this exceeds the 64K-tokens-per-request limit somewhere around 40 downvoted answers. The nightly feedback job then hard-fails permanently — and takes the suppression guardrail down with it, so the bot resumes re-posting answers users have already downvoted.

**Fix:** Use `embedBatched` here. Also consider capping/aging-out the suppression list, which is currently unbounded and cosine-scanned on every bot run.

.github/bot-scripts/updatePostsIndex.cjs
```diff
@@ -0,0 +1,112 @@
...
+	if (existing) {
+		index.posts[index.posts.indexOf(existing)] = entry;
+		console.log(`Updated ${type} #${post.number} in the posts index`);
+	} else {
+		index.posts.push(entry);
+		console.log(`Added ${type} #${post.number} to the posts index`);
+	}
+
+	// indexPath is defined, loadPostsIndex would have returned undefined otherwise
+	await fs.writeFile(
```
robertsLando, line 105: **[Major] · Performance**

Every new post rewrites and re-uploads the *entire* index as a fresh cache entry.

**Why:** Embeddings are stored as JSON float text (1536 dims × ~9 bytes ≈ 14 KB/vector). At today's post count that is roughly 9 MB restored, parsed, re-serialized and re-uploaded **per post**, retained 7 days, growing monotonically and LRU-pressuring the repo's 10 GB Actions cache budget. The cost is O(index), not O(1).

**Fix:** Store vectors as base64 float32 (≈2.3× smaller, cheaper to parse), and/or append deltas to a small separate cache entry that readers merge over the nightly base.

.github/workflows/zwave-js-bot_post.yml
```diff
@@ -0,0 +1,165 @@
...
+          await bot.answerFromDocs({github, context});
+
+  # Add the new post to the posts index, so it can be suggested as a
+  # related post before the next nightly rebuild. Discussions outside
+  # the Q&A category are ignored by updatePostsIndex.cjs itself.
+  # Edits are reconciled by the nightly rebuild instead, so repeated
+  # edits cannot burn Models API requests and cache storage.
+  update-posts-index:
+    runs-on: [ubuntu-latest]
+    permissions:
```
robertsLando, line 146: **[Major] · Performance**

`update-posts-index` spins up a second runner that re-checks-out, re-restores the posts index, and **re-embeds the identical `cleanQuestion(title, body)` text** that `answer-from-docs` embedded seconds earlier in the same workflow.

**Why:** Two runners and two Models embedding requests per post, for one logical unit of work — against a 15 req/min quota.

**Fix:** Fold the append into the answer job so the embedding is computed once and reused, or pass it forward between jobs.

.github/bot-scripts/githubApi.cjs
```diff
@@ -0,0 +1,110 @@
...
+
+const API_BASE = "https://api.github.com";
+
+/**
+ * @param {string} method
+ * @param {string} url
+ * @param {object | undefined} body
+ * @param {string} token
+ */
+async function ghFetch(method, url, body, token) {
```
robertsLando, line 95: **[Major] · Performance / Operability**

`ghFetch` has no retry or backoff, unlike `modelsRequest` which implements Retry-After + exponential backoff.

**Why:** The nightly jobs make several hundred serial REST calls (posts-index pagination, plus a comments page-set and a reactions call per bot-answered issue). A single 429, 5xx, or secondary-rate-limit 403 aborts the entire run with zero partial progress — and both index builders are all-or-nothing.

**Fix:** Mirror `modelsRequest`'s backoff handling in `ghFetch`.

.github/bot-scripts/docsIndex.cjs
```diff
@@ -0,0 +1,158 @@
...
+// Hybrid retrieval over the pre-built docs embeddings index:
+// dense embeddings catch paraphrased questions, BM25 catches exact
+// tokens like API names and error codes. Both rankings are fused
+// via Reciprocal Rank Fusion.
+
+/**
+ * @param {number[]} a
+ * @param {number[]} b
+ */
+function cosineSimilarity(a, b) {
```
robertsLando, line 81: **[Major] · Tests**

~2,900 lines of new logic ship with zero unit tests, yet the highest-risk functions here are pure and network-free: `cosineSimilarity`/`bm25Scores`/`reciprocalRankFusion` (all retrieval), `chunkMarkdown`/`slugify` (determine every doc anchor the bot links users to), `checkSuppression` (the sole guardrail against re-posting a downvoted answer), and `scoreReactions`/`parseAnswerMetadata` (decide which reactions count as a downvote).

**Why:** None of these are exercised by the nightly retrieval evals, so a regression — a `slugify` change producing dead anchors, or `checkSuppression` returning "allow" on a malformed embedding — ships undetected straight to users. Note also that `vitest.config.ts:24` (`include: ['src/**/*.test.{js,ts}', 'test/**/*.test.ts']`) does not glob `.github/bot-scripts/**`, so a test dropped there today would silently never run.

**Fix:** Add vitest specs for the listed pure functions and extend the vitest `include` globs. Prioritize `checkSuppression`, `chunkMarkdown`/`slugify`, and `scoreReactions`.

.github/workflows/docs-embeddings.yml
```diff
@@ -0,0 +1,116 @@
...
+      continue-on-error: true
```
robertsLando, line 88: **[Major] · Tests**

The evals run only on `schedule`/`workflow_dispatch` — never on `pull_request` — and the step is `continue-on-error: true`, so a hit-rate collapse below `MIN_HIT_RATE` never fails anything. It only files a tracking issue the next day.

**Why:** This PR's own retrieval logic (chunking, BM25, RRF, similarity floors) is gated by nothing before merge. The first real exposure is live user issues.

**Fix:** Run the evals against a prebuilt/fixture index in PR CI for changes touching these files, so retrieval regressions block the merge rather than surfacing a day later.

.github/bot-scripts/evalDocsAnswers.cjs
```diff
@@ -0,0 +1,97 @@
...
+// chat request per question, so use sparingly.
+
+const fs = require("node:fs/promises");
+const path = require("node:path");
+const { judgeAnswer } = require("./answerFromDocs.cjs");
+const { retrieve } = require("./docsIndex.cjs");
+const { logCase, reportResults } = require("./evalUtils.cjs");
+const { embed } = require("./modelsApi.cjs");
+
+const NUM_RESULTS = 5;
```
robertsLando, line 22: **[Major] · Tests**

All 26 eval cases across `docsAnswersEvalCases.json` (18) and `relatedPostsEvalCases.json` (8) are positive cases with an expected hit, and all currently pass at 100%.

**Why:** `MIN_SIMILARITY` and `POSTS_MIN_SIMILARITY` — the thresholds that keep the bot **silent** on off-topic questions — are exercised by zero cases. A 100% suite proves the index isn't badly broken; it proves nothing about the guardrail that stops the bot answering questions the docs don't cover.

**Fix:** Add negative cases with empty `expectedFiles`/`expectedPosts` and assert that nothing crosses the confidence floor.

.github/bot-scripts/collectDocsFeedback.cjs
```diff
@@ -0,0 +1,446 @@
...
+};
+
+/**
+ * Extracts the metadata the bot embeds in its answer comments.
+ * Comments from before the metadata tag existed fall back to
+ * parsing the doc links from the comment body.
+ * @param {string} body
+ * @returns {{style: string, confidence: number | null, sections: string[]}}
+ */
+function parseAnswerMetadata(body) {
```
robertsLando, line 111: **[Minor] · Security**

`parseAnswerMetadata`'s non-greedy regex matches the **first** `<!-- DOCS_ANSWER_METADATA {...} -->` in the comment body, not the trailer.

**Why:** An injected answer (see the Blocker on `answerFromDocs.cjs:275`) can plant a metadata comment ahead of the real one, forging `style`/`sections`/`confidence` into the public digest issue and into the suppression list.

**Fix:** Anchor the match to the trailer (last occurrence) and validate the parsed shape.

.github/bot-scripts/updateEvalTrackingIssue.cjs
```diff
@@ -0,0 +1,84 @@
...
+	const { github, context } = param;
+
+	const title = process.env.TRACKING_ISSUE_TITLE;
+	if (!title) {
+		throw new Error("TRACKING_ISSUE_TITLE is not set");
+	}
+	const failed = process.env.EVAL_OUTCOME === "failure";
+
+	// Find an existing tracking issue, open or closed
+	const { data: issues } = await github.rest.issues.listForRepo({
```
robertsLando: **[Minor] · Operability**

`listForRepo` is unpaginated — one page of 100, filtered by creator `github-actions[bot]`.

**Why:** Once more than 100 bot-authored issues/PRs exist, the tracking issue falls off page 1 and a **new duplicate is opened every night**. `updateDocsFeedbackIssue.cjs` already solves exactly this with a label + `ghPaginated`.

**Fix:** Mirror the label-based lookup used by the feedback digest.

.github/bot-scripts/updateEvalTrackingIssue.cjs
```diff
@@ -0,0 +1,84 @@
...
+		creator: "github-actions[bot]",
+		state: "all",
+		per_page: 100,
+	});
+	const tracking = issues.find((i) => i.title === title);
+
+	const runUrl =
+		`${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
+	const body =
+		`The daily evaluation of the ${process.env.EVAL_NAME} did not meet the required hit rate.
```
robertsLando: **[Minor] · Operability**

The issue body always claims the eval "did not meet the required hit rate", but `EVAL_OUTCOME=failure` also covers crashes (missing token, Models 5xx exhausting retries before `reportResults` runs).

**Why:** An infra outage gets misdiagnosed as a quality regression at 3am, and renders as "(no summary available)". False alarms erode trust in the tracking issue.

**Fix:** Have the eval `main()` distinguish an infra error from a threshold miss and report them differently.

.github/bot-scripts/updateDocsFeedbackIssue.cjs
```diff
@@ -0,0 +1,225 @@
...
+function renderEvalCase(record) {
+	const evalCase = {
+		question: record.question.slice(0, MAX_EVAL_QUESTION_LENGTH),
+		expectedFiles: [
+			...new Set(record.sections.map((s) => s.split("#")[0])),
+		],
+	};
+	return `<details><summary>Suggested eval case</summary>
+
+\`\`\`json
```
robertsLando: **[Minor] · Correctness**

`record.question` is pasted inside a ```json fence, but `cleanQuestion` preserves code fences — a question containing ``` breaks out and mangles the digest. Separately, the digest body is unbounded (180-day window, one `<details>` block per good record).

**Why:** GitHub's issue-body limit is 65536 chars; exceeding it fails the PATCH with a 422 and breaks the nightly job.

**Fix:** Strip/escape fences in the rendered question, and cap the number of records rendered.

.github/bot-scripts/collectDocsFeedback.cjs
```diff
@@ -0,0 +1,446 @@
...
+	}
+	return DEFAULT_WEIGHT;
+}
+
+/**
+ * Turns raw reactions into weighted votes and a net score
+ * @param {{user: string, content: string}[]} reactions
+ * @param {string} postAuthor
+ */
+function scoreReactions(reactions, postAuthor) {
```
robertsLando, line 181: **[Minor] · Correctness**

`scoreReactions` has no per-user dedup — it iterates raw reactions.

**Why:** One user reacting 👍 + ❤️ + 🚀 to the same comment counts their weight three times, skewing the score that drives both the suppression list and the feedback digest.

**Fix:** Collapse to one vote per user.

.github/bot-scripts/evalRelatedPosts.cjs
```diff
@@ -0,0 +1,104 @@
...
+		console.log(
+			`⏭️ ${
+				c.question.split("\n")[0]
+			} - no expected post in the index anymore, skipping`,
+		);
+		return false;
+	});
+
+	// A single batched request embeds all eval questions at once
+	const embeddings = await embed(
```
robertsLando, line 69: **[Minor] · Correctness**

If every eval case is skipped (e.g. all expected posts aged out of the index window), `embed([])` is called → API 400 → exit 1 → `EVAL_OUTCOME=failure`. Relatedly, `evalUtils.cjs:37` computes `0/0 = NaN`, and `NaN < minHitRate` is `false`, so a zero-case run would otherwise report a pass.

**Why:** Both ends of the zero-case path are wrong: one crashes and files a false regression issue, the other silently passes.

**Fix:** Short-circuit on zero cases with an explicit error.

.github/bot-scripts/buildDocsIndex.cjs
```diff
@@ -0,0 +1,259 @@
...
+
+	const pushChunk = () => {
+		const text = currentLines.join("\n").trim();
+		if (text.length >= MIN_CHUNK_LENGTH) {
+			for (const part of splitLongText(text)) {
+				chunks.push({
+					file,
+					anchor: currentAnchor,
+					title: currentTitle,
+					breadcrumbs: headingStack.map((h) => h.title),
```
robertsLando: **[Minor] · Correctness**

Content appearing before the first heading yields `breadcrumbs: []`.

**Why:** The chunk then renders as an empty link label — `- [](url)` — in the bot's comment. Latent today (no docs file currently trips the length floor), but it's a live footgun the moment one does.

**Fix:** Fall back to `chunk.title` when the breadcrumb stack is empty.

.github/bot-scripts/updateDocsFeedbackIssue.cjs
```diff
@@ -0,0 +1,225 @@
...
+const { ghPaginated, ghRequest } = require("./githubApi.cjs");
+
+const ISSUE_TITLE = "📊 Docs answer bot feedback";
+const ISSUE_MARKER = "<!-- DOCS_FEEDBACK_ISSUE -->";
+// zwave-js-ui has no dedicated "docs-feedback" label, so the existing
+// "documentation" label is reused to narrow the digest issue lookup to a
+// handful of candidates (searching by creator alone could exceed a single
+// page of results). The hidden marker above still uniquely identifies
+// the digest issue among any other issues carrying this label.
+const ISSUE_LABEL = "documentation";
```
robertsLando: **[Minor] · Design**

`ISSUE_LABEL = "documentation"` reuses the human-facing label.

**Why:** The bot's auto-generated feedback digest now shows up in every real docs-issue filter and search, mixing machine output into a triage queue humans rely on.

**Fix:** A dedicated `docs-feedback` label would keep the two separate.

.github/bot-scripts/docsIndex.cjs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/**
+ * Computes BM25 scores for all chunks against the query
+ * @param {{text: string, breadcrumbs: string[]}[]} chunks
+ * @param {string[]} queryTokens
+ * @returns {number[]}
+ */
+function bm25Scores(chunks, queryTokens) {
+	const k1 = 1.2;
```
robertsLando, line 136: **[Minor] · Readability**

BM25's `k1 = 1.2` and `b = 0.75`, and the RRF smoothing constant `60` at line 108, are bare literals.

**Why:** Every other tunable in this PR gets a named constant with rationale (`MIN_SIMILARITY`, `SUPPRESS_SIMILARITY`, …). These three break that pattern and are easy to miss when someone revisits retrieval quality.

**Fix:** Hoist to `BM25_K1`/`BM25_B`/`RRF_K` with a one-line note that they're the standard defaults.

.github/bot-scripts/evalDocsAnswers.cjs
robertsLando, line 22: **[Minor] · DRY**

The retrieval top-K value `5` is declared three times under two names: `MAX_RETRIEVED_CHUNKS` (`answerFromDocs.cjs:23`), and `NUM_RESULTS` here and in `evalRelatedPosts.cjs:17`.

**Why:** The evals exist to catch regressions in *production* retrieval. If someone tunes `MAX_RETRIEVED_CHUNKS` the evals silently keep measuring the old K, so they stop reflecting what users actually get.

**Fix:** Export the production constant and have the evals import it.

robertsLando:
### Follow-up (non-blocking): can users ask the bot a question on demand?

Reading through the triggers, the docs bot is purely event-driven — `zwave-js-bot_post.yml` fires on `issues: [opened, edited]` and `discussion: [created, edited]` only, and `zwave-js-bot_comment.yml` implements just `fix lint` / `rebase` / `rename commit`, all PR-only and behind `checkAuthorized`. So there's no `issue_comment` path into `answerFromDocs`.

Practically that means the bot gets exactly one shot per post: it answers when the issue is opened, and there's no way to ask a follow-up, re-ask after clarifying the question, or summon it on an existing thread. Given that the retrieval quality is genuinely good (18/18 and 8/8 hit@5), that feels like leaving value on the table — a lot of support questions get *sharper* after the first round of back-and-forth, which is exactly when the bot is currently silent.

Is an `@zwave-js-bot ask <question>` command something you considered upstream and rejected, or just out of scope for this port?

I can see two reasons to be cautious, so I'm asking rather than proposing:

- **The `alreadyAnswered` guard exists precisely to stop repeat posting**, so an on-demand path would need to bypass it deliberately rather than by accident — and that guard is already fragile (single-page comment scan, no `concurrency:` group; see the inline threads).
- **Cost and abuse**: an explicit command is anonymous-triggerable and burns an embedding + chat request per invocation against the 15 req/min quota, so it'd want an authorization gate or a per-thread rate limit that the `opened`-only trigger gets for free.

Happy to leave this as a separate issue if you'd rather keep this PR a faithful port — just wanted to flag it while the design is fresh.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Fixed in 554f21c0 and 2d74596f. Model output is now treated as untrusted: HTML and images are removed with complete delimiter handling, links and mentions are neutralized, and only trusted links are appended separately.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Fixed in 554f21c0. Parsed model responses now require a finite confidence, a string answer, and a string-array relatedExcerpts before they can influence posting.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Fixed in 554f21c0 and eb9278dd. The script derives exclusions from authorizedUsers.cjs and includes the bot; the redundant workflow-level user list was removed.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Fixed in 554f21c0. docsIndex.cjs now owns the index version and validates the cached version, model, chunk shape, and embedding vectors at every load site.

.github/workflows/docs-embeddings.yml
AlCalzone: Fixed in eb9278dd. The docs cache namespace was bumped and its key now hashes the docs plus the indexing and model-related scripts.

.github/workflows/zwave-js-bot_post.yml
AlCalzone: Fixed in eb9278dd. Workflow concurrency is now scoped by event type and post number, with cancel-in-progress enabled.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Fixed in 554f21c0. Existing-answer detection now paginates issue and discussion comments.

.github/workflows/zwave-js-bot_post.yml
AlCalzone: We decided not to add a runtime kill switch. Disabling the workflow remains the emergency control, and missing live indexes now emit explicit warnings rather than failing unrelated jobs.

.github/workflows/docs-embeddings.yml
AlCalzone: Fixed in eb9278dd. Scheduled jobs now fail when an expected index is absent; live post events warn and skip because a cache miss should not fail the user-facing event workflow.

.github/workflows/posts-embeddings.yml
AlCalzone: Fixed in eb9278dd. The scheduled posts job now fails when the expected index is absent.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Fixed in 554f21c0. Models API failures now propagate; only JSON parsing and response-shape failures are handled as invalid model output.

.github/bot-scripts/collectDocsFeedback.cjs
AlCalzone: Fixed in 554f21c0. Suppression embeddings now use embedBatched.

.github/bot-scripts/updatePostsIndex.cjs
AlCalzone: We are keeping full JSON index cache rewrites for this repository-sized corpus. Binary vectors and delta caches add storage and merge complexity that is not justified by the current index size.

.github/workflows/zwave-js-bot_post.yml
AlCalzone: We are keeping answer generation and index updates as separate jobs. This preserves independent permissions and failure handling; the duplicate embedding request is an accepted quota tradeoff.

.github/bot-scripts/githubApi.cjs
AlCalzone: Fixed in 554f21c0. REST and GraphQL helpers now retry bounded 429, secondary-limit 403, GraphQL rate-limit, and 5xx failures using Retry-After, reset headers, or exponential backoff.

.github/bot-scripts/docsIndex.cjs
AlCalzone: Fixed in db924f6d. Vitest now discovers bot tests, with deterministic coverage for retrieval, chunking, suppression, reaction scoring, metadata, API retries, sanitization, and digest rendering.

.github/workflows/docs-embeddings.yml
AlCalzone: Covered in db924f6d with deterministic PR tests. Live embedding/model evaluations remain nightly because PR runs should not depend on model availability, quota, or cache state.

.github/bot-scripts/evalDocsAnswers.cjs
AlCalzone: Covered in db924f6d with deterministic silence and malformed-response tests around the production confidence gates. We decided against live-model negative fixtures because model variability would make PR checks flaky.

.github/bot-scripts/collectDocsFeedback.cjs
AlCalzone: Fixed in 554f21c0. Public answer text is sanitized, and metadata parsing now selects and validates the final trailer record.

.github/bot-scripts/updateEvalTrackingIssue.cjs
AlCalzone: Fixed in 554f21c0. Tracking-issue lookup now paginates and distinguishes infrastructure failures from quality regressions.

.github/bot-scripts/updateEvalTrackingIssue.cjs
AlCalzone: Fixed in 554f21c0. Evaluation scripts now publish an explicit infrastructure-error outcome separately from a threshold regression.

.github/bot-scripts/updateDocsFeedbackIssue.cjs
AlCalzone: Fixed in 554f21c0. Dynamic code fences prevent breakout, and digest rendering preserves a bounded priority prefix below the issue-body limit.

.github/bot-scripts/collectDocsFeedback.cjs
AlCalzone: Fixed in 554f21c0. Reactions are collapsed to one net vote per user before weighting.

.github/bot-scripts/evalRelatedPosts.cjs
AlCalzone: Fixed in 554f21c0. Zero eligible cases now fail explicitly before embedding, and shared result reporting rejects zero-case runs.

.github/bot-scripts/buildDocsIndex.cjs
AlCalzone: Fixed in 554f21c0. Pre-heading content now falls back to the chunk title for its breadcrumb.

.github/bot-scripts/updateDocsFeedbackIssue.cjs
AlCalzone: Fixed in 554f21c0. The digest now uses the dedicated docs-feedback label, which has been created in this repository.

.github/bot-scripts/docsIndex.cjs
AlCalzone: We are leaving the standard BM25 and RRF constants local to their formulas. Naming these literals would not change behavior or reduce duplication.

.github/bot-scripts/evalDocsAnswers.cjs
AlCalzone: We are keeping the docs and related-post retrieval limits local. They tune different retrieval pipelines and do not need to remain coupled.

AlCalzone:
Addressed the unanchored findings as follows:

- Job timeouts were added in eb9278dd.
- We are not adding a subsystem README in this PR; the scripts remain small standalone workflow entry points.
- We are keeping comment composition in the answer orchestrator because extracting it would add indirection without shared behavior.
- We fixed tracking lookup pagination and outcome reporting in 554f21c0, but are not unifying the feedback and evaluation issue updaters because they use different execution contexts and lifecycle rules.

All inline findings have an individual reply with the fixing commit or agreed disposition.

AlCalzone:
Followup questions are currently not intended due to possibility for abuse. If it looks like we need them, we can reconsider.

.github/bot-scripts/sanitizeAnswer.cjs
```diff
@@ -0,0 +1,108 @@
...
+// in its output is, by definition, unexpected and is stripped down to
+// plain text rather than rendered as a clickable link.
+
+/**
+ * Inserts a zero-width space after each @ so GitHub does not resolve
+ * the text into a user/team mention notification
+ * @param {string} text
+ */
+function neutralizeMentions(text) {
+	return text.replace(/@/g, "@\u200b");
```
robertsLando: **[Major · Security]** Residual on the sanitization Blocker. `neutralizeMentions` only breaks literal ASCII `@`. Entity-encoded forms — `&#64;octocat`, `&#x40;team`, `&commat;here` — pass through all five sanitizer passes untouched, and GitHub decodes numeric/named character references to `@` in text nodes *after* markdown→HTML, where its mention filter then fires. Same anonymous-triggerable path as the original Blocker, so a prompt-injected answer emitting `&#64;`-encoded handles can still mass-ping.

**Fix:** decode (or neutralize) numeric + named char-refs for `@` before the mention pass. Rendering-dependent, so worth a quick live confirm, but the code path is unambiguous.

.github/bot-scripts/sanitizeAnswer.cjs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/**
+ * Removes HTML comments and tags, repeating the tag strip to avoid
+ * leaving partial tags behind (e.g. from nested/malformed markup)
+ * @param {string} text
+ */
+function stripHtml(text) {
```
robertsLando, line 141: **[Major · Correctness]** `stripHtml` removes everything between `<` and `>` unconditionally, including inside inline/fenced code: `Array<string>`→`Array`, `Map<string, number>`→`Map`, `` `x < 5 && y > 3` ``→`` `x  3` ``. For a docs bot whose answers routinely contain generics, comparisons and shell redirects, this mangles exactly the technical answers it exists to post.

Fail-safe security-wise, but a real quality regression introduced by the fix. **Fix:** skip stripping inside fenced/inline code spans.

.github/bot-scripts/githubApi.cjs
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * @param {Response} response
+ * @param {number} attempt
+ */
+function getRateLimitDelayMs(response, attempt) {
+	if (attempt >= MAX_RETRIES) return undefined;
+
+	const retryAfter = Number(response.headers.get("retry-after"));
+	if (Number.isFinite(retryAfter) && retryAfter > 0) {
+		return retryAfter * 1000;
```
robertsLando: **[Minor · Operability]** An explicit `Retry-After` value is honored uncapped (`retryAfter * 1000`), unlike the reset and exponential paths which cap at 60s. A large/hostile value would sleep toward the job timeout. `Math.min(retryAfter * 1000, 60_000)` keeps it bounded like the others.

robertsLando (commented):
## Reconciliation re-review at `2d74596f`

Thanks for the thorough remediation — this addresses the prior review point-by-point. Verified each fix against the tree with per-cluster passes:

- **16 of 18 findings verified fixed** — confidence validation, `EXCLUDED_USERS` derivation, comment/discussion pagination, error propagation, REST+GraphQL retry/backoff, reaction dedup, tracking-issue pagination, infra-vs-quality outcomes, digest fence-escaping + bounding, workflow concurrency, and scheduled-job fail-on-missing-index all confirmed present and correct. I've resolved those threads.
- **2 test findings partially landed** — 108 deterministic tests now gate PRs (great), but two gaps remain; see the still-open test threads.
- **All 5 WONTFIX dispositions accepted** — kill switch (mitigated by warn-on-missing-index, confirmed), full-index rewrite, duplicate embedding, BM25/RRF literals, triple-declared top-K. Reasonable calls for this repo's scale.

Remaining items are line-anchored below and on the three still-open threads. None block merge on their own, but the **entity-encoded mention residual is worth closing before treating the sanitizer as the security boundary**.

.github/bot-scripts/answerFromDocs.cjs
robertsLando: Re-reviewed at `2d74596f`: the core fix is verified — model output is now sanitized on the **posted** comment path (not just the digest), and HTML/image/link stripping + trusted-links-only all hold. Two residuals remain on the model-output path, so keeping this open: (1) entity-encoded `@mentions` (`&#64;`/`&#x40;`/`&commat;`) bypass `neutralizeMentions`; (2) `stripHtml` corrupts inline/fenced code. Details in new comments on `sanitizeAnswer.cjs`.

.github/bot-scripts/docsIndex.cjs
robertsLando: Re-reviewed at `2d74596f`: substantial deterministic tests landed and now gate PRs (108 passing) — suppression, sanitization, reactions, metadata, digest and API-retry are all well covered. One residual keeps this open: `bm25Scores` and `reciprocalRankFusion` (the two highest-risk retrieval primitives) are not exported and have no direct test — the single indirect `retrieve` test passes even with fusion removed (dense and lexical agree on that case). A disagree-case fusion test would close it; needs exporting the two helpers.

.github/bot-scripts/evalDocsAnswers.cjs
robertsLando: Re-reviewed at `2d74596f`: malformed-response→no-answer and the suppression similarity gate are now genuinely covered. Still open: the off-topic silence gate itself — `MIN_SIMILARITY` / `POSTS_MIN_SIMILARITY` — has no below-threshold→no-post case (the gates live in non-exported functions). That is the specific concern here; a single silence test asserting no post below threshold would close it.

.github/bot-scripts/sanitizeAnswer.cjs
AlCalzone: Fixed in b5d5e6a7. Numeric, hexadecimal, and named encoded @ forms are neutralized outside code; GitHub’s Markdown renderer confirms they no longer produce mention links.

.github/bot-scripts/sanitizeAnswer.cjs
AlCalzone: Fixed in b5d5e6a7. Valid inline and fenced code is preserved, while malformed delimiters and content outside code remain sanitized.

.github/bot-scripts/githubApi.cjs
AlCalzone: Fixed in b5d5e6a7. Explicit Retry-After delays are now capped at 60 seconds like the reset-header and exponential-backoff paths.

robertsLando (approved):

PR #4753, opened 2026-07-21
fix(bot): keep the answer bot working when its index cache is evicted

robertsLando:
## Problem

The answer bot from #4743 has never posted an answer. Its docs and posts embeddings indexes live only in the Actions cache, which is capped at 10 GB per repository and evicted least-recently-used once full. At 1–2 MB apiece those indexes lose that race constantly.

The only run since the feature landed — [29531097992](https://github.com/zwave-js/zwave-js-ui/actions/runs/29531097992), discussion 4749 — restored nothing:

```
Cache not found for input keys: docs-embeddings-v2-4d1349a..., docs-embeddings-v2-
Cache not found for input keys: posts-embeddings-v1-, posts-embeddings-v1-
Cache not found for input keys: docs-feedback-v1-, docs-feedback-v1-
→ "Answer from documentation"  skipped
→ "Add post to index"          skipped
✓ run reported success
```

All three had been written by green workflow runs ten hours earlier that same morning.

It isn't a scoping or key problem. The nightly `docs-embeddings` run missed the cache **it had saved itself the previous morning, under the byte-identical key**:

```
2026-07-17T05:58:52Z  Cache not found for input keys: docs-embeddings-v2-4d1349a..., docs-embeddings-v2-
2026-07-17T05:59:11Z  Cache saved with key:           docs-embeddings-v2-4d1349a...
```

Today not a single embeddings or feedback cache written before Jul 18 survives.

And the failure mode was invisible: a missing index produced `core.warning` plus a skipped step, and the run went green. A bot that has never worked looked exactly like a healthy one.

## Changes

**Both embeddings workflows upload their index as an artifact** (30 day retention). Artifacts are not LRU-evicted, so they are the dependable source. `include-hidden-files: true` is required and load-bearing — `.docs-index` / `.posts-index` are dot-directories, and the upload glob skips hidden entries by default including the search root, so without it the upload matches zero files and produces no artifact while still reporting success. `if-no-files-found: error` makes that guarantee enforced rather than advisory.

**New composite action `.github/actions/restore-bot-index`**, used at three call sites. It takes the index name and a token:

```yaml
- uses: ./.github/actions/restore-bot-index
  with:
    index: docs
    github-token: ${{ github.token }}
```

It restores from the cache, then falls back to the artifact. Two details that matter:

- It selects the **newest unexpired artifact by name**, not the newest run with a successful conclusion. The index is produced by `build-index` alone, so an unrelated `evaluate`/`feedback` failure would otherwise hide a perfectly good artifact.
- Provenance is pinned to the default branch of this repository (`head_repository_id == repository_id`). A fork PR's `head_branch` can also be `master`, so a branch-name check alone would not be a trust boundary.

The fallback decides on a readable `index.json`, not on the cache key — a prefix match can restore an entry saved while its build was still writing — and clears the directory before extracting so the two sources cannot mix. The lookup is `continue-on-error`, so a transient API blip degrades to "no index" instead of taking down a job that someone triggered by opening an issue.

**Failures now reach a maintainer.** These runs are triggered by whoever opened the issue or discussion, so a red run notifies *them*. An index outage now opens the same kind of tracking issue the nightly evaluations already use, and a later healthy run closes it again. `updateEvalTrackingIssue.cjs` gains two optional env knobs (`TRACKING_ISSUE_BODY`, `TRACKING_ISSUE_QUIET`) for callers that aren't a daily eval and that fire once per user post; both default to today's behaviour.

**Loudness now matches severity.** `answer-from-docs` still fails hard when no index can be restored — that outage is user-visible and total. `update-posts-index` warns and skips instead: its own message says the post is picked up by the next nightly rebuild, so failing it would paint every community post red for something that self-corrects. A partial outage (one index of two) warns rather than passing silently.

**Staleness is checked.** A restored index older than `max-age-days` (default 3) warns, so a broken nightly surfaces before the artifact expires rather than the day after.

The downvote suppression list keeps cache-only restore with no fallback: a missing one just means no suppression is applied, which is already the documented behaviour.

## Verification

- Reproduced the hidden-directory upload bug directly against `@actions/glob`: `excludeHiddenFiles=true → []`, `false → [.docs-index, .docs-index/index.json]`.
- Exercised the artifact-selection jq against the live API — returns empty when no artifact exists, returns the right run id when one does.
- `updateEvalTrackingIssue.cjs` had no test coverage; this PR adds `updateEvalTrackingIssue.test.cjs` (8 cases) covering both new knobs, the recovery-close path, and that the existing eval behaviour is unchanged. Full bot-scripts suite: 125 tests, 9 files.

## Notes

Merging touches both `docs-embeddings.yml` and `posts-embeddings.yml`, which are in their own workflows' push path filters, so both rebuild and upload their first artifact on merge. There is a ~15–20 minute window before that completes during which a new issue would find no artifact; combined with an evicted cache, `answer-from-docs` would fail once.

The incremental `update-posts-index` job also heals itself now: on an evicted cache it restores from the artifact, appends, and re-saves the cache, instead of waiting for the nightly rebuild.

The cache pressure that causes the eviction is addressed separately in #4752 — 10.2 GB of the repository's 10.7 GB sits on PR refs where nothing can ever read it. This PR makes the bot survive eviction regardless.

### DIFF
--- a/.github/actions/restore-bot-index/action.yml
+++ b/.github/actions/restore-bot-index/action.yml
@@ -1,92 +1,179 @@
 name: Restore bot index
 description: >-
-  Restore an embeddings index directory (containing index.json) from the Actions
-  cache, falling back to the artifact uploaded by the workflow that builds it.
+  Restore the docs or posts embeddings index from the Actions cache, falling
+  back to the artifact uploaded by the workflow that builds it.
 
 # The Actions cache is capped per repository and evicted least-recently-used, so
 # a 1-2 MB index regularly disappears within hours of being written. Artifacts
-# are not evicted, so they are what makes the restore actually dependable.
+# are not evicted, so they are what makes the restore dependable.
+#
+# Each index lives in a fixed directory (.docs-index / .posts-index) holding a
+# single index.json, and is published under a fixed artifact name, so the caller
+# only has to say which one it wants.
 
 inputs:
-  path:
-    description: Directory to restore the index into
-    required: true
-  cache-key:
-    description: Primary Actions cache key
-    required: true
-  cache-restore-keys:
-    description: Newline separated cache key prefixes to fall back to
-    required: false
-    default: ''
-  artifact-name:
-    description: Artifact uploaded by the building workflow
-    required: true
-  workflow:
-    description: Workflow file that builds and uploads the artifact
+  index:
+    description: Which index to restore - "docs" or "posts"
     required: true
   github-token:
-    description: Token with actions read scope, used to list runs and download the artifact
+    description: Token with actions read scope, used to find and download the artifact
     required: true
+  max-age-days:
+    description: Warn when the restored index is older than this many days
+    required: false
+    default: '3'
 
 outputs:
   found:
     description: '"true" when the index is available afterwards'
-    value: ${{ steps.check.outputs.found }}
+    value: ${{ steps.report.outputs.found }}
 
 runs:
   using: composite
 
   steps:
-    - name: Restore from cache
-      id: cache
+    - name: Resolve the index
+      id: cfg
+      shell: bash
+      env:
+        INDEX: ${{ inputs.index }}
+      run: |
+        set -euo pipefail
+        case "$INDEX" in
+          docs|posts) ;;
+          *) echo "::error::Unknown index '$INDEX' - expected 'docs' or 'posts'"; exit 1 ;;
+        esac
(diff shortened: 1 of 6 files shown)


robertsLando:
cc @AlCalzone

AlCalzone:
I wonder why this works in my repo without this 🤔

robertsLando:
@AlCalzone the problem in my repo is the high use of cache due to docker images build cache

AlCalzone:
Ohhhh. I see...

PR #4757, opened 2026-07-27
fix(docker): take Node 24 from the Alpine base instead of the node: image

### LINKED ISSUE #4755: Docker build of 11.22.0 still Node v22.20.0
### Checklist

### Deploy method

Docker

### Z-Wave JS UI version

11.22.0

### ZwaveJS version

15.26.0

### Describe the bug

The docker build of 11.22.0 is still using node v22.20.0 even after b9204f20760b4d504b25e9cc69a6b59dddb4938f

```bash
$ podman exec zwavejsui node -v
v22.20.0
```

https://github.com/zwave-js/zwave-js-ui/actions/runs/29827730059/job/88633694759#step:10:309

### To Reproduce

`podman exec zwavejsui node -v`

### Expected behavior

Returns v24.xyz

### Additional context

I want to use [NODE_USE_ENV_PROXY support](https://nodejs.org/learn/http/enterprise-network-configuration)

robertsLando:
Reworked from scratch — the original approach in this PR (bump the tag, source the version from `.nvmrc`) can't work, see below. Rebased onto current master.

## Problem

The published image still ships Node 22.20.0 (#4755). `Dockerfile.debian` and `Dockerfile.contrib` are already on Node 24 — Renovate bumped those — but `docker/Dockerfile`, the one `docker-release.yml` builds for the `master` and `latest` tags, is still pinned to `node:22.20.0-alpine3.22`.

It can't just be bumped, because Node.js dropped 32-bit ARM binaries in v24:

```
v22.23.2 -> linux-arm64, linux-armv7l, linux-x64, linux-ppc64le, linux-s390x
v24.21.0 -> linux-arm64,               linux-x64, linux-ppc64le, linux-s390x
```

which shows up directly in the official images:

```
node:22-alpine3.22 -> amd64, armv6, armv7, arm64/v8, s390x
node:24-alpine3.22 -> amd64,               arm64/v8, s390x
node:24-alpine3.23 -> amd64,               arm64/v8, s390x
```

`docker-release.yml` publishes `linux/amd64, linux/arm64, linux/arm/v7, linux/arm/v6`, so pointing `NODE_IMAGE` at 24 breaks both 32-bit ARM legs. That's what 2c157f5 hit when armv6 was restored to the platform list — Node went back to 22.20.0 in the same commit.

The `.nvmrc` idea in the original commit compounds it: `.nvmrc` is `24.20.0` today, and `node:24.20.0-alpine3.22` isn't a published tag at all.

## Fix

Alpine builds Node itself for 32-bit ARM, so take Node from the base image's own apk packages instead of standing the official binaries on a bare alpine base:

```
alpine v3.23, main/nodejs   24.18.1-r0 -> armhf (armv6), armv7, aarch64, x86_64
alpine v3.23, community/npm 11.11.0-r0 -> same four
```

The `FROM ${NODE_IMAGE} AS node` stage and its four `COPY --from=node` lines are gone, replaced by `nodejs` and `npm` in the `apk add` that was already there. One less stage, and Node 24 on every architecture we publish.

`nodejs-dev` is added to the build dependencies of both build stages. The `node:` image supplied node's headers at `/usr/local/include`, so without it node-gyp goes to nodejs.org for them on every architecture — the first CI run of this branch failed on exactly that, `ECONNRESET` fetching `node-v24.18.1-headers.tar.gz` on the arm/v7 leg. With `nodejs-dev` installed node-gyp finds them locally and makes no HTTP request at all; the existing `apk del .build-dependencies` takes it back out.

## Known limits, documented in the files

- **Patch level follows Alpine, not nodejs.org.** 3.23 ships 24.18.1; `.nvmrc` says 24.20.0. They agree on major only. Comment at the top of `docker/Dockerfile`.
- **Moving `ALPINE_IMAGE` can move the Node major.** 3.22 ships Node 22, 3.23 ships Node 24, and nothing in the image tag says so. The Renovate rule that grouped `node` + `alpine` only existed to keep the two tag suffixes in sync, so it's removed; the alpine major/minor `dependencyDashboardApproval` gate stays, with its reasoning rewritten to the new one.
- **`package_armv7.yml` stays on `arm32v7/node:22-bullseye`.** Same upstream cause, and there is no `arm32v7/node:24-*` to move to. Comment added so nobody "fixes" it.
- **`engines.node`: `>= 24.20.0` -> `>= 24.18.1`**, so the image doesn't violate our own floor. The pinned patch came from Renovate's node preset in #4812; nothing needs 24.20 specifically.

## Also in here

`docker-test.yml` was building the host platform only, three times: the matrix key is `platforms` but the step reads `matrix.platform`, so the value was empty and the arm legs tested nothing. Renamed the key and added `linux/arm/v6` so CI covers the same set the release publishes. Happy to split this out if you'd rather, but without it CI can't actually verify this change.

## Verified locally

- amd64 image: `node -v` -> `v24.18.1`, `@serialport/bindings-cpp` loads, app boots and listens on 8091.
- `linux/arm/v7` image built under qemu with `--no-cache`: builds clean, `npm rebuild @serialport/bindings-cpp` compiles, zero `gyp http GET` lines in the log, `node -v` -> `v24.18.1`, `process.arch` -> `arm`, binding loads.
- `alpine:3.23` under qemu on `linux/arm/v6` and `linux/arm/v7`: `apk add nodejs npm` -> Node 24.18.1, npm 11.11.0.

`NODE_USE_ENV_PROXY` is available once this ships.

Closes #4755


robertsLando (commented):
Still not happy with this solution BTW...

PR #4763, opened 2026-07-28
feat(bot): compute embeddings locally instead of using GitHub Models

AlCalzone:
## Description

GitHub Models retires on July 30. This is the first of three PRs migrating off it (porting zwave-js/zwave-js#8985 to this repo), and covers all embeddings usage (docs answer bot, related-posts suggestions).

- Embeddings are computed on the runner with `@huggingface/transformers` and the `Xenova/all-MiniLM-L6-v2` model — the same model the zwave-js bots use, pinned to the same revision. Since this repo is not a monorepo, the dependency lives in a dedicated `.github/bot-scripts/package.json`, installed with `npm ci --ignore-scripts` (which also skips the `onnxruntime-node` CUDA download, see zwave-js/zwave-js#8974) by a shared `setup-bot-embeddings` composite action that owns the node_modules and model-weight caches.
- Doc chunks are re-chunked (`MAX_CHUNK_LENGTH` 4000 → 1200), sized empirically against the pipeline's 512-token window (p99 ≈ 440 tokens, 0.6% slight overflow). Texts are batched by length to cut padding waste.
- Similarity floors recalibrated for MiniLM's cosine distribution (`MIN_SIMILARITY` 0.2 → 0.35, `POSTS_MIN_SIMILARITY` 0.55 → 0.6), based on measured on-topic/off-topic separation against the rebuilt indexes.
- Index cache keys bumped (`docs-embeddings-v3`, `posts-embeddings-v2`, `docs-feedback-v2`) in the workflows and the `restore-bot-index` action; indexes built with the old model are ignored (with a workflow warning) until the nightly rebuilds replace them.
- `models: read` and `MODELS_TOKEN` removed from all embedding paths. The chat-based answer judge still uses the Models API and is migrated in the follow-up PRs.

## Validation

- Retrieval evals with the new model and chunking: docs hit@5 18/18 (100%, required 90%) at every chunk cap tried (700/1000/1200/1400), posts hit@5 7/8 (87.5%, required 80%)
- Docs index: 310 chunks; build takes a few minutes of runner CPU
- All 136 bot-script tests pass

### DIFF
--- a/.github/actions/restore-bot-index/action.yml
+++ b/.github/actions/restore-bot-index/action.yml
@@ -65,8 +65,8 @@ runs:
       uses: actions/cache/restore@v6
       with:
         path: .docs-index
-        key: docs-embeddings-v2-${{ hashFiles('docs/**/*.md', '.github/bot-scripts/buildDocsIndex.cjs', '.github/bot-scripts/docsIndex.cjs', '.github/bot-scripts/modelsApi.cjs') }}
-        restore-keys: docs-embeddings-v2-
+        key: docs-embeddings-v3-${{ hashFiles('docs/**/*.md', '.github/bot-scripts/buildDocsIndex.cjs', '.github/bot-scripts/docsIndex.cjs', '.github/bot-scripts/localEmbeddings.cjs') }}
+        restore-keys: docs-embeddings-v3-
 
     - name: Restore the posts index from cache
       if: inputs.index == 'posts'
@@ -75,8 +75,8 @@ runs:
         path: .posts-index
         # There is no exact key to match, the newest index is picked
         # via the prefix
-        key: posts-embeddings-v1-
-        restore-keys: posts-embeddings-v1-
+        key: posts-embeddings-v2-
+        restore-keys: posts-embeddings-v2-
 
     # Decide on the file, not on the cache key: a prefix match can restore an
     # entry that was saved while its build was still writing it
--- a/.github/actions/restore-bot-index/action.yml
+++ b/.github/actions/restore-bot-index/action.yml
@@ -65,7 +65,7 @@ runs:
       uses: actions/cache/restore@v6
       with:
         path: .docs-index
-        key: docs-embeddings-v3-${{ hashFiles('docs/**/*.md', '.github/bot-scripts/buildDocsIndex.cjs', '.github/bot-scripts/docsIndex.cjs', '.github/bot-scripts/localEmbeddings.cjs') }}
+        key: docs-embeddings-v3-${{ hashFiles('docs/**/*.md', '.github/bot-scripts/buildDocsIndex.cjs', '.github/bot-scripts/docsIndex.cjs', '.github/bot-scripts/localEmbeddings.cjs', '.github/bot-scripts/package-lock.json') }}
         restore-keys: docs-embeddings-v3-
 
     - name: Restore the posts index from cache
--- a/.github/bot-scripts/answerFromDocs.cjs
+++ b/.github/bot-scripts/answerFromDocs.cjs
@@ -7,7 +7,8 @@ const { authorizedUsers } = require("./authorizedUsers.cjs");
 const { cosineSimilarity, loadDocsIndex, retrieve } = require(
 	"./docsIndex.cjs",
 );
-const { CHAT_MODEL, embed, modelsRequest } = require("./modelsApi.cjs");
+const { EMBEDDING_MODEL, embed } = require("./localEmbeddings.cjs");
+const { CHAT_MODEL, modelsRequest } = require("./modelsApi.cjs");
 const {
 	QUESTION_CATEGORY_SLUGS,
 	cleanQuestion,
@@ -32,14 +33,18 @@ const MAX_RETRIEVED_CHUNKS = 5;
 // If not even the best dense match reaches this cosine similarity,
 // the post is considered off-topic and no chat request is made.
 // The real relevance judgment is left to the chat model.
-const MIN_SIMILARITY = 0.2;
+// Calibrated for all-MiniLM-L6-v2: on-topic questions score >= 0.36
+// against the docs index, clearly off-topic ones <= 0.17
+const MIN_SIMILARITY = 0.35;
 // Confidence thresholds for the different response styles
 const ANSWER_CONFIDENCE = 75;
 const LINKS_CONFIDENCE = 40;
 
 // A related post is only suggested above this cosine similarity.
 // A wrong suggestion is worse than a missed one, so keep this high.
-const POSTS_MIN_SIMILARITY = 0.55;
+// Calibrated for all-MiniLM-L6-v2: confirmed-related posts score
+// 0.55-0.78 on the eval set, while unrelated ones reach up to 0.58
+const POSTS_MIN_SIMILARITY = 0.6;
 const MAX_RELATED_POSTS = 3;
 
 // Questions at least this similar to a previously downvoted answer
@@ -297,10 +302,15 @@ async function buildDocsAnswerSection(
 
 	// Ask the model whether the docs answer the question. judgeAnswer()
 	// already degrades malformed/invalid model content to a safe "no
-	// answer" result internally - a genuine Models API/network error
-	// (thrown by modelsRequest) is intentionally NOT caught here, so it
-	// fails the job instead of silently posting nothing.
-	const result = await judgeAnswer(question, ranked, token);
+	// answer" result internally. Skip the docs section on a genuine API
+	// error, the related-posts section is still useful.
+	let result;
+	try {
+		result = await judgeAnswer(question, ranked, token);
+	} catch (e) {
+		console.log(`Chat request failed: ${e.message}`);
+		return;
+	}
 	console.log("Model response:", JSON.stringify(result));
 
 	const related = (result.relatedExcerpts ?? [])
@@ -445,7 +455,8 @@ ${links}`;
  * documentation, and/or suggests similar existing posts, in a single comment.
  *
  * Expects the following environment variables:
- * - MODELS_TOKEN: token with models:read permission
+ * - MODELS_TOKEN: token with models:read permission, enables the docs
+ *   answer section. Without it, only related posts are suggested.
  * - DOCS_INDEX_PATH: path to the embeddings index created by buildDocsIndex.cjs
  * - POSTS_INDEX_PATH: path to the embeddings index created by buildPostsIndex.cjs
  *
@@ -455,10 +466,6 @@ async function main(param) {
 	const { github, context } = param;
 
 	const modelsToken = process.env.MODELS_TOKEN;
-	if (!modelsToken) {
-		console.log("No MODELS_TOKEN provided, skipping");
-		return;
-	}
 	// Figure out where the question comes from
 	const isDiscussion = !!context.payload.discussion;
 	const post = context.payload.discussion ?? context.payload.issue;
@@ -497,7 +504,7 @@ async function main(param) {
 	// Load the pre-built embeddings indices. Either may be missing,
 	// each one enables its part of the comment.
 	const docsIndexPath = process.env.DOCS_INDEX_PATH;
-	const docsIndex = await loadDocsIndex(docsIndexPath);
+	let docsIndex = await loadDocsIndex(docsIndexPath);
 	if (docsIndex) {
 		console.log(
 			`Loaded docs index with ${docsIndex.chunks.length} chunks (created ${docsIndex.createdAt})`,
@@ -517,26 +524,26 @@ async function main(param) {
 		);
 	}
 
-	if (!docsIndex && !postsIndex) return;
-
 	const question = cleanQuestion(post.title, post.body ?? "");
 
-	// A single embedding request serves both docs retrieval and
-	// related-post ranking. Similarities are only comparable within
-	// one model, so an index embedded with a different model is skipped.
-	const embeddingModel = docsIndex?.model ?? postsIndex?.model;
-	if (postsIndex && postsIndex.model !== embeddingModel) {
+	// The question is embedded locally. Similarities are only comparable
+	// within one model, so indexes built with a different model are skipped
+	// until the nightly rebuild replaces them.
+	if (docsIndex && docsIndex.model !== EMBEDDING_MODEL) {
 		console.log(
-			`Posts index model ${postsIndex.model} does not match ${embeddingModel}, ignoring it`,
+			`Docs index model ${docsIndex.model} does not match ${EMBEDDING_MODEL}, ignoring it`,
+		);
+		docsIndex = undefined;
+	}
+	if (postsIndex && postsIndex.model !== EMBEDDING_MODEL) {
+		console.log(
+			`Posts index model ${postsIndex.model} does not match ${EMBEDDING_MODEL}, ignoring it`,
 		);
 		postsIndex = undefined;
-		if (!docsIndex) return;
 	}
-	const [questionEmbedding] = await embed(
-		[question],
-		modelsToken,
-		embeddingModel,
-	);
+	if (!docsIndex && !postsIndex) return;
+
+	const [questionEmbedding] = await embed([question]);
 
 	// Feedback guardrail: check the question against previously
 	// downvoted answers collected by collectDocsFeedback.cjs
@@ -553,11 +560,12 @@ async function main(param) {
 		suppression = checkSuppression(
 			questionEmbedding,
 			feedback,
-			embeddingModel,
+			EMBEDDING_MODEL,
 		);
 	}
 
-	const docsSection = docsIndex && suppression !== "silent"
+	// The docs section needs the chat model as relevance judge
+	const docsSection = docsIndex && modelsToken && suppression !== "silent"
 		? await buildDocsAnswerSection(
 			question,
 			questionEmbedding,
--- a/.github/bot-scripts/answerFromDocs.cjs
+++ b/.github/bot-scripts/answerFromDocs.cjs
@@ -7,7 +7,9 @@ const { authorizedUsers } = require("./authorizedUsers.cjs");
 const { cosineSimilarity, loadDocsIndex, retrieve } = require(
 	"./docsIndex.cjs",
 );
-const { EMBEDDING_MODEL, embed } = require("./localEmbeddings.cjs");
+const { EMBEDDING_MODEL, embed, indexMatchesModel } = require(
+	"./localEmbeddings.cjs",
+);
 const { CHAT_MODEL, modelsRequest } = require("./modelsApi.cjs");
 const {
 	QUESTION_CATEGORY_SLUGS,
@@ -303,12 +305,15 @@ async function buildDocsAnswerSection(
 	// Ask the model whether the docs answer the question. judgeAnswer()
 	// already degrades malformed/invalid model content to a safe "no
 	// answer" result internally. Skip the docs section on a genuine API
-	// error, the related-posts section is still useful.
+	// error, the related-posts section is still useful - but annotate
+	// the run so the degradation is visible without reading the log.
 	let result;
 	try {
 		result = await judgeAnswer(question, ranked, token);
 	} catch (e) {
-		console.log(`Chat request failed: ${e.message}`);
+		console.log(
+			`::warning::Chat request failed, answering without a docs section: ${e.message}`,
+		);
 		return;
 	}
 	console.log("Model response:", JSON.stringify(result));
@@ -529,16 +534,10 @@ async function main(param) {
 	// The question is embedded locally. Similarities are only comparable
 	// within one model, so indexes built with a different model are skipped
 	// until the nightly rebuild replaces them.
-	if (docsIndex && docsIndex.model !== EMBEDDING_MODEL) {
-		console.log(
-			`Docs index model ${docsIndex.model} does not match ${EMBEDDING_MODEL}, ignoring it`,
-		);
+	if (docsIndex && !indexMatchesModel(docsIndex, "docs index")) {
 		docsIndex = undefined;
 	}
-	if (postsIndex && postsIndex.model !== EMBEDDING_MODEL) {
-		console.log(
-			`Posts index model ${postsIndex.model} does not match ${EMBEDDING_MODEL}, ignoring it`,
-		);
+	if (postsIndex && !indexMatchesModel(postsIndex, "posts index")) {
 		postsIndex = undefined;
 	}
 	if (!docsIndex && !postsIndex) return;
@@ -660,7 +659,10 @@ module.exports.alreadyAnswered = alreadyAnswered;
 module.exports.judgeAnswer = judgeAnswer;
 module.exports.validateJudgeResponse = validateJudgeResponse;
 module.exports.checkSuppression = checkSuppression;
+module.exports.buildRelatedPostsSection = buildRelatedPostsSection;
 module.exports.chunkUrl = chunkUrl;
+module.exports.MIN_SIMILARITY = MIN_SIMILARITY;
+module.exports.POSTS_MIN_SIMILARITY = POSTS_MIN_SIMILARITY;
 module.exports.DOCS_ANSWER_COMMENT_TAG = DOCS_ANSWER_COMMENT_TAG;
 module.exports.DOCS_ANSWER_METADATA_TAG = DOCS_ANSWER_METADATA_TAG;
 module.exports.DOCS_ANSWER_METADATA_VERSION = DOCS_ANSWER_METADATA_VERSION;
--- a/.github/bot-scripts/buildDocsIndex.cjs
+++ b/.github/bot-scripts/buildDocsIndex.cjs
@@ -13,12 +13,13 @@ const { EMBEDDING_MODEL, embedBatched } = require("./localEmbeddings.cjs");
 // Chunks shorter than this are unlikely to contain useful information
 const MIN_CHUNK_LENGTH = 50;
 // Sections longer than this are sub-split so nothing gets truncated away.
-// The model is trained on short passages and its reference implementation
-// caps input at 256 word pieces (the tokenizer itself accepts 512), so
-// chunks are sized for that budget - code tokenizes at ~2-3 chars/token
-const MAX_CHUNK_LENGTH = 700;
+// Sized empirically against the pipeline's real 512-token window: at
+// 1200 chars, p99 of the chunks measures ~440 tokens and only 0.6%
+// exceed the window slightly. That is past the model's 256-word-piece
+// training length, but retrieval quality holds (eval hit@5 18/18).
+const MAX_CHUNK_LENGTH = 1200;
 // Overlap between sub-splits so answers spanning a split boundary aren't lost
-const CHUNK_OVERLAP = 150;
+const CHUNK_OVERLAP = 200;
 
 /**
  * Removes HTML tags, repeating to avoid leaving partial tags behind
--- a/.github/bot-scripts/buildDocsIndex.cjs
+++ b/.github/bot-scripts/buildDocsIndex.cjs
@@ -1,22 +1,24 @@
 // @ts-check
 
 // Builds a semantic search index over the documentation by chunking all
-// markdown files and embedding the chunks using GitHub Models.
+// markdown files and embedding the chunks with a local model.
 // Usage: node buildDocsIndex.cjs <docs-dir> <output-file>
-// Requires GITHUB_TOKEN with models:read permission.
 
 const crypto = require("node:crypto");
 const fs = require("node:fs/promises");
 const path = require("node:path");
 const { DOCS_INDEX_VERSION } = require("./docsIndex.cjs");
-const { embedBatched, EMBEDDING_MODEL } = require("./modelsApi.cjs");
+const { EMBEDDING_MODEL, embedBatched } = require("./localEmbeddings.cjs");
 
 // Chunks shorter than this are unlikely to contain useful information
 const MIN_CHUNK_LENGTH = 50;
-// Sections longer than this are sub-split so nothing gets truncated away
-const MAX_CHUNK_LENGTH = 4000;
+// Sections longer than this are sub-split so nothing gets truncated away.
+// The embedding model truncates input at 256 tokens, and code blocks
+// tokenize at ~2-3 characters/token, so oversized chunks would lose
+// their tails to truncation
+const MAX_CHUNK_LENGTH = 700;
 // Overlap between sub-splits so answers spanning a split boundary aren't lost
-const CHUNK_OVERLAP = 400;
+const CHUNK_OVERLAP = 150;
 
 /**
  * Removes HTML tags, repeating to avoid leaving partial tags behind
@@ -184,12 +186,6 @@ async function main() {
 		);
 		process.exit(1);
 	}
-	const token = process.env.GITHUB_TOKEN;
-	if (!token) {
-		console.error("GITHUB_TOKEN environment variable is required");
-		process.exit(1);
-	}
-
 	// Reuse embeddings from a previous index for unchanged chunks
 	/** @type {Map<string, number[]>} */
 	const previousEmbeddings = new Map();
@@ -241,7 +237,6 @@ async function main() {
 
 	const embeddings = await embedBatched(
 		pending.map((c) => c.embeddedText),
-		token,
 	);
 	pending.forEach((chunk, i) => chunk.embedding = embeddings[i]);
 
--- a/.github/bot-scripts/buildDocsIndex.cjs
+++ b/.github/bot-scripts/buildDocsIndex.cjs
@@ -13,9 +13,9 @@ const { EMBEDDING_MODEL, embedBatched } = require("./localEmbeddings.cjs");
 // Chunks shorter than this are unlikely to contain useful information
 const MIN_CHUNK_LENGTH = 50;
 // Sections longer than this are sub-split so nothing gets truncated away.
-// The embedding model truncates input at 256 tokens, and code blocks
-// tokenize at ~2-3 characters/token, so oversized chunks would lose
-// their tails to truncation
+// The model is trained on short passages and its reference implementation
+// caps input at 256 word pieces (the tokenizer itself accepts 512), so
+// chunks are sized for that budget - code tokenizes at ~2-3 chars/token
 const MAX_CHUNK_LENGTH = 700;
 // Overlap between sub-splits so answers spanning a split boundary aren't lost
 const CHUNK_OVERLAP = 150;
--- a/.github/bot-scripts/buildDocsIndex.cjs
+++ b/.github/bot-scripts/buildDocsIndex.cjs
@@ -13,10 +13,10 @@ const { EMBEDDING_MODEL, embedBatched } = require("./localEmbeddings.cjs");
 // Chunks shorter than this are unlikely to contain useful information
 const MIN_CHUNK_LENGTH = 50;
 // Sections longer than this are sub-split so nothing gets truncated away.
-// Sized empirically against the pipeline's real 512-token window: at
-// 1200 chars, p99 of the chunks measures ~440 tokens and only 0.6%
-// exceed the window slightly. That is past the model's 256-word-piece
-// training length, but retrieval quality holds (eval hit@5 18/18).
+// Sized for the 512-token window the tokenizer enforces: at 1200 chars the
+// median chunk measures 132 tokens and 0.6% exceed the window. The 256 word
+// pieces on the model card are sentence-transformers' inference default,
+// which this ONNX port does not ship.
 const MAX_CHUNK_LENGTH = 1200;
 // Overlap between sub-splits so answers spanning a split boundary aren't lost
 const CHUNK_OVERLAP = 200;
--- a/.github/bot-scripts/localEmbeddings.cjs
+++ b/.github/bot-scripts/localEmbeddings.cjs
@@ -79,14 +79,24 @@ async function embed(inputs) {
 	if (inputs.length === 0) return [];
 	const extractor = await getExtractor();
 
+	// Each batch is padded to its longest member, so batch texts of
+	// similar length together and restore the input order afterwards
+	const order = inputs
+		.map((_, i) => i)
+		.sort((a, b) => inputs[a].length - inputs[b].length);
+
 	/** @type {number[][]} */
-	const results = [];
-	for (let i = 0; i < inputs.length; i += BATCH_SIZE) {
-		const batch = inputs.slice(i, i + BATCH_SIZE);
-		const output = await extractor(batch, {
-			pooling: "mean",
-			normalize: true,
-		});
+	const results = new Array(inputs.length);
+	let done = 0;
+	for (let i = 0; i < order.length; i += BATCH_SIZE) {
+		const batchOrder = order.slice(i, i + BATCH_SIZE);
+		const output = await extractor(
+			batchOrder.map((j) => inputs[j]),
+			{
+				pooling: "mean",
+				normalize: true,
+			},
+		);
 		// dims is [batch, hidden]; anything else means the pipeline
 		// changed shape, and flattening it blindly would push corrupted
 		// vectors into the index
@@ -95,11 +105,13 @@ async function embed(inputs) {
 				`Unexpected embedding tensor shape [${output.dims}]`,
 			);
 		}
-		results.push(...output.tolist());
+		const vectors = output.tolist();
+		batchOrder.forEach((j, k) => results[j] = vectors[k]);
+		done += batchOrder.length;
 		// Index builds run for minutes; leave evidence of progress in
 		// case the job hits its timeout
 		if (inputs.length > BATCH_SIZE) {
-			console.log(`Embedded ${results.length}/${inputs.length} texts`);
+			console.log(`Embedded ${done}/${inputs.length} texts`);
 		}
 	}
 	return results;
--- a/.github/bot-scripts/localEmbeddings.cjs
+++ b/.github/bot-scripts/localEmbeddings.cjs
@@ -11,6 +11,15 @@ const { join } = require("node:path");
 // quality stay comparable between the two bots
 const EMBEDDING_MODEL = "Xenova/all-MiniLM-L6-v2";
 const MODEL_REVISION = "751bff37182d3f1213fa05d7196b954e230abad9";
+// q8 quantization quarters the download and memory footprint; the
+// retrieval evals showed no quality loss against the full weights
+const MODEL_DTYPE = "q8";
+
+// CI cache key for the model weights: changes exactly when different
+// weights would be downloaded, so unrelated edits to this file don't
+// evict the cache. Consumed by .github/actions/setup-bot-embeddings.
+const MODEL_CACHE_KEY = `${EMBEDDING_MODEL}@${MODEL_REVISION}@${MODEL_DTYPE}`
+	.replaceAll("/", "_");
 
 // Process a bounded number of texts per pipeline call to limit peak memory
 const BATCH_SIZE = 32;
@@ -26,17 +35,38 @@ function defaultModelCacheDir() {
 /** @type {Promise<any> | undefined} */
 let extractorPromise;
 
+async function createExtractor() {
+	// The package is ESM-only, hence the dynamic import
+	const { pipeline } = await import("@huggingface/transformers");
+	// A cache miss downloads the weights from huggingface.co. Retry with
+	// backoff so a hiccup there does not fail a user-triggered run.
+	let lastError;
+	for (let attempt = 1; attempt <= 3; attempt++) {
+		try {
+			return await pipeline("feature-extraction", EMBEDDING_MODEL, {
+				cache_dir: process.env.BOT_MODEL_CACHE_DIR?.trim()
+					|| defaultModelCacheDir(),
+				revision: MODEL_REVISION,
+				dtype: MODEL_DTYPE,
+			});
+		} catch (e) {
+			lastError = e;
+			console.log(
+				`::warning::Loading the embedding model failed (attempt ${attempt}/3): ${e.message}`,
+			);
+			await new Promise((resolve) =>
+				setTimeout(resolve, attempt * 5000)
+			);
+		}
+	}
+	throw new Error(
+		`Could not load the embedding model ${EMBEDDING_MODEL}: ${lastError.message}`,
+		{ cause: lastError },
+	);
+}
+
 function getExtractor() {
-	extractorPromise ??= (async () => {
-		// The package is ESM-only, hence the dynamic import
-		const { pipeline } = await import("@huggingface/transformers");
-		return pipeline("feature-extraction", EMBEDDING_MODEL, {
-			cache_dir: process.env.BOT_MODEL_CACHE_DIR?.trim()
-				|| defaultModelCacheDir(),
-			revision: MODEL_REVISION,
-			dtype: "q8",
-		});
-	})();
+	extractorPromise ??= createExtractor();
 	return extractorPromise;
 }
 
@@ -57,8 +87,20 @@ async function embed(inputs) {
 			pooling: "mean",
 			normalize: true,
 		});
-		const data = output.tolist();
-		results.push(...(Array.isArray(data[0]) ? data : [data]));
+		// dims is [batch, hidden]; anything else means the pipeline
+		// changed shape, and flattening it blindly would push corrupted
+		// vectors into the index
+		if (output.dims?.length !== 2) {
+			throw new Error(
+				`Unexpected embedding tensor shape [${output.dims}]`,
+			);
+		}
+		results.push(...output.tolist());
+		// Index builds run for minutes; leave evidence of progress in
+		// case the job hits its timeout
+		if (inputs.length > BATCH_SIZE) {
+			console.log(`Embedded ${results.length}/${inputs.length} texts`);
+		}
 	}
 	return results;
 }
@@ -77,8 +119,27 @@ async function embedBatched(texts) {
 	);
 }
 
+/**
+ * Checks that an index was embedded with the pinned model, so its
+ * similarities are comparable to freshly embedded questions. Emits a
+ * workflow warning on mismatch, so a skipped index shows up in the run
+ * annotations instead of only in the log.
+ * @param {{model?: string} | undefined} index
+ * @param {string} description
+ */
+function indexMatchesModel(index, description) {
+	if (!index) return false;
+	if (index.model === EMBEDDING_MODEL) return true;
+	console.log(
+		`::warning::The ${description} was created with ${index.model}, not ${EMBEDDING_MODEL} - ignoring it until it is rebuilt`,
+	);
+	return false;
+}
+
 module.exports = {
 	embed,
 	embedBatched,
+	indexMatchesModel,
 	EMBEDDING_MODEL,
+	MODEL_CACHE_KEY,
 };
--- a/.github/bot-scripts/localEmbeddings.cjs
+++ b/.github/bot-scripts/localEmbeddings.cjs
@@ -1,0 +1,84 @@
+// @ts-check
+
+// Local embedding pipeline running on the CI runner, replacing the
+// retired GitHub Models embeddings API
+
+const { homedir } = require("node:os");
+const { join } = require("node:path");
+
+// Keep model, revision and dtype in sync with the zwave-js repository
+// (.github/bot-scripts/localEmbeddings.cjs there), so evals and answer
+// quality stay comparable between the two bots
+const EMBEDDING_MODEL = "Xenova/all-MiniLM-L6-v2";
+const MODEL_REVISION = "751bff37182d3f1213fa05d7196b954e230abad9";
+
+// Process a bounded number of texts per pipeline call to limit peak memory
+const BATCH_SIZE = 32;
+
+function defaultModelCacheDir() {
+	const xdg = process.env.XDG_CACHE_HOME;
+	const base = xdg && xdg.trim().length > 0
+		? xdg
+		: join(homedir(), ".cache");
+	return join(base, "zwave-js-ui-bot", "models");
+}
+
+/** @type {Promise<any> | undefined} */
+let extractorPromise;
+
+function getExtractor() {
+	extractorPromise ??= (async () => {
+		// The package is ESM-only, hence the dynamic import
+		const { pipeline } = await import("@huggingface/transformers");
+		return pipeline("feature-extraction", EMBEDDING_MODEL, {
+			cache_dir: process.env.BOT_MODEL_CACHE_DIR?.trim()
+				|| defaultModelCacheDir(),
+			revision: MODEL_REVISION,
+			dtype: "q8",
+		});
+	})();
+	return extractorPromise;
+}
+
+/**
+ * Embeds one or more texts, returning the embeddings in input order
+ * @param {string[]} inputs
+ * @returns {Promise<number[][]>}
+ */
+async function embed(inputs) {
+	if (inputs.length === 0) return [];
+	const extractor = await getExtractor();
+
+	/** @type {number[][]} */
+	const results = [];
+	for (let i = 0; i < inputs.length; i += BATCH_SIZE) {
+		const batch = inputs.slice(i, i + BATCH_SIZE);
+		const output = await extractor(batch, {
+			pooling: "mean",
+			normalize: true,
+		});
+		const data = output.tolist();
+		results.push(...(Array.isArray(data[0]) ? data : [data]));
+	}
+	return results;
+}
+
+/**
+ * Embeds texts destined for an index
+ * @param {string[]} texts
+ * @returns {Promise<number[][]>} Embeddings in input order
+ */
+async function embedBatched(texts) {
+	console.log(`Embedding ${texts.length} texts locally...`);
+	const embeddings = await embed(texts);
+	// Round to reduce index size, this has no measurable impact on similarity
+	return embeddings.map((embedding) =>
+		embedding.map((x) => Math.round(x * 1e5) / 1e5)
+	);
+}
+
+module.exports = {
+	embed,
+	embedBatched,
+	EMBEDDING_MODEL,
+};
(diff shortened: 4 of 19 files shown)


.github/workflows/docs-embeddings.yml
```diff
@@ -13,7 +13,7 @@ on:
       - 'docs/**/*.md'
       - '.github/bot-scripts/buildDocsIndex.cjs'
       - '.github/bot-scripts/docsIndex.cjs'
-      - '.github/bot-scripts/modelsApi.cjs'
+      - '.github/bot-scripts/localEmbeddings.cjs'
       - '.github/bot-scripts/githubApi.cjs'
       - '.github/bot-scripts/evalDocsAnswers.cjs'
```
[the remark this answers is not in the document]
AlCalzone: Fixed in b86aae92 — both manifests are now push trigger paths, so a dependency bump gets a validating run immediately instead of at the next nightly.

.github/actions/restore-bot-index/action.yml
```diff
@@ -65,8 +65,8 @@ runs:
       uses: actions/cache/restore@v6
       with:
         path: .docs-index
-        key: docs-embeddings-v2-${{ hashFiles('docs/**/*.md', '.github/bot-scripts/buildDocsIndex.cjs', '.github/bot-scripts/docsIndex.cjs', '.github/bot-scripts/modelsApi.cjs') }}
-        restore-keys: docs-embeddings-v2-
+        key: docs-embeddings-v3-${{ hashFiles('docs/**/*.md', '.github/bot-scripts/buildDocsIndex.cjs', '.github/bot-scripts/docsIndex.cjs', '.github/bot-scripts/localEmbeddings.cjs') }}
+        restore-keys: docs-embeddings-v3-
```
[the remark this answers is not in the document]
AlCalzone: Intentionally omitted: the embedding output is pinned by model id, revision and dtype in localEmbeddings.cjs, which is part of the key. A dependency bump does not change embeddings, and the index reuses embeddings per content hash anyway, so keying on the manifests would only churn the caches.

.github/workflows/docs-embeddings.yml
```diff
@@ -42,16 +41,34 @@ jobs:
...
+      if: steps.cache.outputs.cache-hit != 'true'
+      working-directory: .github/bot-scripts
+      run: npm ci --ignore-scripts
+
+    - name: Restore embedding model
+      if: steps.cache.outputs.cache-hit != 'true'
+      uses: actions/cache@v6
+      with:
+        path: ~/.cache/zwave-js-ui-bot/models
+        key: embedding-model-${{ hashFiles('.github/bot-scripts/localEmbeddings.cjs') }}
```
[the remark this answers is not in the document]
AlCalzone: actions/cache saves in a `post-if: success()` step, so nothing is saved when the job fails (the deprecated `save-always` input existed precisely because of that default). A partial model download fails the build step, which fails the job, so no corrupt cache entry can be persisted.

.github/workflows/posts-embeddings.yml
```diff
@@ -49,9 +48,26 @@ jobs:
...
+
+    - name: Install embedding dependencies
+      working-directory: .github/bot-scripts
+      run: npm ci --ignore-scripts
+
+    - name: Restore embedding model
+      uses: actions/cache@v6
+      with:
+        path: ~/.cache/zwave-js-ui-bot/models
+        key: embedding-model-${{ hashFiles('.github/bot-scripts/localEmbeddings.cjs') }}
```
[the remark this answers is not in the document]
AlCalzone: Same as above: the cache save step is gated on `post-if: success()`, a failed job never saves.

.github/workflows/zwave-js-bot_post.yml
```diff
@@ -164,6 +164,23 @@ jobs:
...
+
+    - name: Install embedding dependencies
+      working-directory: .github/bot-scripts
+      run: npm ci --ignore-scripts
+
+    - name: Restore embedding model
+      uses: actions/cache@v6
+      with:
+        path: ~/.cache/zwave-js-ui-bot/models
+        key: embedding-model-${{ hashFiles('.github/bot-scripts/localEmbeddings.cjs') }}
```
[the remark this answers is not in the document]
AlCalzone: Same as above: the cache save step is gated on `post-if: success()`, a failed job never saves.

robertsLando (dismissed):
considering the high dupliucation there is between zwave-js and ui could we create a repo where we can share some of this files/constants so a change on that is ported automatically on both?

robertsLando:
Direction looks right to me — no vendor, no token, no rate limit, and nothing that can be retired out from under us again. Two things worth considering, neither blocking.

## Embedding model choice

The 256-token window on `all-MiniLM-L6-v2` is what forces `MAX_CHUNK_LENGTH` 4000 → 700, and that re-chunking is the biggest quality cost in this PR — a section that used to be one chunk is now ~6, so retrieval can't see a whole section at once anymore. The model itself is also from 2021.

All of these are ONNX + transformers.js, so they're drop-in for the same `pipeline("feature-extraction", …)` call:

| model | ctx | dims | note |
|---|---|---|---|
| `Xenova/all-MiniLM-L6-v2` | 256 | 384 | current |
| `Xenova/bge-small-en-v1.5` | 512 | 384 | same size and speed, 2× the window |
| `Xenova/gte-small` | 512 | 384 | same class |
| `nomic-ai/nomic-embed-text-v1.5` | 8192 | 768 (Matryoshka) | no re-chunking needed at all, ~274 MB |
| `onnx-community/embeddinggemma-300m-ONNX` | 2048 | 768 | best multilingual <500M on MTEB, ~622 MB |

`bge-small-en-v1.5` looks like the free swap — identical cost profile, double the window. `nomic-embed-text-v1.5` or EmbeddingGemma would let us keep 4000-char chunks and drop the re-chunking change entirely, at the cost of a bigger one-time weights download (cached across runs) and slower CPU inference.

Fully aware the real constraint isn't technical: `localEmbeddings.cjs` pins model + revision + dtype deliberately to match zwave-js, so this is a decision for both repos or neither. Raising it here mostly so it's on the record before the indexes get rebuilt — if it's worth changing, changing it now is cheaper than after.

## Threshold margins

Independent of the model: `POSTS_MIN_SIMILARITY = 0.6` with the comment noting unrelated posts reach 0.58 is a 0.02 margin, and it also cuts off real matches in the 0.55–0.6 band. Hand-tuned absolute cosine floors are going to need re-measuring every time the docs change shape.

A cross-encoder reranker over the top ~20 dense hits would replace the floor with an actual relevance score, and it's cheap because you only score a handful of candidates. `Xenova/ms-marco-MiniLM-L-6-v2`, `Xenova/bge-reranker-base` and `mixedbread-ai/mxbai-rerank-xsmall-v1` are all ONNX/transformers.js. Probably more retrieval quality per line of code than any embedder swap. Follow-up material, not for this PR.

## Copilot CLI for the chat judge

For the two follow-up PRs: Copilot CLI can now run in Actions on the built-in `GITHUB_TOKEN` — no PAT, no stored secret, just `copilot-requests: write`. That landed on 2026-07-02, one day after the Models retirement notice, and it maps onto the `openai/gpt-4o` relevance judge fairly directly.

https://github.blog/changelog/2026-07-02-copilot-cli-no-longer-needs-a-personal-access-token-in-github-actions/

Caveats before anyone gets too excited:

- It needs the org policy *"Allow use of Copilot CLI billed to the organization"*, i.e. it is **billed to the org** per request, where Models was free. That's the actual question, not a detail.
- Non-interactive use requires `--yolo`, and GitHub's own docs warn that invoking Copilot CLI directly in workflow steps "gives it broad access to your workflow environment". `zwave-js-bot_post` runs with `issues: write`, so that wants fencing.
- It's an agent CLI, not a completions endpoint. `judgeAnswer()` wants one structured JSON verdict; we'd be wrapping a shell agent to get it.

Worth noting for completeness that Copilot is a non-starter for *this* PR either way — the Copilot REST API is admin surface only (seat management, usage metrics, content exclusion, agent config), with no inference or embeddings endpoint. The `/v1/embeddings` implementations floating around are third-party proxies replaying an editor OAuth token, which isn't something to put in public CI.

AlCalzone:
> considering the high dupliucation there is between zwave-js and ui could we create a repo where we can share some of this files/constants so a change on that is ported automatically on both?

Makes sense. Lets do this as a followup.

> Xenova/bge-small-en-v1.5

is an option. The larger models take far too long to run on CPU.

> localEmbeddings.cjs pins model + revision + dtype deliberately to match zwave-js, so this is a decision for both repos or neither.

Cache is separate, so there is no definitive need to align.

> Copilot CLI for the chat judge

Not possible for reasons I can't discuss here.

> A cross-encoder reranker over the top ~20 dense hits would replace the floor with an actual relevance score, and it's cheap because you only score a handful of candidates.

Sounds interesting - we should investigate this as a followup

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -517,26 +524,26 @@ async function main(param) {
...
 		);
 		postsIndex = undefined;
-		if (!docsIndex) return;
 	}
-	const [questionEmbedding] = await embed(
-		[question],
-		modelsToken,
-		embeddingModel,
-	);
+	if (!docsIndex && !postsIndex) return;
```
robertsLando, line 543: **[Blocker] · Operability**

An index restored with the previous embedding model is discarded with a bare `console.log` and `main()` returns, while `restore-bot-index` has already reported `found=true` / `stale=false` and `report-index-status` has closed the tracking issue as healthy.

**Why:** For the entire degraded window — and on any future model bump — the bot posts nothing on every new issue, every run is green, the tracking issue says healthy, and the "Fail if neither index is available" guard (written specifically to stop a permanently broken bot reporting green) never fires. There is no 3am signal at all.

**Fix:** Make the embedding model an index *health* property rather than a script-internal skip — have the action's jq check compare `.model` against an expected value passed in, so `found`/`stale` and the tracking issue reflect it. At minimum, emit `::warning::` instead of `console.log` on the skip.

.github/workflows/zwave-js-bot_post.yml
```diff
@@ -164,6 +164,23 @@ jobs:
...
+        cache: 'npm'
+        cache-dependency-path: .github/bot-scripts/package-lock.json
+
+    - name: Install embedding dependencies
+      working-directory: .github/bot-scripts
+      run: npm ci --ignore-scripts
+
+    - name: Restore embedding model
+      uses: actions/cache@v6
+      with:
```
robertsLando: **[Major] · Security**

`answer-from-docs` now `npm ci`s 74 third-party packages including `onnxruntime-node`'s native `.node`/`libonnxruntime.so`, and loads them in the same Node process that holds `secrets.BOT_TOKEN` — `answerFromDocs.cjs` requires `localEmbeddings.cjs`, which `import()`s `@huggingface/transformers` inside the `actions/github-script` step whose `github-token:` is the PAT.

**Why:** Before this PR the job's only third-party code was `fetch`. A compromised tarball or substituted ONNX blob executes native code with a write-scoped PAT in memory, on a workflow any user can trigger by opening an issue.

**Fix:** Compute the question embedding in a plain `run:` step with no secrets in env and hand the vector to the github-script step.

.github/bot-scripts/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "zwave-js-ui-bot-scripts",
+  "private": true,
+  "description": "Runtime dependencies for the bot scripts, installed separately (npm ci --ignore-scripts in this directory) so the bot jobs don't pay for a full application install",
```
robertsLando, line 7: **[Major] · Security**

`--ignore-scripts` is advertised as the hardening measure but does not cover what matters. It blocks `onnxruntime-node`'s and `sharp`'s install scripts, but the CPU binary ships inside the `onnxruntime-node` tarball (`bin/napi-v6/linux/x64/onnxruntime_binding.node`, verified against an unpacked 1.24.x copy) and is `dlopen`'d at `require` time regardless.

**Why:** Zero protection against a malicious published version, while reading as though the risk is mitigated.

**Fix:** Rely on the lockfile pin plus process isolation, and drop the claim so it is not read as a mitigation.

.github/workflows/docs-embeddings.yml
```diff
@@ -13,7 +13,9 @@ on:
       - 'docs/**/*.md'
       - '.github/bot-scripts/buildDocsIndex.cjs'
       - '.github/bot-scripts/docsIndex.cjs'
-      - '.github/bot-scripts/modelsApi.cjs'
+      - '.github/bot-scripts/localEmbeddings.cjs'
```
robertsLando, line 16: **[Major] · DRY**

The `paths:` trigger now lists `package.json`/`package-lock.json`, but the `docs-embeddings-v3-` cache key's `hashFiles(...)` set at :46 and :111 does not, and neither does restore-bot-index/action.yml:68.

**Why:** A `@huggingface/transformers` bump that changes embedding output triggers the workflow, hits the stale key, and skips the rebuild — the index silently keeps old-model vectors. The same input set is spelled out in three places and has already drifted once.

**Fix:** Make the key's file list match the trigger's, and derive both from one place.

.github/workflows/zwave-js-bot_post.yml
robertsLando: **[Major] · Performance/Operability**

The model cache key is `hashFiles('localEmbeddings.cjs')` with no `restore-keys:`, so any edit to that file — a comment, `BATCH_SIZE`, a JSDoc fix — evicts the model in all 7 jobs and forces a ~23 MB q8 ONNX + tokenizer re-download from huggingface.co inside the user-facing answer job. Same pattern at :250 and 6 further sites.

**Why:** Cold-start regression plus a hard runtime dependency on huggingface.co availability on the interactive path — the opposite of what "compute locally" is meant to buy.

**Fix:** Key on model identity (`EMBEDDING_MODEL` + revision + dtype, e.g. a tiny pinned sidecar), and add a `restore-keys:` prefix so a near-miss still seeds the cache dir.

.github/workflows/docs-embeddings.yml
```diff
@@ -42,16 +43,34 @@ jobs:
...
+      if: steps.cache.outputs.cache-hit != 'true'
+      working-directory: .github/bot-scripts
+      run: npm ci --ignore-scripts
+
+    - name: Restore embedding model
+      if: steps.cache.outputs.cache-hit != 'true'
+      uses: actions/cache@v6
+      with:
+        path: ~/.cache/zwave-js-ui-bot/models
+        key: embedding-model-${{ hashFiles('.github/bot-scripts/localEmbeddings.cjs') }}
```
robertsLando: **[Major] · DRY**

The identical 3-step block (setup-node 22 + `cache-dependency-path` + `npm ci --ignore-scripts` + model cache) is pasted 7 times — docs-embeddings.yml :53-69, :123-137, :172-186; posts-embeddings.yml :58-72, :131-145; zwave-js-bot_post.yml :168-182, :234-250.

**Why:** The repo already establishes the opposite convention — shared bot boilerplate lives in `.github/actions/`, and that action's own header comment argues a string used by multiple workflows must live in one place. A node bump or cache-key change now needs 7 edits, and a partial edit silently degrades to a full model re-download instead of failing.

**Fix:** Extract a `.github/actions/setup-bot-embeddings` composite that owns the model cache path and key.

robertsLando (commented):
# Code Review: feat(bot): compute embeddings locally instead of using GitHub Models

**41 findings** — 1 Blocker, 17 Major, 16 Minor, 6 Nit, 1 FYI. 31 posted as inline line comments; 10 moved into this body because their lines fall outside the PR diff.

Head: `314106a9` · Base: `master` (merge-base `9643f6e9`) · 8/8 review lanes ran

## Summary

Verdict: **Needs work**. Scope: 15 files, +331/−202 LOC (`package-lock.json` excluded from review).

### Top 3 risks

1. A stale old-model index restores as "healthy", is then silently discarded, and the bot posts nothing on every issue while every run stays green and the tracking issue reports healthy — the exact "permanently broken bot reporting green run after run" scenario the existing guard was written to prevent.
2. The cache-key bump is incomplete across its three surfaces (cache key, artifact name, `paths:` trigger), so the degraded window can persist past the first nightly rebuild rather than self-healing.
3. `MAX_QUESTION_LENGTH` stayed at 6000 chars while the embedding window shrank to 512 tokens, so ~60% of every long post is silently dropped from its index vector — and `POSTS_MIN_SIMILARITY` was raised on top of the degraded vectors.

### Strengths

- Reuses the existing `embed`/`embedBatched`/`EMBEDDING_MODEL` seam, so call sites are one-line import swaps rather than a new abstraction.
- Lazy `import()` in `getExtractor()` means the six bot jobs that never embed still load `index.cjs` without needing `node_modules`.
- `modelsApi.cjs` narrowed to chat-only rather than deleted — exactly the shape the stacked PRs need.
- Thresholds documented with the measured distributions they were calibrated against, not just asserted.

### Themes

- **Health signals don't know about the model.** Index validity is now a two-part predicate (schema *and* embedding model), but every health mechanism only understands the first part.
- **Loud failure quietly became silent degradation.** The `judgeAnswer` try/catch reverses a documented deliberate choice, and its own JSDoc now contradicts the behaviour.
- **One logical change, spelled out in 7-8 places.** Setup block, model cache key, node version, and the model-equality check are each copy-pasted.

### Open question to settle before acting on two findings

A live read of `tokenizer_config.json` at pinned revision `751bff37` reports `model_max_length: 512`, while the comment at `buildDocsIndex.cjs:16` claims the model truncates at 256 tokens. Both the chunk-size and `MAX_QUESTION_LENGTH` findings depend on which is correct.

### Coverage

All 8 specialist lanes ran (Correctness, Security, DRY, Tests, Performance, Design/API, Operability, Readability). `package-lock.json` excluded by request (~1030 of the 1361 added lines). Review ran against a detached worktree at PR head `314106a9` with full filesystem context, so repo-wide DRY/codebase-fit checks were not degraded.

### Verified as fine (recorded so they are not re-raised)

`splitLongText` arithmetic is sound — no infinite loop, no dropped tail, no oversized part, 1 MB input in 3ms. Mean pooling and `tolist()` return `[1, 384]` even for a single input, so there is no malformed-vector path. `embed([]) -> []` avoids invoking the extractor on an empty batch. Cross-dimension mixing is closed on every path. Threshold operators use `>=` so NaN is excluded. The model is pinned by full commit SHA and HF LFS pointers make that content-binding. Every lockfile entry carries an `integrity` hash. Post-`models: read` permissions are minimal across all six jobs. Cache/artifact poisoning is not reachable — none of the three workflows use `pull_request`/`pull_request_target`, and the artifact fallback requires `head_repository_id == repository_id`. `POSTS_MIN_SIMILARITY = 0.6` is consistent with its own comment (unrelated posts reach 0.58), so the precision-over-recall choice is deliberate, not a defect.

### Out of scope, separate pass

`zwave-js-bot_post.yml:49` interpolates `${{ github.event.issue.body }}` straight into the `prompt:` of `actions/ai-inference`, and `check-issue-repo` has no `permissions:` block so the repo default applies. Pre-existing, not introduced here.

## Findings outside the diff

These 10 findings could not be anchored as inline comments — their target lines are not part of this PR's diff — so they are recorded here in full and are just as actionable.

### F4 · Major · Correctness

`.github/bot-scripts/postsIndex.cjs:18` _(not anchorable: line(s) outside every hunk)_

`MAX_QUESTION_LENGTH` stayed at 6000 chars while the embedding window shrank from 8191 tokens to 512. A 6000-char body tokenizes to ~1322 tokens and `truncation: true` cuts it to 512, so ~60% of every long post is silently dropped from its index vector. `cleanQuestion`'s 8-head/4-tail code-block trimming was budgeted for 6000 chars, so a couple of log blocks can eat the whole window before any prose.

**Why:** `POSTS_MIN_SIMILARITY` was raised to 0.6 on top of the degraded vectors, compounding the recall loss.

**Fix:** Budget post text to the real window, or chunk long posts the way docs are chunked.

### F5 · Major · Design/API

`.github/actions/restore-bot-index/action.yml:120-160` _(not anchorable: line(s) outside every hunk)_

Cache keys were bumped (`docs-embeddings-v3`, `posts-embeddings-v2`) but the artifact fallback still selects by the unchanged names `docs-index` / `posts-index` with 30-day retention.

**Why:** After merge and before the first nightly rebuild, the newest default-branch artifact is an old-model index. The action's only content gate is `((.chunks // .posts)|length) > 0`, so it reports healthy, and the failure lands downstream where nothing is watching. The cache-key bump does not actually isolate old-model data.

**Fix:** Make the version part of the artifact identity (name, or a `.model`/`.version` assertion in the jq check) so a stale-model index reports as not-found/stale.

### F6 · Major · Design/API

`.github/bot-scripts/docsIndex.cjs:48-60` _(not anchorable: file not in diff)_

`loadDocsIndex`/`loadPostsIndex` validate version, shape, and per-vector finiteness, but not model or vector length; the model check is instead copy-pasted into five callers. Same applies at postsIndex.cjs:90-100.

**Why:** The one guard that makes this migration safe is opt-in per caller, and nothing checks dimensionality — a future revision or dtype change under the same model string would load and silently mix vector spaces.

**Fix:** Move the model equality check plus an expected-dimension check into the loaders so callers cannot forget it.

### F16 · Major · Performance

`.github/actions/restore-bot-index/action.yml:6` _(not anchorable: line(s) outside every hunk)_

The action's own comment says the 1-2 MB index "regularly disappears within hours" under the repo's LRU cache cap; this PR adds a ~100 MB npm entry plus a ~24 MB model entry, written from 7 jobs.

**Why:** ~50-100x the index's footprint now competes for the same quota, so the index is evicted sooner and the bot falls through to the slower artifact path more often — which is also the path carrying the stale-model risk in F5.

**Fix:** Budget the cache — restore-only in consumer jobs, a single designated writer.

### F18 · Major · Readability

`.github/bot-scripts/answerFromDocs.cjs:116-123` _(not anchorable: line(s) outside every hunk)_

`validateJudgeResponse`'s JSDoc still states that a genuine Models API/network error "should" fail the whole job, while this same diff makes the caller swallow it.

**Why:** A maintainer reading the contract will believe API failures fail CI loudly when they now silently downgrade to "related posts only" — the documentation asserts the opposite of the behaviour.

**Fix:** Update the comment to match the new caller semantics.

### F21 · Minor · Correctness

`.github/bot-scripts/buildDocsIndex.cjs:85` _(not anchorable: line(s) outside every hunk)_

Sub-parts from `splitLongText` bypass the `MIN_CHUNK_LENGTH` gate, which is only applied pre-split. `"a".repeat(60) + " ".repeat(1400) + "b".repeat(60)` yields `["a"*60, "", "b"*60]`; the empty part gets embedded (breadcrumbs only) and passes `isValidChunk`, since an empty string produces a valid non-zero normalized vector.

**Why:** It becomes a retrieval candidate that renders no text. Current docs do not hit it (397 chunks, 0 empty/short), so this is latent.

**Fix:** Filter parts below `MIN_CHUNK_LENGTH` after splitting.

### F24 · Minor · Design/API

`.github/bot-scripts/docsIndex.cjs:12` _(not anchorable: file not in diff)_

`DOCS_INDEX_VERSION`/`POSTS_INDEX_VERSION` (postsIndex.cjs:12) stay at 1 while chunking (4000 to 700, overlap 400 to 150) and `cleanQuestion` both changed incompatibly, contradicting the constants' own "bump whenever the chunking logic changes" comment.

**Why:** The stated versioning contract becomes unreliable for the next change.

**Fix:** Bump both.

### F25 · Minor · Design/API

`.github/bot-scripts/postsIndex.cjs:18` _(not anchorable: line(s) outside every hunk)_

`MAX_QUESTION_LENGTH = 6000` untouched while docs chunks were retuned to 700 chars for the model's window.

**Why:** The asymmetry is undocumented while `POSTS_MIN_SIMILARITY` was recalibrated against it. See F4 for the correctness consequence.

**Fix:** Revisit, or comment why truncation is acceptable here.

### F32 · Minor · Tests

`.github/bot-scripts/buildDocsIndex.cjs:67` _(not anchorable: line(s) outside every hunk)_

`splitLongText` is not exported and no test content exceeds ~200 chars, so the paragraph-preference and `CHUNK_OVERLAP` stepping logic is untested — before and after a change that makes it fire far more often.

**Why:** The split path now runs on most documents rather than rarely.

**Fix:** Export it and test with text well over 700 chars.

### F41 · FYI · Design/API

`.github/bot-scripts/postsIndex.cjs:1` _(not anchorable: line(s) outside every hunk)_

The `cleanQuestion` HTML-comment regex to index-scan rewrite is ReDoS hardening unrelated to the embeddings migration (the head commit is titled "strip HTML comments without a backtracking regex").

**Why:** Reasonable in itself, but it is scope creep in this PR and it changes `cleanQuestion` output, which feeds the index-version question in F24.

**Fix:** Consider splitting; no action required.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+
+// Process a bounded number of texts per pipeline call to limit peak memory
+const BATCH_SIZE = 32;
+
+function defaultModelCacheDir() {
+	const xdg = process.env.XDG_CACHE_HOME;
+	const base = xdg && xdg.trim().length > 0
+		? xdg
+		: join(homedir(), ".cache");
+	return join(base, "zwave-js-ui-bot", "models");
```
robertsLando, line 32: **[Major] · DRY**

`defaultModelCacheDir()` computes `~/.cache/zwave-js-ui-bot/models` in JS while all 7 workflow sites hardcode the same literal, and the `BOT_MODEL_CACHE_DIR` override the module provides is used by none of them.

**Why:** The one mechanism that would make this coupling explicit is dead code; changing the JS path breaks every cache with no signal.

**Fix:** Set `BOT_MODEL_CACHE_DIR` from the extracted action so the path has one source, or drop the unused override.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+function getExtractor() {
+	extractorPromise ??= (async () => {
+		// The package is ESM-only, hence the dynamic import
+		const { pipeline } = await import("@huggingface/transformers");
+		return pipeline("feature-extraction", EMBEDDING_MODEL, {
+			cache_dir: process.env.BOT_MODEL_CACHE_DIR?.trim()
+				|| defaultModelCacheDir(),
+			revision: MODEL_REVISION,
+			dtype: "q8",
+		});
```
robertsLando: **[Major] · Operability**

`pipeline()` downloads from huggingface.co with no retry, no timeout, and no wrapped error message, on the user-triggered answer path.

**Why:** The restore/report design deliberately routes index outages to a tracking issue because a failed run notifies the issue author, not maintainers. An HF outage reintroduces exactly that: a red run on a reporter's issue with a raw stack trace, and nothing in the tracking issue.

**Fix:** Bound and retry the model load, and route its failure through the same degrade-and-report path as a missing index.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -297,10 +302,15 @@ async function buildDocsAnswerSection(
...
-	// (thrown by modelsRequest) is intentionally NOT caught here, so it
-	// fails the job instead of silently posting nothing.
-	const result = await judgeAnswer(question, ranked, token);
+	// answer" result internally. Skip the docs section on a genuine API
+	// error, the related-posts section is still useful.
+	let result;
+	try {
+		result = await judgeAnswer(question, ranked, token);
+	} catch (e) {
+		console.log(`Chat request failed: ${e.message}`);
```
robertsLando: **[Major] · Operability**

The new `judgeAnswer` catch logs at `console.log` and returns.

**Why:** Reverses the documented "fail the job instead of silently posting nothing" choice without replacing it with any signal; a sustained Models outage degrades every answer with no annotation. Compounding case: chat failure plus no related posts = no comment at all.

**Fix:** Keep the graceful degrade, but surface it as `::warning::` so the run carries an annotation.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+
+	/** @type {number[][]} */
+	const results = [];
+	for (let i = 0; i < inputs.length; i += BATCH_SIZE) {
+		const batch = inputs.slice(i, i + BATCH_SIZE);
+		const output = await extractor(batch, {
+			pooling: "mean",
+			normalize: true,
+		});
+		const data = output.tolist();
```
robertsLando: **[Major] · Tests**

`results.push(...(Array.isArray(data[0]) ? data : [data]))` is a shape-detection heuristic deciding whether an embedding is flattened correctly or silently corrupted, and `localEmbeddings.cjs` has no test sibling at all. The `BATCH_SIZE` chunking loop and the `embed([]) -> []` short-circuit are equally unverified.

**Why:** A batch of 1 misclassified as unbatched (or vice versa) pushes a malformed vector into the index — a data-integrity failure that surfaces only as degraded answer quality.

**Fix:** Add `localEmbeddings.test.cjs` with a stubbed `@huggingface/transformers` pipeline returning canned tensors; assert single-item, batch-boundary, and empty-input shapes. Note the Correctness lane found the branch is currently unreachable (pipeline always returns dims `[batch, 384]`) — so prefer asserting `output.dims.length === 2` over keeping the heuristic.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -32,14 +33,18 @@ const MAX_RETRIEVED_CHUNKS = 5;
 // If not even the best dense match reaches this cosine similarity,
 // the post is considered off-topic and no chat request is made.
 // The real relevance judgment is left to the chat model.
```
robertsLando, line 37: **[Major] · Tests**

Both similarity thresholds changed (0.2 to 0.35 at :35, 0.55 to 0.6 at :47) as the core of the recalibration, but neither constant is exported or referenced by any test.

**Why:** The gating behaviour is verified only by eval scripts run against a live model; an accidental change to either constant passes all 128 unit tests.

**Fix:** Export the thresholds or the gating decision functions and add boundary-value tests just below and at each floor.

.github/workflows/zwave-js-bot_post.yml
```diff
@@ -164,6 +164,23 @@ jobs:
...
+    - name: Setup Node.js
+      uses: actions/setup-node@v5
+      with:
+        node-version: 22
+        cache: 'npm'
+        cache-dependency-path: .github/bot-scripts/package-lock.json
+
+    - name: Install embedding dependencies
+      working-directory: .github/bot-scripts
+      run: npm ci --ignore-scripts
```
robertsLando: **[Major] · Performance**

`npm ci` installs `onnxruntime-node` (~220 MB unpacked) and `onnxruntime-web` (~134 MB, a hard dep of `@huggingface/transformers`, unused on Node) in 7 jobs; setup-node's `cache: npm` caches the download, not the ~355 MB extraction.

**Why:** Tens of seconds of pure extraction added to every issue and comment.

**Fix:** Cache `node_modules` itself keyed on the lockfile, or prune the web backend via an override.

.github/workflows/zwave-js-bot_post.yml
```diff
@@ -213,12 +229,31 @@ jobs:
...
+      uses: actions/setup-node@v5
+      with:
+        node-version: 22
+        cache: 'npm'
+        cache-dependency-path: .github/bot-scripts/package-lock.json
+
+    - name: Install embedding dependencies
+      if: steps.posts-index.outputs.found == 'true'
+      working-directory: .github/bot-scripts
+      run: npm ci --ignore-scripts
```
robertsLando: **[Major] · Performance**

`update-posts-index` independently repeats setup-node + `npm ci` + model load + ORT session init to produce one 384-dim vector, duplicating what the `answer` job in the same workflow already paid for.

**Why:** Roughly doubles runner minutes per issue/comment for a single embedding.

**Fix:** Fold the single-post embedding into the answer job.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -297,10 +302,15 @@ async function buildDocsAnswerSection(
...
-	const result = await judgeAnswer(question, ranked, token);
+	// answer" result internally. Skip the docs section on a genuine API
+	// error, the related-posts section is still useful.
+	let result;
+	try {
+		result = await judgeAnswer(question, ranked, token);
+	} catch (e) {
+		console.log(`Chat request failed: ${e.message}`);
+		return;
+	}
```
robertsLando, line 318: **[Minor] · Correctness**

The new catch is unbounded: a `TypeError` inside `judgeAnswer`/`validateJudgeResponse` is swallowed identically to a network failure, and reported via `console.log` rather than a warning annotation. Combined with dropping the `!modelsToken` early return, the bot can degrade to posts-only permanently with a green run.

**Why:** A parse or programming bug now permanently disables the docs section behind a `console.log`.

**Fix:** Narrow to the API-error case, surface as a workflow warning.

.github/bot-scripts/buildDocsIndex.cjs
```diff
@@ -1,22 +1,24 @@
...
 
 // Chunks shorter than this are unlikely to contain useful information
 const MIN_CHUNK_LENGTH = 50;
-// Sections longer than this are sub-split so nothing gets truncated away
-const MAX_CHUNK_LENGTH = 4000;
+// Sections longer than this are sub-split so nothing gets truncated away.
+// The embedding model truncates input at 256 tokens, and code blocks
+// tokenize at ~2-3 characters/token, so oversized chunks would lose
+// their tails to truncation
+const MAX_CHUNK_LENGTH = 700;
```
robertsLando: **[Minor] · Correctness**

The comment justifies `MAX_CHUNK_LENGTH = 700` by saying the model truncates at 256 tokens, but `tokenizer_config.json` at the pinned revision `751bff37` has `model_max_length: 512` (verified live). 700 chars measures 156 tokens, ~30% of the real window.

**Why:** The docs index is fragmented roughly 2x more than necessary, on a stated premise that does not match the pinned model.

**Fix:** Re-derive from 512, or restate 256 as a deliberate sentence-transformers training-length quality choice rather than as truncation.

.github/bot-scripts/postsIndex.cjs
```diff
@@ -27,14 +27,20 @@ const MAX_QUESTION_LENGTH = 6000;
...
+	for (;;) {
+		const start = text.indexOf("<!--", searchFrom);
+		if (start === -1) break;
+		const end = text.indexOf("-->", start + 4);
+		if (end === -1) break;
+		text = text.slice(0, start) + text.slice(end + 3);
+		// Removing a comment can splice the surrounding text into a new
+		// comment opener, re-check just before the removal point
+		searchFrom = Math.max(0, start - 3);
+	}
```
robertsLando, line 43: **[Minor] · Correctness**

The index scan is not equivalent to the regex it replaced on malformed/overlapping markup. A 400k-case fuzz found 5 divergences. The stated motivation is also not met: the new loop is O(n^2) (full `slice` per removal) and measured slower than the regex on adversarial input.

**Why:** Well-formed templates agree, so impact is low — but the change does not buy the ReDoS hardening it is named for.

**Fix:** If ReDoS is the goal, build the output in a single left-to-right pass with no re-slicing; otherwise keep the regex.

.github/workflows/zwave-js-bot_post.yml
robertsLando: **[Minor] · Security**

The model cache is saved by `actions/cache@v6` from a job with `timeout-minutes: 10`, with no equivalent of the index's "Check what the cache gave us" validation.

**Why:** A run cut short mid-download persists a truncated ONNX under the exact immutable key, and every later run restores the broken model until `localEmbeddings.cjs` changes.

**Fix:** Save the model cache only from the nightly builders (`cache/restore` elsewhere), or validate before use.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+				|| defaultModelCacheDir(),
+			revision: MODEL_REVISION,
+			dtype: "q8",
+		});
+	})();
+	return extractorPromise;
+}
+
+/**
+ * Embeds one or more texts, returning the embeddings in input order
```
robertsLando, line 74: **[Minor] · Design/API**

`embed` already batches at `BATCH_SIZE`, so `embedBatched` differs only by 1e-5 rounding plus a log line — a vestige of the old throttled-request split.

**Why:** The name misdescribes the real distinction (unrounded query vector vs rounded index vector), inviting the wrong pick at call sites.

**Fix:** Rename around the actual difference, e.g. `embed` / `embedForIndex`.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -517,26 +524,26 @@ async function main(param) {
...
 
-	// A single embedding request serves both docs retrieval and
-	// related-post ranking. Similarities are only comparable within
-	// one model, so an index embedded with a different model is skipped.
-	const embeddingModel = docsIndex?.model ?? postsIndex?.model;
-	if (postsIndex && postsIndex.model !== embeddingModel) {
+	// The question is embedded locally. Similarities are only comparable
+	// within one model, so indexes built with a different model are skipped
+	// until the nightly rebuild replaces them.
+	if (docsIndex && docsIndex.model !== EMBEDDING_MODEL) {
```
robertsLando: **[Minor] · DRY**

Seven hand-rolled `index.model !== EMBEDDING_MODEL` checks with near-identical messages — answerFromDocs.cjs:532-541, evalDocsAnswers.cjs:48, evalRelatedPosts.cjs:27, updatePostsIndex.cjs:57, buildDocsIndex.cjs:196, buildPostsIndex.cjs:173.

**Why:** This is the correctness guard for the whole migration; a missed site yields meaningless similarities rather than an error.

**Fix:** A single `indexMatchesModel(index)` helper next to `EMBEDDING_MODEL`, with callers keeping their own reaction. Largely subsumed if F6 is taken.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+ * @param {string[]} inputs
+ * @returns {Promise<number[][]>}
+ */
+async function embed(inputs) {
+	if (inputs.length === 0) return [];
+	const extractor = await getExtractor();
+
+	/** @type {number[][]} */
+	const results = [];
+	for (let i = 0; i < inputs.length; i += BATCH_SIZE) {
```
robertsLando: **[Minor] · Performance**

Fixed 32-item batches over unsorted inputs; transformers.js pads each batch to its longest sequence, so a 50-char breadcrumb chunk is computed at the batch maximum.

**Why:** With chunk lengths now spanning 50-700 chars, padding waste dominates the ~96s build.

**Fix:** Sort by length, batch, then restore input order.

.github/bot-scripts/buildDocsIndex.cjs
robertsLando: **[Minor] · Performance**

`MAX_CHUNK_LENGTH` is a character budget guarding a token limit; the comment itself notes code tokenizes at 2-3 chars/token, so 700 chars under-fills the window for prose while still truncating code.

**Why:** The worst case sets build time, index size, and cache churn for every chunk.

**Fix:** Budget by tokenizer length.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/** @type {Promise<any> | undefined} */
+let extractorPromise;
+
+function getExtractor() {
+	extractorPromise ??= (async () => {
+		// The package is ESM-only, hence the dynamic import
+		const { pipeline } = await import("@huggingface/transformers");
+		return pipeline("feature-extraction", EMBEDDING_MODEL, {
```
robertsLando: **[Minor] · Performance**

No session/threading options passed, so ORT defaults decide intra-op parallelism on a 2-4 vCPU runner.

**Why:** Unverified whether the ~96s build is core-bound or effectively single-threaded.

**Fix:** Measure, then pin `session_options` explicitly.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+	return results;
+}
+
+/**
+ * Embeds texts destined for an index
+ * @param {string[]} texts
+ * @returns {Promise<number[][]>} Embeddings in input order
+ */
+async function embedBatched(texts) {
+	console.log(`Embedding ${texts.length} texts locally...`);
```
robertsLando, line 126: **[Minor] · Operability**

`embedBatched` logs once before a multi-minute CPU loop; the removed API version logged per batch plus a total.

**Why:** A build that hits `timeout-minutes: 15` leaves no evidence of how far it got.

**Fix:** Log per-batch progress with a count.

.github/bot-scripts/package.json
```diff
@@ -0,0 +1,8 @@
+{
```
robertsLando, line 1: **[Minor] · Design/API**

No `engines`/`packageManager`, while `node-version: 22` is hardcoded in seven new workflow steps and `.nvmrc` exists (and master is moving to 24).

**Why:** No single source for the bot runtime.

**Fix:** Add `engines.node`, or use `node-version-file: .nvmrc`.

.github/bot-scripts/postsIndex.cjs
```diff
@@ -27,14 +27,20 @@ const MAX_QUESTION_LENGTH = 6000;
...
+	// an index scan, a regex would backtrack polynomially on crafted input
 	let text = body;
-	let previous;
-	do {
-		previous = text;
-		text = text.replace(/<!--[\s\S]*?-->/g, "");
-	} while (text !== previous);
+	let searchFrom = 0;
+	for (;;) {
+		const start = text.indexOf("<!--", searchFrom);
```
robertsLando: **[Minor] · Readability**

The `start - 3` backup offset in the re-scan loop is a magic number with no stated derivation; it is only correct because `"<!--".length - 1 === 3`.

**Why:** If the delimiters ever changed, this silently becomes wrong.

**Fix:** Derive it from the delimiter literal.

.github/workflows/docs-embeddings.yml
robertsLando: **[Nit] · Performance**

All 7 sites use `actions/cache` in save mode with an identical model key; concurrent misses each upload ~24 MB and all but one is discarded.

**Why:** Wasted upload bandwidth and cache quota.

**Fix:** `cache/restore` in consumers, one designated saver.

.github/workflows/docs-embeddings.yml
```diff
@@ -42,16 +43,34 @@ jobs:
       uses: actions/cache@v6
       with:
         path: .docs-index
-        key: docs-embeddings-v2-${{ hashFiles('docs/**/*.md', '.github/bot-scripts/buildDocsIndex.cjs', '.github/bot-scripts/docsIndex.cjs', '.github/bot-scripts/modelsApi.cjs') }}
+        key: docs-embeddings-v3-${{ hashFiles('docs/**/*.md', '.github/bot-scripts/buildDocsIndex.cjs', '.github/bot-scripts/docsIndex.cjs', '.github/bot-scripts/localEmbeddings.cjs') }}
```
robertsLando: **[Nit] · Design/API**

The `docs-embeddings-v3-${{ hashFiles(...) }}` expression is duplicated in three places across two files (also :111).

**Why:** The next bump has three chances to drift.

**Fix:** Single source.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+// Local embedding pipeline running on the CI runner, replacing the
+// retired GitHub Models embeddings API
+
+const { homedir } = require("node:os");
+const { join } = require("node:path");
+
+// Keep model, revision and dtype in sync with the zwave-js repository
+// (.github/bot-scripts/localEmbeddings.cjs there), so evals and answer
+// quality stay comparable between the two bots
+const EMBEDDING_MODEL = "Xenova/all-MiniLM-L6-v2";
```
robertsLando, line 12: **[Nit] · Design/API**

The `EMBEDDING_MODEL` env override is gone while `CHAT_MODEL`'s survives in `modelsApi.cjs:6`.

**Why:** Deliberate (revision pinning), but the sibling asymmetry is worth one comment line.

**Fix:** One comment.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+
+function getExtractor() {
+	extractorPromise ??= (async () => {
+		// The package is ESM-only, hence the dynamic import
+		const { pipeline } = await import("@huggingface/transformers");
+		return pipeline("feature-extraction", EMBEDDING_MODEL, {
+			cache_dir: process.env.BOT_MODEL_CACHE_DIR?.trim()
+				|| defaultModelCacheDir(),
+			revision: MODEL_REVISION,
+			dtype: "q8",
```
robertsLando: **[Nit] · Readability**

`dtype: "q8"` is asserted with no rationale, unlike the model, revision, and `BATCH_SIZE` values in the same file which all carry an explicit "why".

**Why:** Inconsistent with the file's own standard.

**Fix:** One line on the size/quality tradeoff.

.github/bot-scripts/localEmbeddings.cjs
```diff
@@ -0,0 +1,84 @@
...
+	/** @type {number[][]} */
+	const results = [];
+	for (let i = 0; i < inputs.length; i += BATCH_SIZE) {
+		const batch = inputs.slice(i, i + BATCH_SIZE);
+		const output = await extractor(batch, {
+			pooling: "mean",
+			normalize: true,
+		});
+		const data = output.tolist();
+		results.push(...(Array.isArray(data[0]) ? data : [data]));
```
robertsLando: **[Nit] · Correctness**

`Array.isArray(data[0]) ? data : [data]` is dead code — the pipeline always returns dims `[batch, 384]`, verified for batch 1 and batch 3. If it ever did fire it would push 384 scalars as one "vector" and `cosineSimilarity` would consume it silently.

**Why:** A defensive branch that cannot fire but would corrupt data if it did.

**Fix:** Prefer asserting `output.dims.length === 2`.

.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -32,14 +33,18 @@ const MAX_RETRIEVED_CHUNKS = 5;
 // If not even the best dense match reaches this cosine similarity,
 // the post is considered off-topic and no chat request is made.
 // The real relevance judgment is left to the chat model.
-const MIN_SIMILARITY = 0.2;
+// Calibrated for all-MiniLM-L6-v2: on-topic questions score >= 0.36
+// against the docs index, clearly off-topic ones <= 0.17
+const MIN_SIMILARITY = 0.35;
```
robertsLando, line 40: **[Nit] · DRY**

The thresholds are documented as calibrated for `all-MiniLM-L6-v2` but live two modules away from where that model is pinned, and the eval scripts never import them.

**Why:** Calibration and the thing it is calibrated against can drift silently.

**Fix:** Optional — co-locate, or have the evals assert against them.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Partially addressed in 7456e94a: every mismatch skip now goes through a shared `indexMatchesModel()` helper that emits a `::warning::` annotation, so the degraded state is visible on the run instead of only in the log. Wiring the model into `restore-bot-index`'s health predicate would mean teaching the bash action about a JS constant; with the nightly rebuild replacing a mismatched index within a day, the annotation seemed like the right cost/benefit.

.github/workflows/zwave-js-bot_post.yml
AlCalzone: Declining this one: the lockfile pins exact versions with integrity hashes, so a substituted tarball can't install without a reviewed lockfile change, and the PAT in that process belongs to the bot account, not a maintainer. Splitting the embedding into a secretless step would need to be redone for the gh-aw pre-activation in #4765 and adds a second handoff for a threat the lockfile already covers.

.github/bot-scripts/package.json
AlCalzone: The flag was never meant as a security boundary - it skips onnxruntime-node's postinstall, which downloads GPU execution providers we don't want to pay for. The composite action from 7456e94a documents exactly that; no hardening claim is attached to it anywhere.

.github/workflows/docs-embeddings.yml
AlCalzone: Fixed in 7456e94a - the dependency lockfile is part of the docs index key in all three places.

.github/workflows/zwave-js-bot_post.yml
AlCalzone: Fixed in 7456e94a - the key is derived from the model pin (`id@revision@dtype`, exported as `MODEL_CACHE_KEY`) with a `restore-keys:` prefix, owned by a single composite action, so editing unrelated code in localEmbeddings.cjs no longer evicts the weights.

.github/workflows/docs-embeddings.yml
AlCalzone: Fixed in 7456e94a - `.github/actions/setup-bot-embeddings` owns the whole block, all seven sites use it.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Fixed in 7456e94a - the composite action sets `BOT_MODEL_CACHE_DIR`, so the path literal has one owner and the JS override is live.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Fixed in 7456e94a - the model load retries three times with backoff, annotates each failed attempt, and wraps the final error with the model name.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Fixed in 7456e94a - the degrade is surfaced as a `::warning::` annotation.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Fixed in 7456e94a - the shape heuristic is replaced by a dims assertion that throws on anything but `[batch, hidden]`, and `localEmbeddings.test.cjs` covers single-item, batch-boundary, empty-input and rounding behavior against a stubbed pipeline.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Partially fixed in 7456e94a: both thresholds are exported and the posts floor has boundary tests at and just below the floor. The docs floor sits behind the retrieval pipeline in this PR's shape; #4765 refactors that into `retrieveDocsChunks` and its round-trip test exercises the floor there.

.github/workflows/zwave-js-bot_post.yml
AlCalzone: Fixed in 7456e94a - `node_modules` is cached as a whole, keyed on the lockfile, so the ~350 MB extraction happens once per dependency change instead of per run. Pruning onnxruntime-web via overrides felt too invasive for what the cache already buys.

.github/workflows/zwave-js-bot_post.yml
AlCalzone: Declining: #4765 removes the answer job from this workflow entirely (answering moves to the agentic workflow), leaving update-posts-index as the only embedding consumer here - there is nothing left to fold into. The node_modules and model caches from 7456e94a take the repeated setup down to seconds.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: The warning annotation is in 7456e94a. Narrowing on the error type doesn't separate what you'd want it to: fetch network failures are TypeErrors too, so type-based classification can't tell an outage from a bug. The judge path is deleted in #4765 either way.

.github/bot-scripts/buildDocsIndex.cjs
AlCalzone: Comment fixed in 7456e94a - 256 word pieces is the model's reference-implementation input cap (its training length), not tokenizer truncation, and the chunk size is stated as the eval-gated quality choice it is (18/18 hit@5).

.github/bot-scripts/postsIndex.cjs
AlCalzone: Fixed in 7456e94a - the strip is now a single left-to-right pass: O(n), no re-slicing, no magic offset, and it matches the lazy-regex semantics on spliced text (a splice-created `<!--` is not re-scanned, same as the regex).

.github/workflows/zwave-js-bot_post.yml
AlCalzone: The save runs in actions/cache's post step, which is skipped when the job failed - and a truncated download fails the build step, so a partial model can't be persisted that way. Regardless, 7456e94a moves saving to the nightly builders only; user-triggered jobs restore-only now.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Keeping the names for symmetry with the zwave-js copy of this file - the rounding distinction is documented at the definition, and renaming across both repos isn't worth the churn.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Fixed in 7456e94a - a shared `indexMatchesModel()` covers the five mismatch sites. The two builders keep their inline comparison deliberately: there it gates embedding reuse, not a skip, and warning would be wrong.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Declining - the full build embeds 397 chunks in ~6 s wall locally and ~2 min on a runner, far inside the nightly budget. Not worth the sort/restore-order machinery.

.github/bot-scripts/buildDocsIndex.cjs
AlCalzone: Declining - budgeting by tokenizer length couples the chunker to the model package for a build that is already fast, and the retrieval evals gate the outcome either way. The char budget erring conservative for prose is acceptable.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Declining for the same reason - the build is nowhere near its budget, so pinning ORT threading would be speculative tuning.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Fixed in 7456e94a - per-batch progress is logged for multi-batch runs.

.github/bot-scripts/package.json
AlCalzone: Fixed in 7456e94a - `engines.node` in the manifest, and the composite action uses `node-version-file: .nvmrc`, so the version lives in one place instead of seven.

.github/bot-scripts/postsIndex.cjs
AlCalzone: Gone in 7456e94a with the single-pass rewrite.

.github/workflows/docs-embeddings.yml
AlCalzone: Fixed in 7456e94a - consumers use `cache/restore`, only the nightly builders save.

.github/workflows/docs-embeddings.yml
AlCalzone: The remaining duplication is structural: two of the three sites are adjacent jobs in one file and the third is the composite action, and Actions has no cross-file expression reuse. The lockfile addition touched all three in one commit, which is the practical mitigation.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: modelsApi.cjs and `CHAT_MODEL` are deleted entirely in #4765, which resolves the asymmetry without a comment.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Fixed in 7456e94a - one line on the size/quality tradeoff.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Fixed in 7456e94a - replaced by the dims assertion.

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: The thresholds are exported and referenced by tests now (7456e94a). They stay next to the gating logic that uses them - one require away from the model pin, with the calibration comment naming the model explicitly.

.github/bot-scripts/buildDocsIndex.cjs
AlCalzone: You are right, my earlier reply undersold this — the pipeline truncates at the tokenizer's 512, so 700 chars (sized for 256) left half the window unused. Revisited empirically in d71aa354: at 1200 chars, p99 measures ~440 tokens with 0.6% slight overflow (vs 4.5% over 256 today), the index shrinks from 397 to 310 chunks, eval hit@5 stays 18/18 at every cap tried (1000/1200/1400), and the similarity floor separation is unchanged. 1400+ bought almost nothing, so 1200 it is.

.github/bot-scripts/localEmbeddings.cjs
AlCalzone: Reversing my earlier decline: with the chunk cap raised to 1200 (d71aa354) the length spread got wide enough that this is worth having. Implemented in 696a06fe — texts are sorted by length before batching and the input order is restored afterwards, with a test pinning the restoration.

.github/bot-scripts/buildDocsIndex.cjs
AlCalzone: Following up after the chunk-size revisit (d71aa354): the under-fill half of this is addressed empirically — at 1200 chars, p99 measures ~440 tokens against the real 512 window and only 0.6% of chunks overflow slightly (vs 4.5% over 256 before). Still not budgeting by tokenizer length: it would couple the chunker to the model package, and the measured distribution plus the eval gate (18/18 at every cap tried) achieve the same outcome without it.

robertsLando (approved):

PR #4771, opened 2026-08-01
feat(ui): use a consistent design language for dialogs

AlCalzone:
This PR continues the dashboard redesign by focusing on the dialogs:

image

The dialog contents are not touched yet, this will be done in a followup PR.

It also adds a new wizard style that works on mobile and desktop:
image

image


src/components/Confirm.vue
```diff
@@ -242,11 +204,14 @@ import 'prismjs/components/prism-clike'
 import 'prismjs/components/prism-javascript'
 import 'prismjs/themes/prism-tomorrow.css'
 import { wrapFunc, noop } from '../lib/utils'
+import { cancelAction, confirmAction } from '../lib/dashboard-types'
```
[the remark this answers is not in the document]
AlCalzone: stay consistent with other files. eslint will complain if this is not correct.

src/components/dialogs/Password.vue
```diff
@@ -1,105 +1,81 @@
...
+				</v-row>
+			</v-form>
+		</v-container>
+	</ZwDialog>
 	<!-- END DIALOG PASSWORD -->
 </template>
 
 <script>
+import ZwDialog from '@/components/dashboard/dialogs/ZwDialog.vue'
+import { cancelAction, confirmAction } from '@/lib/dashboard-types'
```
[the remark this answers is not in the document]
AlCalzone: dito

src/components/dialogs/DialogSceneValue.vue
```diff
@@ -1,111 +1,95 @@
...
+				</v-row>
+			</v-form>
+		</v-container>
+	</ZwDialog>
 </template>
 
 <script>
 import { defineAsyncComponent } from 'vue'
+import ZwDialog from '@/components/dashboard/dialogs/ZwDialog.vue'
+import { cancelAction, confirmAction } from '@/lib/dashboard-types'
```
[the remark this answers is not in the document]
AlCalzone: dito

src/components/dialogs/DialogGatewayValue.vue
```diff
@@ -257,14 +231,18 @@ import 'prismjs/themes/prism-tomorrow.css' // import syntax highlighting styles
 
 import { mapState } from 'pinia'
 import useBaseStore from '../../stores/base.js'
+import { cancelAction, confirmAction } from '../../lib/dashboard-types'
```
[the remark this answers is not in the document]
AlCalzone: dito

src/components/dialogs/DialogNodesManager.vue
```diff
@@ -1251,41 +941,38 @@ export default {
...
+			})
+		},
 		async show(stepOrStepsValues) {
 			this.isOpen = true
 			this.$emit('open')
+			this.init(true)
 			if (typeof stepOrStepsValues === 'object') {
-				this.init(true)
 				this.steps = []
 				for (const s in stepOrStepsValues) {
```
robertsLando, line 951: **[Blocker] · Correctness**

`availableSteps` no longer has an `action` key (it was removed with the action step), but `src/mixins/InstancesMixin.js:623` still calls `showNodesManager({ action: { action: 1 }, replaceFailed: …, replaceInclusionMode: … })` for "Replace failed node". This loop hits `pushStep('action')` first, `s` resolves to `undefined`, and `s.index = …` throws.

**Why:** `show()` sets `isOpen = true` before the loop, so the dialog opens with `steps = []` — no title, no footer, no way forward. `showNodesManager` doesn't await `show()`, so it surfaces only as an unhandled rejection. `replaceFailed`/`replaceInclusionMode` are never pushed either.

**Fix:** Update the mixin payload to the new step keys (or route it through `showForAction('replace-failed')`), and make `pushStep` fail loudly on an unknown key rather than TypeError-ing.

src/components/Confirm.vue
```diff
@@ -290,6 +254,51 @@ export default {
...
+				case 'success':
+					return 'success'
+				case 'info':
+					return 'info'
+				default:
+					return 'default'
+			}
+		},
+		// Quantise `options.width` to a DialogSize breakpoint
+		dialogSize() {
```
robertsLando, line 277: **[Major] · Design/API**

`dialogSize()` compares `options.width` numerically, but `nodes-table/OTAUpdates.vue:60` and `nodes-table/OTWUpdates.vue:59` pass the string `'500px'`. Every `w <= N` is false via NaN coercion, so both fall through to `'xl'` = 920px. The pre-PR template used `:max-width="options.width"`, which accepted CSS strings.

**Why:** Silent width regression on an app-wide helper with 62 call sites — no error, no warning, just wrong-sized dialogs.

**Fix:** Coerce with `parseInt` in the mapper, or narrow the documented `width` option to `number` and fix the two string callers.

src/views/ControlPanelNew.vue
```diff
@@ -195,16 +199,26 @@ async function onAction(device: Device, action: DeviceAction) {
...
 		const key = actionPendingKey(device, action)
 		if (key) setStatus(key, 'pending')
+		// Restore can brick the controller, so block interaction until it settles
+		app.showBlockingLoader(
+			'Restoring NVM',
+			'While doing the restore the Z-Wave radio will be turned on/off. Do not disconnect the controller…',
+		)
 		let ok = false
 		try {
 			const response = await app.apiRequest(
```
robertsLando, line 209: **[Major] · Operability**

`apiRequest('restoreNVM', …)` has no timeout, and `finishLoader()` runs only in the `finally`. If the socket drops mid-restore — plausible for an operation that power-cycles the radio and can take down the driver — the promise never settles.

**Why:** `dialogLoader` stays up with `blocking: true`, which hides the close button and suppresses Esc/scrim, so the user is stuck on an indeterminate spinner with no outcome and no exit. That is worse than the flow it replaced, on the single riskiest operation in the app.

**Fix:** Drive a failed terminal state from a timeout or a `socket.on('disconnect')` handler, and re-enable dismissal when it fires.

src/components/dialogs/DialogExcludeDevice.vue
```diff
@@ -0,0 +1,215 @@
...
+			}
+		},
+		onModel(v) {
+			if (!v) this.close()
+		},
+		close() {
+			this.$emit('update:modelValue', false)
+			this.$emit('close')
+		},
+		reset() {
```
robertsLando: **[Major] · Correctness**

Three overlapping lifetime problems here: `stop()` awaits `stopExclusion` and only then calls `settle()`, so closing during the await leaves a 3s timer that outlives the dialog; `begin()` never calls `clearSettle()`; and `reset()` fires an un-awaited `stopExclusion` with `infoSnack:false, errorSnack:false` and no catch.

**Why:** Reopening within 3s flips a genuinely-running exclusion to `phase = 'stopped'` ("No device excluded") with the Stop button gone; the restart races the in-flight stop with no ordering guarantee; and if the stop call fails, the controller stays in exclusion mode with zero signal in the UI or the console.

**Fix:** Clear the timer in `begin()`, guard `stop()`/`settle()` behind a generation counter or an `open` flag, and at minimum log the stop failure.

src/components/dialogs/DialogHealthCheck.vue
```diff
@@ -442,14 +411,15 @@ export default {
...
 				)
-			} else if (open === false) {
-				this.unbindEvents()
-				this.results = []
-				this.loading = false
-				this.targetNode = null
-				this.averages = null
 			}
 		},
+		teardown() {
```
robertsLando, line 426: **[Major] · Correctness**

`init()` no longer clears `results` / `averages` / `loading` — that moved here, to `teardown()` on `@after-leave`. Vue does not emit `afterLeave` when a leave transition is cancelled by a re-open.

**Why:** Close-then-immediately-reopen shows the previous run's results with a stuck `loading` spinner. Same shape at `DialogAssociation.vue:10` and `Password.vue:10` (stale group/target/validation state, stale password fields).

**Fix:** Reset result state in `init(open)` as well; treat `after-leave` as best-effort cleanup, never as the only reset path.

src/components/dashboard/dialogs/ZwWizardDialog.vue
```diff
@@ -0,0 +1,90 @@
+<template>
+	<ZwDialog
+		:model-value="modelValue"
+		:size="size"
+		:persistent="persistent"
```
robertsLando: **[Major] · Correctness**

The wizard forwards `persistent` but not `blocking`, and `ZwDialogStepProgress.vue:51` renders `ZwDialogClose` unconditionally.

**Why:** `:persistent="loading || state === 'start'"` blocks Esc and scrim only — the X still tears the dialog down mid-inclusion, and `close()` never stops the running inclusion on the controller.

**Fix:** Forward `blocking` through the wizard and gate the rail/progress close button on it.

src/components/dashboard/dialogs/ZwDialog.vue
```diff
@@ -0,0 +1,324 @@
...
+	warning: AlertIcon,
+	danger: AlertIcon,
+}
+
+const chipIcon = computed(
+	() => props.icon ?? DEFAULT_ICON[props.severity] ?? null,
+)
+const showClose = computed(() => !props.blocking)
+
+function onModel(v: boolean) {
```
robertsLando, line 173: **[Major] · Design/API**

`close` only ever fires from `onModel(false)`, so it is fully redundant with `update:modelValue false` — and consumers wire both (`DialogHealthCheck.vue:8-11`, `DialogGatewayValue.vue:8-9`). Worse, the footer's own Close action (`DialogHealthCheck.vue:273`) emits *only* `close` and never lowers `modelValue`, while the X emits both.

**Why:** Two dismissal paths with different payloads is exactly the ambiguity every future dialog on this branch will get wrong.

**Fix:** Pick one signal: either drop `close`, or make it the sole dismissal event and stop emitting `update:modelValue` internally.

src/components/dashboard/dialogs/ZwDialog.vue
```diff
@@ -0,0 +1,324 @@
...
+
+const WIDTHS: Record<DialogSize, number> = {
+	sm: 400,
+	md: 540,
+	lg: 720,
+	xl: 920,
+}
+const widthPx = computed(() => WIDTHS[props.size])
+
+const RULE = { success: true, warning: true, danger: true } as Record<
```
robertsLando: **[Major] · Design/API**

`severity="info"` is a no-op: `RULE` has no `info` key (no rule bar) and `DEFAULT_ICON` has no `info` entry (no chip), so it renders identically to `default`. `Confirm.vue:266` already maps `color: 'info'` onto it.

**Why:** A public enum member that silently does nothing will keep getting used, on every future dialog, with nobody noticing it has no effect.

**Fix:** Give `info` a rule and a default icon, or drop it from `DialogSeverity`.

src/components/dashboard/dialogs/ZwDialog.vue
```diff
@@ -0,0 +1,324 @@
+<template>
+	<v-dialog
```
robertsLando, line 2: **[Major] · DRY/Codebase fit**

This builds on Vuetify's `v-dialog`, but `dashboard/**` had already standardised on `@vuetify/v0` headless primitives: `Dialog.Root`/`Dialog.Content` in `layout/ZwSidebar.vue:3-17,36`, plus `Popover` (ZwDropdown, ZwToggleMenu, ZwValueRow, ZwControllerOptionRow, ZwValuesActionsMenu), `Progress` (ZwActivityReadout), `Tabs`, `Slider`, `Switch`, `Button`.

**Why:** Two overlay stacks now coexist on `dashboard-neo` with different focus-trap, scrim and teleport semantics — and this wrapper has to reach into Vuetify internals (`.v-overlay__content`, `.v-overlay__scrim`) with unscoped CSS to undo defaults, a coupling the v0 route avoids.

**Fix:** Either build on `Dialog.Root`/`Dialog.Content` like `ZwSidebar`, or record an explicit decision that v0's Dialog is insufficient and migrate `ZwSidebar` the other way.

src/components/dashboard/dialogs/ZwWizardDialog.vue
```diff
@@ -0,0 +1,90 @@
...
+)
+
+const emit = defineEmits<{
+	'update:modelValue': [boolean]
+	close: []
+	afterLeave: []
+}>()
+
+// Switch to circular header below the two-pane breakpoint
+const mql =
```
robertsLando: **[Major] · DRY/Codebase fit**

`TWO_PANE_BREAKPOINT` is consumed here via `window.matchMedia`, but `src/lib/dashboard-breakpoints.ts:1` documents these constants as *container*-width, not window-width. Every other consumer honours that by threading a measured `viewport` prop (`ZwNodeDetailsBody.vue:487,500`, `ZwDeviceRow.vue:6`, `ZwDeviceDrawer.vue:73`, `deviceRowGrid.ts:36`) fed by `ResizeObserver`.

**Why:** The dialog's own width is `min(920px, 100vw − 32px)`, so the rail-vs-ring switch disagrees with the actual container width near the threshold. This is also the only raw `matchMedia` used for layout anywhere in `src/`.

**Fix:** Measure the dialog content element and compare against the constant, consistent with the other consumers.

src/assets/css/tokens.css
```diff
@@ -213,6 +219,17 @@
 	}
 }
 
+/* Indeterminate progress sweep — a short segment slides across the track
+   and back, for operations with no known percentage. */
+@keyframes zw-indeterminate {
```
robertsLando, line 225: **[Major] · Performance**

`zw-indeterminate` animates `left` on an absolutely-positioned bar, `1.4s ease-in-out infinite`, for the whole time any dialog is in `loading` state (NVM restore/backup, inclusion wait, health check).

**Why:** Animating `left` forces a layout recalculation every frame rather than a compositor-only repaint. On a Pi-hosted instance viewed on a phone this runs continuously — for minutes during an NVM restore — concurrently with live Socket.IO-driven re-renders elsewhere in the tree.

**Fix:** Animate `transform: translateX(…)` instead so the sweep stays on the compositor.

src/components/dashboard/dialogs/ZwDialogStepProgress.vue
```diff
@@ -0,0 +1,153 @@
...
+const segArc = computed(() => ((360 / divisor.value - gap.value) / 360) * C)
+const gapArc = computed(() => (gap.value / 360) * C)
+const rot = computed(() => `rotate(${gap.value / 2 - 90} ${cx} ${cx})`)
+
+const bgDash = computed(() =>
+	total.value > 1 ? `${segArc.value} ${gapArc.value}` : `${C}`,
+)
+
+// Last gap absorbs leftover circumference so the pattern sums to exactly C
+function ringDash(start: number, count: number) {
```
robertsLando: **[Minor] · Correctness**

The empty-steps guard added in a5881a7d is incomplete. `ringDash(0, 0)` leaves `arr[-1] = NaN` and returns `dasharray: ''`; `count > steps.length` produces a negative trailing segment, which invalidates the entire `stroke-dasharray`; and `segArc` goes negative once `steps.length >= 17`, since 360/17 = 21.2° is below the 22° gap.

**Why:** The `v-if` guards mask the `count = 0` case today but not the other two. This is also pure math with no component dependencies — extractable and unit-testable under the existing vitest setup, and a regression test would have caught the bug a5881a7d fixed.

**Fix:** Clamp `current` into `[0, total]`, floor `segArc` at 0, and pin the boundary inputs (0, 1, > total, >= 17 steps) in a unit test.

src/components/dialogs/DialogNodesManager.vue
```diff
@@ -1251,41 +941,38 @@ export default {
...
 			}
 		},
 		close() {
 			this.isOpen = false
 			this.$emit('close')
-			this.init(false)
 		},
-		init(bind, step = 'action') {
+		init(bind) {
 			this.steps = []
```
robertsLando, line 962: **[Minor] · Correctness**

`init()` no longer resets `currentStep` — the old one did, implicitly, via `pushStep('action')`.

**Why:** On reopen, `current = currentStep - 1` indexes past the freshly-emptied `steps` for one tick.

**Fix:** Reset `currentStep` alongside `steps` here.

src/App.vue
```diff
@@ -807,8 +812,15 @@ export default {
 			}
 		},
 		showNodesManager(stepOrStepsValues) {
-			// used in ControlPanel.vue
-			this.$refs.nodesManager.show(stepOrStepsValues)
+			this.$refs.nodesManager.show(
+				stepOrStepsValues || { inclusionNaming: {} },
```
robertsLando, line 820: **[Minor] · Correctness**

`showNodesManager('')` (from `onGrantSecurityClasses`) falls through to this `{ inclusionNaming: {} }` default.

**Why:** A driver-initiated `grantSecurityClasses` then shows a dead "Name and Location" step ahead of Security Classes, whose values are never read.

**Fix:** Push only the granted step for the driver-initiated path.

src/components/dialogs/DialogExcludeDevice.vue
```diff
@@ -0,0 +1,215 @@
...
+	components: { ZwDialog, AlertIcon, CheckIcon },
+	mixins: [InstancesMixin],
+	emits: ['update:modelValue', 'close'],
+	props: {
+		modelValue: Boolean,
+		socket: Object,
+	},
+	data() {
+		return {
+			phase: 'running',
```
robertsLando: **[Minor] · Correctness**

`phase` initialises to `'running'`, and the component is mounted unconditionally from `App.vue:353`.

**Why:** `beforeUnmount → reset()` therefore fires a spurious `stopExclusion` on teardown even when the dialog was never opened.

**Fix:** Add an `idle` initial phase so `reset()` is a no-op until exclusion actually starts.

robertsLando (commented):
## Deep review — `feat(ui): use a consistent design language for dialogs`

**Verdict: Needs work.** The design layer itself is sound and worth landing — I have no architectural objection. `ZwDialog` owns chrome only and leaves overlay mechanics to Vuetify, which is the right split; `DialogAction[]` is properly factored through `cancelAction`/`confirmAction` (`dashboard-types.ts:280-298`) and all 12 action-bearing dialogs use it rather than escaping to slots; secret handling survived the migration intact (password masking, `validate()` before emit, the DSK PIN never leaking into the shared step template, Enter unable to bypass PIN validation); and dropping `@keydown.esc` on the nodes manager is a genuine hardening. 17 dialogs onto one shell is a real win.

What needs fixing first is one verified crash, one lost entry point, and a cluster of state-reset races.

**Top 3 risks**
1. "Replace failed node" throws and opens an empty dialog.
2. Exclusion has no entry point left on the default route.
3. Cleanup moved to `@after-leave`, which doesn't fire on a cancelled leave transition — stale state across five dialogs.

### Themes

Most individual findings trace back to three root causes:

- **`@after-leave` replaced the `modelValue` watcher as the reset hook.** The watcher fired on open *and* close, synchronously. `@after-leave` fires only when a leave transition completes, and Vue skips it when a re-open cancels the leave. Every dialog that moved reset there can now show the previous session's state: `DialogHealthCheck`, `DialogAssociation`, `Password`, `DialogNodesManager`, `DialogLinkReliability`.
- **Dismissal is encoded in four overlapping signals** — `persistent`, `blocking`, `close`, `update:modelValue` — with no single owner. `ZwWizardDialog` forwards `persistent` but not `blocking`; `Confirm` maps only `persistent`; the footer Close action and the X emit different things. A single `dismiss: 'all' | 'button' | 'none'` would close most of this.
- **Removing the `action` step wasn't swept.** It had three branches (include / replace / exclude, values 0/1/2 in the pre-PR `nextStep()`). Removing it orphaned one caller and one entry point — the two blockers below.

### Findings outside the diff

These are real but anchor to lines this PR doesn't touch, so they can't be line comments:

- **[Blocker · Correctness] Exclusion is unreachable on the legacy control panel.** The old `action` step offered exclusion as option 2. Now `showNodesManager()` with no args hard-defaults to `{ inclusionNaming: {} }` (`App.vue:814-817`), and `showExcludeDevice` is wired only to `ZwAddDeviceSplitButton` in `ControlPanelNew`. `router/index.js:40` still sets `Routes.main = Routes.controlPanel`, so the legacy panel is the default landing route and its "Manage nodes" FAB (`views/ControlPanel.vue:134` → `:433`) is the only way in. Either wire that FAB to `showExcludeDevice` too, or — if the legacy panel is slated for deletion on this branch — say so on the PR so this reads as a known temporary gap rather than an oversight.
- **[Major · Correctness] `onNodeRemoved` still reports removals as additions.** `DialogNodesManager.vue:999` still binds `nodeRemoved` and its handler still calls `showResults()`, but `showResults` lost its `currentAction === 'Exclusion'` branch when exclusion moved out. A `nodeRemoved` arriving while `waitTimeout` is armed renders "Node N added with security None" for a node that was *removed*. Unbind it here, or have it only clear the wait state.

### Also worth a look (nits, not blocking)

- `ZwDialog.vue:174` — `variantFor` maps `kind:'filled' + tone:'neutral'` to `'primary'`, silently dropping the neutral tone.
- `ZwDialog.vue:56` — `:autofocus` is passed to `ZwButton`, which declares no such prop; it works only by attr fallthrough onto `Button.Root`.
- `ZwDialogStepRail.vue` / `ZwButton.vue` — hardcoded 210px rail width, 22px mark, radii and padding where tokens exist or should.
- `DialogAssociation.vue:119`, `DialogGatewayValue.vue:234` — relative `'../../lib/dashboard-types'` where the other 15 migrated dialogs use the `@/lib/…` alias.
- Confirm's footer order flipped (was Confirm→Cancel, now Cancel→Confirm) — intentional? Removing `zIndex: 2500` is safe; Vuetify's `useStack` auto-increments.
- The remaining `v-stepper` (`TemplateWizard.vue:9`) is a full-page view, not a dialog, and is already flagged for migration at `plugins/palette.ts:22`.

### Coverage

Reviewed at `a5881a7d` across nine dimensions — correctness (two lanes: new primitives/behavioural, and the migrated dialogs), security, DRY & codebase fit, design/API/backcompat, performance, operability, tests, readability. All 28 changed files covered. Security found no vulnerabilities, only the naming FYI below. No unresolved prior review threads.

src/components/dialogs/DialogExcludeDevice.vue
```diff
@@ -0,0 +1,215 @@
...
+				this.settle()
+			}
+		},
+	},
+	methods: {
+		begin() {
+			this.phase = 'running'
+			this.removed = null
+			this.subscribeChannels(['nodes'])
+			this.bindEvent('nodeRemoved', this.onNodeRemoved.bind(this))
```
robertsLando, line 120: **[Minor] · Correctness**

`bindEvent` calls `socket.on` but records only the last handler. Same shape at `DialogHealthCheck.vue:410` and `DialogLinkReliability.vue:280`.

**Why:** Two `begin()`/`init()` calls without an intervening `unbindEvents()` leak the first listener, so `nodeRemoved` gets handled twice.

**Fix:** Unbind before binding.

src/components/dialogs/DialogExcludeDevice.vue
```diff
@@ -0,0 +1,215 @@
...
+		},
+	},
+	watch: {
+		modelValue(v) {
+			if (v) this.begin()
+		},
+		controllerStatus(status) {
+			if (
+				!status ||
+				(this.phase !== 'running' && this.phase !== 'stopping') ||
```
robertsLando: **[Minor] · Correctness**

The `controllerStatus` watcher opens the settle window but leaves `phase === 'running'`.

**Why:** The footer keeps an enabled "Stop" that re-issues `stopExclusion` and restarts the timer.

**Fix:** Move to a distinct phase when the settle window opens.

src/components/Confirm.vue
```diff
@@ -1,235 +1,197 @@
...
-		:max-width="options.width"
-		:style="{ zIndex: options.zIndex }"
-		@keydown.esc="cancel"
-		:persistent="options.persistent"
+	<ZwDialog
+		:model-value="show"
+		:size="dialogSize"
+		:severity="severity"
+		:title="title"
+		:blocking="options.persistent"
```
robertsLando: **[Minor] · Correctness**

`blocking` is derived from `options.persistent` only, not from `noCancel`. `App.vue:1699` passes `noCancel: true` with no `persistent`.

**Why:** Since `ZwDialog` computes `showClose = !blocking`, that acknowledgement dialog now gets an X. (Esc and scrim already dismissed it pre-PR via `@keydown.esc="cancel"` on a non-persistent `v-dialog`, so the delta is the added button, not the whole contract.)

**Fix:** Derive `blocking` from `noCancel || persistent`, or keep the X but suppress dismissal when `noCancel`.

src/components/dashboard/dialogs/ZwDialog.vue
```diff
@@ -0,0 +1,324 @@
...
+
+function onModel(v: boolean) {
+	emit('update:modelValue', v)
+	if (!v) emit('close')
+}
+function close() {
+	onModel(false)
+}
+
+// VNode so the `header` slot can replace the whole region
```
robertsLando: **[Minor] · Design/API**

`headerVNode` builds the default header with a manual `h()` render function. The comment justifies it as letting the `header` slot replace the whole region — but scoped-slot *fallback content* already does exactly that, in template syntax. This `<style>` block isn't scoped either, so there's no styling motive, and no consumer currently overrides `header`.

**Why:** ~14 untyped `h()` lines carrying a rationale that doesn't hold, in the file that every future dialog is written against.

**Fix:** Move the default header into `<slot name="header" :close="close">…</slot>` as fallback content.

src/components/dashboard/dialogs/ZwWizardDialog.vue
```diff
@@ -0,0 +1,90 @@
+<template>
+	<ZwDialog
+		:model-value="modelValue"
+		:size="size"
+		:persistent="persistent"
+		:loading="loading"
+		:actions="actions"
+		:title="railLayout ? steps[current] : ''"
```
robertsLando, line 4: **[Minor] · Design/API**

The required `title` prop is silently dropped below `TWO_PANE_BREAKPOINT` — `ZwDialogStepProgress` never receives it — and only a subset of `ZwDialog`'s props is forwarded (no `severity`, `icon`, or `blocking`).

**Why:** A required prop that vanishes at one breakpoint, and a wrapper whose capabilities silently shrink relative to what it wraps.

**Fix:** Pass `title` through to the narrow presentation, and forward with `v-bind` rather than hand-copied props.

src/components/dashboard/dialogs/ZwDialog.vue
```diff
@@ -0,0 +1,324 @@
...
+
+const props = withDefaults(
+	defineProps<{
+		modelValue: boolean
+		size?: DialogSize
+		severity?: DialogSeverity
+		title?: string
+		subtitle?: string
+		icon?: Component | null
+		blocking?: boolean
```
robertsLando: **[Minor] · Design/API**

`blocking` and `persistent` encode three meaningful states in two booleans, leaving one combination meaningless.

**Why:** This is the root cause behind several separate findings on this PR (wizard forwards one but not the other; `Confirm` maps only one; the footer Close and the X diverge).

**Fix:** Collapse to a single `dismiss: 'all' | 'button' | 'none'` prop.

src/components/dashboard/atoms/ZwProgressBar.vue
```diff
@@ -0,0 +1,60 @@
...
+	background: var(--zw-accent);
+	transition: width 0.3s ease;
+}
+
+.zw-progress__fill--indeterminate {
+	position: absolute;
+	top: 0;
+	bottom: 0;
+	width: 40%;
+	animation: zw-indeterminate 1.4s ease-in-out infinite;
```
robertsLando, line 54: **[Minor] · DRY/Codebase fit**

Second indeterminate-bar implementation: `ZwActivityReadout.vue:11-21,104-125` already renders track+fill via `Progress.Root`/`Progress.Fill` with its own sweep (`zw-tx-indeterminate`, 1.2s, track `rgba(var(--v0-primary),.18)`), while this hand-rolls `role="progressbar"` plus a new global keyframe (1.4s, `--zw-accent-soft`). Migrated dialogs still also use `v-progress-linear`/`v-progress-circular` (`DialogHealthCheck.vue:108`, `DialogLinkReliability.vue:86`, `DialogNodesManager.vue:272,313`, `SmartView.vue:290`).

**Why:** Two divergent visuals and two a11y implementations for one concept, plus a third from Vuetify still in play.

**Fix:** Wrap `Progress.Root` here — indeterminate is the genuine gap — and have `ZwActivityReadout` consume this atom.

src/components/dashboard/dialogs/ZwDialogClose.vue
```diff
@@ -0,0 +1,43 @@
...
+import { ICON_SIZE, XIcon } from '@/lib/icons'
+
+const emit = defineEmits<{ click: [MouseEvent] }>()
+</script>
+
+<style scoped>
+/* Negative margin pulls the tap target back to the header edge without
+   shrinking it. Scoped rules ride the component's own data-attr, so they
+   still apply after v-dialog teleports the content out of the DOM subtree. */
+.zw-dlg-close {
```
robertsLando: **[Minor] · DRY/Codebase fit**

This consolidates the 17 migrated dialogs but leaves near-identical copies behind: `ZwDeviceDrawer.vue:30-38,197-215` (same `XIcon` at `ICON_SIZE.topbar`, same transparent/6px-padding/hover-tint recipe, its own `:focus-visible` instead of `zw-focus-ring`) and `ZwActivityStrip.vue:30`.

**Why:** N+1 rather than N→1 — and the drawer's hardcoded `rgba(0,0,0,.04)` and `999px` have already drifted from this token-based version.

**Fix:** Point the existing dashboard close buttons at this atom in the same PR.

src/components/dashboard/atoms/ZwButton.vue
```diff
@@ -117,4 +113,27 @@ const emit = defineEmits<{ click: [MouseEvent] }>()
 	font: var(--zw-text-mono-small);
```
robertsLando, line 108: **[Minor] · DRY/Codebase fit**

`danger` is added alongside the existing `destructive`, and `text` alongside `ghost`; `ZwDialog.vue:171-179` maps to only 5 of the now-8 variants.

**Why:** Two overlapping danger variants and two overlapping low-emphasis variants invite arbitrary call-site choices that will diverge visually over time.

**Fix:** Collapse to a tone × emphasis pair, or delete `destructive` if the dialog set supersedes it.

src/components/dashboard/dialogs/ZwDialog.vue
```diff
@@ -0,0 +1,324 @@
...
+/* Non-scoped: v-dialog teleports its content out of this component's DOM
+   subtree, so scoped selectors would not reach it. Everything is
+   namespaced under .zw-dlg to stay isolated. */
+
+/* Content box: width = min(breakpoint, viewport − margin); the specific
+   selector overrides Vuetify's default calc(100% − 48px) cap so the
+   dialog fills the screen with only a small margin on phones. */
+.v-overlay__content.zw-dlg__vcontent {
+	max-width: calc(100vw - 32px);
+	max-height: 90vh;
```
robertsLando: **[Minor] · Readability**

Two different `max-height` values — 90vh on the overlay content here, 86vh on `.zw-dlg` at line 216 — with no explanation for the 4vh delta.

**Why:** Conflicting magic numbers leave the next reader unsure which one is authoritative or whether the gap is intentional.

**Fix:** Unify, or name the gap as a token and say what it's for.

src/components/dashboard/dialogs/ZwDialogStepRail.vue
```diff
@@ -0,0 +1,134 @@
...
+	flex-shrink: 0;
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+	font: var(--zw-text-caption);
+	font-weight: 600;
+}
+
+.zw-rail__num--done {
+	background: rgba(var(--v0-success), 0.12);
```
robertsLando: **[Minor] · DRY/Codebase fit**

`rgba(var(--v0-success), 0.12)` is duplicated verbatim from `tokens.css:118` (`.zw-tone-ok { --tone-bg }`), and again at line 108. `tokens.css:20-27` has `--zw-accent-soft` / `--zw-warn-soft` / `--zw-danger-soft` but no ok-soft sibling.

**Why:** A token value inlined in a component is the drift that the token file exists to prevent.

**Fix:** Add the missing soft token, or apply `zw-tone-ok` as `ZwDialog.vue:160` already does for the header chip.

src/lib/dashboard-types.ts
```diff
@@ -241,3 +241,59 @@ export type DeviceAction =
...
+	| 'text-danger'
+
+export type DialogButtonKind = 'text' | 'filled' | 'outline'
+export type DialogButtonTone = 'accent' | 'danger' | 'neutral'
+
+export interface DialogAction {
+	label: string
+	kind?: DialogButtonKind
+	tone?: DialogButtonTone
+	onClick?: (e: MouseEvent) => void
```
robertsLando: **[Minor] · Design/API**

`onClick` is optional, so `{ label }` alone type-checks — a live, clickable, inert button.

**Why:** The type permits an invalid state on the primitive every future dialog's footer is built from.

**Fix:** Make it a union: handler required unless `disabled: true`.

src/components/Confirm.vue
```diff
@@ -402,15 +405,8 @@ export default {
 				this.reject = this.beforeClose(reject)
 			})
 		},
-		/** Util function to prevent changing options while dialog is still shown */
 		beforeClose(fn) {
```
robertsLando: **[Minor] · Readability**

`beforeClose` is now an identity wrapper that just calls `fn(...args)` — dead indirection left behind by the removed `closeTimeout`.

**Why:** Reads like a hook point that does something.

**Fix:** Remove it and inline the callback, or document why it's retained.

src/components/dialogs/DialogLinkReliability.vue
```diff
@@ -300,14 +276,16 @@ export default {
...
 				this.subscribeChannels(['diagnostics'])
 				this.bindEvent('linkReliability', this.onProgress.bind(this))
-			} else if (open === false) {
-				this.unbindEvents()
-				if (wasRunning) {
-					this.abortLinkReliabilityCheck()
-				}
+			}
+		},
+		teardown() {
```
robertsLando, line 289: **[Minor] · Operability**

Cancellation-on-close moved from the synchronous `modelValue` watcher to `@after-leave`. Same shape in `DialogHealthCheck` (`teardown`) and `DialogNodesManager.vue:966-968`.

**Why:** The abort/stop call now fires only after the leave transition finishes, rather than the instant the user clicks close.

**Fix:** If that delay isn't intentional, fire the abort synchronously on close and defer only the visual/state teardown to `after-leave`.

src/App.vue
```diff
@@ -820,27 +832,59 @@ export default {
...
+			this.loaderText = text || ''
+			this.loaderProgress = 0
+			this.loaderIndeterminate = true
+		},
+		// 0% has no bar to show yet, so fall back to the indeterminate sweep
+		setLoaderProgress(percent) {
+			this.loaderProgress = percent
+			this.loaderIndeterminate = percent === 0
+		},
+		finishLoader(text) {
```
robertsLando: **[FYI] · Security**

`finishLoader(text)`'s argument lands in `loaderText`, which `DialogLoader.vue:9` renders with `v-html`. Nothing is exploitable today — the one caller reproduces the pre-existing OTW string verbatim, and `onControllerStatus` assigns only two literals after a regex match.

**Why:** The name and signature give no hint the parameter is trusted HTML, so a future caller passing a socket- or node-derived string (device name, firmware filename, error text) turns this into stored XSS.

**Fix:** Rename to something like `finishLoaderHtml`, or move the `<span>` wrapper into `DialogLoader` and keep the public helper text-only.

src/components/dialogs/DialogHealthCheck.vue
AlCalzone: would only resetting in init(open) be enough?

src/components/dashboard/dialogs/ZwDialog.vue
AlCalzone: This had something to do with a visual bug that caused the dialog contents to briefly change while closing. A fix needs to ensure this does not get reintroduced.

src/components/dashboard/dialogs/ZwDialog.vue
AlCalzone: The design document had these variants:
image

src/components/dashboard/dialogs/ZwDialog.vue
AlCalzone: Ouch, good catch

src/components/dashboard/atoms/ZwProgressBar.vue
AlCalzone: The dialog _contents_ will be addressed in a followup PR. We should attempt to reuse the atoms in the new dialog chrome though.

robertsLando:
@AlCalzone I would not defer such points to a follow up but implement them here

AlCalzone:
Ok I'll point that PR at the 20-dialogs branch then, so we can merge it first and have it be part of this PR.

AlCalzone:
As discussed, I'd rather land this one first, then continue in a followup PR (already prepared).

PR #4791, opened 2026-08-19
feat: add speed column for control panel

csstup:
Added a speed column to the control panel for each device, capturing the LWR speed in kbps (9.6, 40, 100, 100 LR)

This often helps debugging or figuring out mesh issues.

The value is available in the Network Graph but its much nicer to have it on each device in a summary form.

### DIFF
--- a/src/components/nodes-table/index.vue
+++ b/src/components/nodes-table/index.vue
@@ -310,6 +310,13 @@
 			</div>
 			<div v-else></div>
 		</template>
+		<template #[`item.lwrSpeed`]="{ item }">
+			<rich-value
+				v-if="!item.isControllerNode && !item.virtual"
+				:value="richValue(item, 'lwrSpeed')"
+			/>
+			<div v-else></div>
+		</template>
 		<template #[`item.failed`]="{ item }">
 			<rich-value
 				v-if="!item.isControllerNode && !item.virtual"
--- a/src/components/nodes-table/nodes-table.js
+++ b/src/components/nodes-table/nodes-table.js
@@ -32,6 +32,9 @@ import {
 import { instances, manager } from '../../lib/instanceManager.js'
 import { defineAsyncComponent } from 'vue'
 
+const RATE_LABELS = { 1: '9.6k', 2: '40k', 3: '100k', 4: '100k LR' }
+const RATE_TO_BPS = { 1: 9600, 2: 40000, 3: 100000, 4: 100000 }
+
 export default {
 	props: {
 		socket: Object,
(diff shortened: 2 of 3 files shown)


robertsLando (dismissed):
Seems legit to me, @AlCalzone anything against this? I think we should hide this prop on controllers and virtual nodes BTW

robertsLando:
Thanks for this @csstup, the speed column is a nice addition — the value being buried in the network graph has bugged me too.

I pushed a follow-up commit to your branch (1a29445) since there were a couple of blockers. Happy to explain or revert any of it.

**The column rendered blank.** `VDataTable` only calls `richValue()` from an explicit `#[item.<key>]` cell slot, and there wasn't one for `lwrSpeed` in `src/components/nodes-table/index.vue`. `ManagedItems._getTableHeaderForColumn` does copy `richValue` onto the header object, but nothing consumes it — so the table fell back to `item.lwrSpeed`, which doesn't exist on the node object, and every cell came out empty. Added the slot next to the `protocol` one, guarded with `v-if="!item.isControllerNode && !item.virtual"` like every sibling column. That guard matters on its own: the virtual view reuses the same `nodesProps`, so broadcast/multicast rows and the controller would otherwise show a stray `-`.

**Existing installs never saw the column.** `ManagedItems.initialize()` filters persisted `tableColumns` down to what's already in localStorage and never adds new props; the only path that picks them up is the `clearedN` flag reset. So anyone who had opened the node table before upgrading would get no Speed column *and* no checkbox for it in the Columns menu, recoverable only via "Reset Table". Bumped `cleared3` -> `cleared4`, same as was done when the protocol column landed in 3545.

**Sorting/filtering/grouping were operating on `undefined`.** With no `customValue`, `getPropValue` returns `item['lwrSpeed']` for every node, so the column filter's value list was empty, grouping put everything in one bucket, and the table search never matched. Added `customValue` plus `undefinedPlaceholder`. Also added a `customSort` on the rate enum — sorting the labels as strings orders `100k` before `40k` before `9.6k`.

**Reused the upstream helpers.** `ProtocolDataRate` and `protocolDataRateToString` are already imported from `@zwave-js/core` in `ZwaveGraph.vue` and `NodePanel.vue`, and `protocolDataRateToString` gives exactly the tooltip text the `RATE_TO_BPS` map was reconstructing. Switched the label map to enum members as keys and dropped `RATE_TO_BPS`. The literal `1..4` keys would have silently produced `NaN kbps` if zwave-js ever adds a rate. `maxDataRate` is genuinely in bps so the `/1000` there was correct and stayed.

One cosmetic thing: `align: 'center'` has no effect when `icon: ''` — `RichValue.vue` only applies alignment in the `v-layout` branch. Left the key in place, but the text is left-aligned. Worth fixing in `RichValue` separately rather than here.

Verified with `npm run lint`, `npm run build:ui` and the `ManagedItems` tests.

PR #4796, opened 2026-08-24
fix(api): refresh access control state on User Code CC value changes

### LINKED ISSUE #4786: Kwikset HC620 zwave.set_lock calls from HA show inverted or incorrect result in ZWaveJS-UI
### Checklist

### Deploy method

Docker

### Z-Wave JS UI version

11.22.2

### ZwaveJS version

15.27.0

### Describe the bug

In the ZWaveJS UI, the user access code enabled state for a kwikset HC620 lock appears enabled when it is disabled, and appears disabled when it is enabled. a user code that has never been explicitly enabled or disabled always appears enabled, which can add confusion.

Image

### To Reproduce

use either the ZWaveJS UI or HA developer tools to call set_lock to enable/disable a user access code on a kwikset HC620 lock. if you've never set the value, it will appear enabled. 

Image

Image

Setting this value from HA using something like:
```   actions:
      - action: zwave_js.set_user
        target:
          entity_id: lock.living_room_front_door_lock
        data:
          user_id: 2
          active: "{{ is_state('input_boolean.front_door_guest_code_enabled', 'on') }}"
        response_variable: set_user_result
```
results in the ZWaveJS UI having a reversed sense:

Image

### Expected behavior

I expect the green indicator in ZWaveJS-UI to indicate correctly that the lock user/code is enabled or disabled

### Additional context

_No response_

robertsLando:
Fixes #4786

## The bug

`node.accessControl` is built once at node-ready and afterwards refreshed only on six zwave-js events — `user added`/`modified`/`deleted` and `credential added`/`modified`/`deleted`.

Those events are emitted from exactly two places in zwave-js:

- `lib/node/feature-apis/AccessControl.js` — the high-level `setUser` / `addUser` / `deleteUser` API, and only when its own verification reports success.
- `lib/node/CCHandlers/UserCredentialCC.js` — unsolicited reports, **User Credential CC only**. There is no `CCHandlers/UserCodeCC.js`.

So on a User Code CC lock, every status change that does not go through the high-level API updates zwave-js's value DB but emits only `value updated`:

- Home Assistant's `set_lock_usercode` / `clear_lock_usercode` (raw CC API)
- another controller, or the lock keypad
- a high-level `setUser` on a node without Supervision, where the unsupervised verify path requires `verifiedStatus !== existingStatus` to consider the write successful — the verification GET persists the new status, but no event is emitted

We ignored all of those, so the Access Control tab kept rendering the snapshot taken at interview time. The status dot lags one change behind, which reads as inverted when a code is toggled back and forth, and a slot that was `Enabled` at interview stays green forever.

The reporter's Kwikset HC620 hits the last case on every write: its device config removes Supervision (`zwave-js/packages/config/config/devices/0x0090/hc620.json`), so `setUser` always takes the unsupervised verify path.

## The fix

Refresh from the existing `observedCCProps` value observers for User Code CC (`userIdStatus` and `userCode`), throttled to once per second through the existing `throttle` helper so an interview burst doesn't rebuild the whole state once per slot. The refresh no-ops until the node's first read has happened.

User Credential CC needs nothing — its CC handler already emits the user events on unsolicited reports.

## Not changed

"A code that has never been explicitly enabled or disabled always appears enabled" is inherent to User Code CC: a slot in use is `UserIDStatus.Enabled`, there is no distinct "never configured" state. Only User Credential CC has a real active flag.

## Verification

`tsc --noEmit`, eslint and prettier clean; full test suite passes. Not reproduced against physical hardware — the mechanism was traced from the zwave-js sources listed above.

https://claude.ai/code/session_01HQ7eZNp2LFHJFSCUbh36w7

### DIFF
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -350,9 +350,14 @@ const observedCCProps: {
 		},
 	},
 	[CommandClasses['User Code']]: {
+		userCode(node) {
+			this.scheduleAccessControlRefresh(node.id)
+		},
 		userIdStatus(node, value) {
 			const userId = value.propertyKey as number
 			const status = value.value as UserIDStatus
+
+			this.scheduleAccessControlRefresh(node.id)
 
 			if (!node.userCodes) {
 				return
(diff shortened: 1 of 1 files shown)


AlCalzone:
See https://github.com/zwave-js/zwave-js-ui/issues/4786#issuecomment-5407551748

robertsLando:
should I close this then?

PR #4799, opened 2026-08-25
fix(ui): always show both ping directions in health check results

### LINKED ISSUE #3557: Displaying the failedPingsController attribute
Hi,

Is the failedPingsNode attribute displayed all the time even when it's 0 ? 
I ran 2 tests on the same node and once it appeared, another time it didn't.

I don't have a custom route, and I haven't done any reconstruction in the meantime. Can you explain to me ?

robertsLando:
Closes #3557

## Problem

The `node → controller` row in the health check table only rendered when the driver reported `failedPingsController`. The driver only sets that field when **no** power level was found where all pings were ACKed — i.e. only when things went badly:

```js
// zwave-js Node.ts, checkLifelineHealthInternal
if (powerlevel == undefined) {
    ret.minPowerlevel = Powerlevel["Normal Power"]
    ret.failedPingsController = failedPingsController  // only here
} else {
    ret.minPowerlevel = powerlevel                     // no failed pings field
}
```

So a healthy check (0 failed pings) and a check on a node without Powerlevel CC (nothing measurable) both rendered as *nothing at all*. That's what the reporter saw: the same node showing the row on one run and not the next.

## Fix

Both directions are always rendered now, and each "missing" case says why:

| driver result | before | after |
| --- | --- | --- |
| `minPowerlevel: 2` | *(row hidden)* | `1 ← 102: 0/10` |
| `minPowerlevel: 0, failedPingsController: 2` | `1 ← 102: 2/10` | unchanged |
| `minPowerlevel: undefined` (no Powerlevel CC) | *(row hidden)* | `1 ← 102: not measured` + tooltip |

`minPowerlevel === undefined` is the only case that means "not measurable"; when a level *was* found, every ping at that level was ACKed, so the count is 0.

The min. power level cell also no longer claims `Normal Power` was error-free when `failedPingsController > 0` — it shows `none` (there is no error-free level) with a tooltip.

Route checks get the same treatment per side. There, `failedPingsToSource`/`minPowerlevelTarget` and `failedPingsToTarget`/`minPowerlevelSource` are set together, and a side is skipped when that node lacks Powerlevel CC — or, for the target side, when the source node sleeps. The tooltip names the actual reason.

Rows still render nothing while a round is in progress (guarded on `rating !== undefined`).

The info dialog was updated to match: the stale "Failed pings Controller" entry is folded into the "Failed pings" and "Min Power Level" entries, which now explain `not measured` and `none`.

## Notes

Follows @AlCalzone's read of the semantics in the issue thread, and @gdt's more recent ask — don't hide zeros, say clearly when something couldn't be measured. I kept the two-column layout rather than merging failed pings into the power level column, since the direction labels already disambiguate.

Verified with `tsc` + `vite build` and eslint/prettier. No screenshots: reproducing the `not measured` path needs a real node without Powerlevel CC.

## Second commit: direction notes in the info dialog

From https://github.com/zwave-js/zwave-js-ui/discussions/4798 — the table gives no hint that most columns describe a single direction, so a `1 → 12: 0/10` next to `Max latency: 20 ms` reads as if the latency covered both.

Verified against the driver:

- **Route changes** are counted on the controller → node `NoOperation` ping loop only.
- **SNR margin** is `txReport.ackRSSI - measuredNoiseFloor`, measured by the controller on the ACKs it receives, so it actually describes node → controller.
- **Max latency** is a round trip (`NoOperation` sent → ACK received). Only the controller can start it, but the value itself is independent of the direction — thanks @AlCalzone for the correction, the third commit fixes the wording.

The info dialog now states this, with a short intro explaining that the two directions are measured by different mechanisms (pings vs. a Powerlevel CC node test).

The remaining half of that discussion — `failedPingsController: 10` when the node's Powerlevel CC test never runs, which forces the rating to 0 — is not fixable here and is filed as https://github.com/zwave-js/zwave-js/issues/9127.

### DIFF
--- a/src/components/dialogs/DialogHealthCheck.vue
+++ b/src/components/dialogs/DialogHealthCheck.vue
@@ -150,107 +150,195 @@
 
 						<template #[`item.minPowerlevel`]="{ item }">
 							<strong
+								v-if="item.failedPingsController > 0"
+								class="text-error"
+								v-tooltip:bottom="noErrorFreePowerlevelHint"
+								>none</strong
+							>
+							<strong
+								v-else-if="item.minPowerlevel !== undefined"
 								:class="getPowerLevelColor(item.minPowerlevel)"
-								v-if="item.minPowerlevel !== undefined"
 								>{{ getPowerLevel(item.minPowerlevel) }}</strong
+							>
+							<em
+								v-else-if="item.rating !== undefined"
+								class="text-medium-emphasis"
+								v-tooltip:bottom="
+									noPowerlevelHint(activeNode.id)
+								"
+								>not measured</em
 							>
 						</template>
 
 						<template #[`item.failedPingsNode`]="{ item }">
-							<p
-								class="mb-0"
-								v-if="item.failedPingsNode !== undefined"
-							>
-								{{ resultsTargetNode }} → {{ activeNode.id }}:
-								<strong
-									:class="
-										getFailedPingsColor(
-											item.failedPingsNode,
-										)
-									"
-									>{{ item.failedPingsNode }}/10</strong
-								>
-							</p>
-							<p
-								class="mb-0"
-								v-if="item.failedPingsController !== undefined"
-							>
-								{{ resultsTargetNode }} ← {{ activeNode.id }}:
-								<strong
-									:class="
-										getFailedPingsColor(
-											item.failedPingsController,
-										)
-									"
-									>{{ item.failedPingsController }}/10</strong
-								>
-							</p>
+							<template v-if="item.rating !== undefined">
+								<p class="mb-0">
+									{{ resultsTargetNode }} →
+									{{ activeNode.id }}:
+									<strong
+										:class="
+											getFailedPingsColor(
+												item.failedPingsNode,
+											)
+										"
+										>{{ item.failedPingsNode }}/10</strong
+									>
+								</p>
+								<p class="mb-0">
+									{{ resultsTargetNode }} ←
+									{{ activeNode.id }}:
+									<strong
+										v-if="
+											failedPingsToController(item) !==
+											undefined
+										"
+										:class="
+											getFailedPingsColor(
+												failedPingsToController(item),
+											)
+										"
+										>{{
+											failedPingsToController(item)
+										}}/10</strong
+									>
+									<em
+										v-else
+										class="text-medium-emphasis"
+										v-tooltip:bottom="
+											noPowerlevelHint(activeNode.id)
+										"
+										>not measured</em
+									>
+								</p>
+							</template>
 						</template>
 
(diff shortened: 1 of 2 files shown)


AlCalzone:
> a `1 → 12: 0/10` next to `Max latency: 20 ms` reads as if the latency covered both.

The latency is by definition a round-trip from sending the NoOp to receiving the corresponding ACK. We can only measure it by pinging the device from the controller, but this value should be independent of the direction.

robertsLando:
You're right, thanks — fixed in 0622ba83. The dialog now says latency is measured by pinging the node from the controller, but since it's a round trip (NoOperation sent, ACK received) it applies to both directions. Only the *starting* side is fixed, not the measurement. Route changes and SNR margin keep their direction notes. PR description updated too.

gdt:
> The dialog now says latency is measured by pinging the node from the controller, but since it's a round trip (NoOperation sent, ACK received) it applies to both directions. Only the _starting_ side is fixed, not the measurement.

I see why you say this, but I don't think it's really true.   Long ago, I was doing ad hoc routing (over 802.11 IBSS), and we were measuring ETX by the sender counting transmissions until ACK.    We also had multicast traffic, and at first assumed that the ETX metric was reasonable to use for that.     We found two issues: multicast traffic wasn't acked/retransmitted, which was critical and led to redoing the metric somehow (don't remember, 20 years ago).  The second issue was that there was a driver bug and multicast traffic didn't make it at all.

This experience led me to firmly believe that metrics should measure the traffic that you actually care about, sent in production, rather than something that you believe is a good proxy for that.

While I believe you are right that the ping times from A to B (and B back to A) should match the B to A (and A back to B) times, that's an assumption rather than an observed fact.

Thus, if the measurement is of a ping from A->B->A, it should be displayed that way.  If you can't measure the other round-trip time, that's how it is, and it's good if it appears unmeasured in the UI.

I realize that you're juggling between specialists and normal people struggling to get their networks working, but I find "latency" non-specific, and would prefer "RTT" as the normal abbreviation for round-trip time.   I'm guessing you'll  say no, too confusing, but it would be nice if the help said that crisply.    Looking at the text, it now says "max time to send ping" and that it can either be time to ack vs time to get a response.  I suggest always showing the response time, and if you want to show the tx-report time-to-ack also, add a column.   Putting varying measurements in a cell is confusing.

AlCalzone:
I'm about to head out, so just a short comment. We use RTT in other locations for the actual RTT of production commands like Get-Report flows. I'd really prefer to avoid mixing two measurements under the same umbrella term.

gdt:
(totally understand schedules and async)

I see - so the question is what the RTT is of, so maybe "Ping RTT".  Indeed, RTT of specific commands is going to be different given processing delay.

robertsLando:
@AlCalzone can this be merged?

PR #4815, opened 2026-09-03
fix(mqtt): resolve unknown command class names via zwave-js

### LINKED ISSUE #4803: Window Covering shown als unknownClass_106
### Checklist

### Deploy method

Docker

### Z-Wave JS UI version

11.22.3

### ZwaveJS version

15.27.1

### Describe the bug

The Gateway is configured to use named .topics
Image

I have a new Fibaro FGR-224, his topics related to the Window Covering Class are showing up in an unexpected manner (as unknownClass_106). 

Image

### To Reproduce

-

### Expected behavior

The part "unknownClass_106" in the topic should be replaced by the right name.

### Additional context

_No response_

robertsLando:
Closes #4803

## Problem

`Constants.commandClass()` resolved a CC id to a topic segment through `_commandClassMap`, a hardcoded OpenZWave-era table:

```ts
export function commandClass(cmd: number): string {
	return _commandClassMap[cmd] || `unknownClass_${cmd}`
}
```

The map has **91 entries; zwave-js knows 125 command classes**. Window Covering (`0x6a`) is not in it, so with the NAMED gateway type `Gateway.valueTopic()` built topics like:

```
zwave/<node>/unknownClass_106/endpoint_0/currentValue
```

Unrelated to #4794 — that one only adds the HA discovery config, the topic name comes from here.

## All CCs affected

44 command classes were missing. The ones users actually meet:

`0x53` Schedule, `0x59` Association Group Information, `0x64` Humidity Control Setpoint, `0x6a` Window Covering, `0x6b` Irrigation, `0x6c` Supervision, `0x6d` Humidity Control Mode, `0x6e` Humidity Control Operating State, `0x6f` Entry Control, `0x74` Inclusion Controller, `0x78` Node Provisioning, `0x83` User Credential, `0x9f` Security 2.

Plus the Z/IP, metering-tariff, HRV, prepayment and authentication classes nobody is likely to hit.

## Change

Fall back to `getCCName()` from `@zwave-js/core`, snake_cased, when the CC is not in the map:

```
commandClass(0x6a) -> "window_covering"   // was "unknownClass_106"
commandClass(0x9f) -> "security_2"
commandClass(0x26) -> "switch_multilevel" // unchanged, legacy name wins
commandClass(-1)   -> "unknownClass_-1"   // unchanged
```

`getCCName()` returns `unknown (0x24)` for ids it doesn't know, so that case still produces `unknownClass_<id>`.

### The map keeps only the names that actually differ

It can't just be replaced by `getCCName()` — **42 of its 91 names are OZW names that disagree** with the zwave-js ones:

| map | zwave-js |
| --- | --- |
| `switch_binary` | `binary_switch` |
| `switch_multilevel` | `multilevel_switch` |
| `sensor_multilevel` | `multilevel_sensor` |
| `multi_instance` | `multi_channel` |
| `firmware_update_md` | `firmware_update_meta_data` |
| `chimney_fan` | *(zwave-js doesn't know 0x2a)* |
| ... (36 more) | |

Dropping those would rename topics on nearly half of every existing install.

The other **49 entries resolved to exactly the same string** `getCCName()` produces (`no_operation`, `battery`, `central_scene`, `door_lock`, `thermostat_setpoint`, ...), so they were dead weight and are gone. The map is now the 42 real overrides.

### Guard against upstream renames

Reducing the map means 49 names now come from zwave-js, and a rename there would silently change a user's topic. `test/lib/Constants.test.ts` pins **all 91 original names** — one assertion each — so that fails CI instead. The fix if it ever trips is to add the CC back to `_commandClassMap`, not to update the fixture; the comment above it says so.

Verified the guard bites: flipping one fixture entry fails the run.

## Breaking change for FGR-224 owners

Anyone already running a device with an unmapped CC sees the topic change, e.g. `unknownClass_106` -> `window_covering`. That is the point of the fix, but their existing automations/subscriptions on the old topic will need updating.

## Tests

Three behaviour cases (legacy name wins, unmapped CC falls back, punctuation normalized) plus the 91-name lock-in fixture. Full suite: 312 passed.

Not verified on real hardware — I don't have an FGR-224.


node_modules
AlCalzone, line 1: I don't think you wanted to check this in

AlCalzone (changes requested):

robertsLando:
I was going to fix it before claude stopped working

PR #4818, opened 2026-09-04
feat: local names for access-control users

werddomain:
## Summary
- Access-control users only expose the device's own `userName` (part of the User Credential CC) — there was no way to label a user without writing that label back to the lock. Adds `accessControlSetUserLocalName` and a `localName` field, persisted locally in `store/nodes.json` (same mechanism already used for node name/location) and never sent to the device. The name is cleaned up automatically when a user is deleted.
- Adds an inline editor in the Users list (chip + text field, matching the existing admin PIN inline-edit pattern already in this component) to set/edit that local name without opening a dialog.

## Test plan
- [x] `npx tsc --noEmit -p tsconfig.json` — no errors
- [x] `npx eslint` on all changed files — no errors
- [x] `npx vitest run` — 310/320 passing; the 10 failures are pre-existing symlink/unix-socket tests that need privileges unavailable on this Windows dev machine, unrelated to this change
- [x] Added unit tests in `test/lib/ZwaveClient.test.ts` covering `accessControlSetUserLocalName`, and the local-name cleanup in `accessControlDeleteUser` / `accessControlDeleteAllUsers`
- [ ] Manual UI verification against a real/mock device (not performed in this environment)

### DIFF
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -321,6 +321,7 @@ export const allowedApis = validateMethods([
 	'accessControlCancelCredentialLearn',
 	'accessControlGetAdminCode',
 	'accessControlSetAdminCode',
+	'accessControlSetUserLocalName',
 ] as const)
 
 export type ZwaveNodeEvents = ZWaveNodeEvents | 'statistics updated'
@@ -607,7 +608,10 @@ export interface ZUIScheduleConfig<T> {
 	slots: ZUISlot<T>[]
 }
 
-export type ZUIAccessControlUser = UserData
+export type ZUIAccessControlUser = UserData & {
+	/** Local-only friendly name, stored in zwave-js-ui and never sent to the device. */
+	localName?: string
+}
 
 export type ZUIAccessControlCredential = {
 	userId: number
@@ -734,6 +738,8 @@ export type ZUINode = {
 		enabled: number[]
 	}
 	accessControl?: ZUIAccessControl
+	/** Local-only access-control user names, keyed by endpointIndex then userId. Never sent to the device. */
+	accessControlUserNames?: Record<number, Record<number, string>>
 	defaultTransitionDuration?: string
 	defaultVolume?: number
 	protocol?: Protocols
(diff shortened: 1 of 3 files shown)


robertsLando:
@AlCalzone may be worth adding a support for this on driver?

AlCalzone:
I'm not sure I fully understand the use case. Why do you need local names that differ from the ones stored on-device?

werddomain:
My device doesn't store names locally; it only uses integers. From what I've read, I would need the device's gateway to view or create IDs. In the JS UI, I only see IDs (1 to 99) without names, and I don't plan to buy or use the gateway. It seems the lock maps the ID to the name in the cloud... Because I can never remember if my Housekeeper is ID 32 or 33, I always have to check a spreadsheet. I built this for myself and thought I would share it.

I can update it based on your requirements. Since I only use this integration for a single device, maybe I missed an existing feature, or perhaps this might conflict with something else?

AlCalzone:
Oh, so you have a lock that uses User Code CC, not User Credential CC? I didn't grasp that from the PR body.

We could maybe do something like we already do for the node name and location:
https://github.com/zwave-js/zwave-js/blob/7b5361cf4073e0fa80aa9021c0f8eb02161c9156/packages/zwave-js/src/lib/node/Node.ts#L372-L412 (plus a few more code paths)

...except through the access control API. If the device supports it, just store the name on the device. If not, reuse the internal User Credential CC value ID to store it in zwave-js's cache.

I'll have to think about it though. It means we would have to fake support for the naming feature in the capabilities, which could mislead applications.

werddomain:
> Oh, so you have a lock that uses User Code CC, not User Credential CC? I didn't grasp that from the PR body.
> 
> We could maybe do something like we already do for the node name and location: https://github.com/zwave-js/zwave-js/blob/7b5361cf4073e0fa80aa9021c0f8eb02161c9156/packages/zwave-js/src/lib/node/Node.ts#L372-L412 (plus a few more code paths)
> 
Yes, I used the same mechanism that is already used for the node name and location.

> ...except through the access control API. If the device supports it, just store the name on the device. If not, reuse the internal User Credential CC value ID to store it in zwave-js's cache.
>
I'm not currently familiar with how to check if the device supports it, but I can look into it.
 
> I'll have to think about it though. It means we would have to fake support for the naming feature in the capabilities, which could mislead applications.
>
To avoid confusion, if we allow renaming it in the UI, we could call it something like `Friendly Name`. Let me know what you decide!
Meanwhile, I will investigate how to send the name to the device if it happens to support it.

AlCalzone:
Here is a good starting point: https://zwave-js.github.io/zwave-js/#/api/endpoint?id=credential-management

PR #4824, opened 2026-09-13
fix(api): do not let a stuck driver shutdown block the client restart

adamjernst:
## Problem

When the driver emits a `Driver_Failed` error, the client schedules a restart. `restart()` calls `close()`, which awaits `this.server.destroy()` and `this._driver.destroy()` with no bound. If either never settles, the restart hangs forever and the client never reconnects.

I observed this with zwave-js 15.27.0 and an `esphome://` connection. After a network outage the driver gave up reopening the port and started destroying itself, but its `destroy()` never completed (the `driver instance destroyed` log line never appeared and the instance's timers kept firing for two more days). The client stayed stuck in `close()` and no new driver was ever created until the container was restarted. See zwave-js/zwave-js#9249 for the driver-side logging change.

## Change

- Add `settledWithin()` to `utils.ts`, which waits for a promise for at most a given time and reports whether it settled. A rejection is propagated when it happens in time and ignored afterwards so it cannot become an unhandled rejection.
- In `close()`, wait at most 60 seconds each for the server and the driver to shut down. When a step times out, log an error naming it and continue, leaking the stuck instance instead of blocking the restart.
- Unit tests for the helper.

The 60 second bound is longer than the 30 second watchdog proposed in the driver PR, so the driver's warning naming the stuck step arrives in the log before the client gives up on it.

### DIFF
--- a/api/lib/ZwaveClient.ts
+++ b/api/lib/ZwaveClient.ts
@@ -175,6 +175,10 @@ export const configManager = new ConfigManager({
 
 const logger = LogManager.module('Z-Wave')
 
+// How long close() waits for the server and the driver to shut down before
+// carrying on without them
+const CLOSE_STEP_TIMEOUT = 60_000
+
 const NEIGHBORS_LOCK_REFRESH = 60 * 1000
 
 /** Maximum length of a multicast group name (avoids bloating MQTT topics). */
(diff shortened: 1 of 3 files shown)


robertsLando (commented):
I'm not sure, I think the best way to recover from this would be closing the application as for example an half open serial would still block consecutive connection so while this could fix some issues I'm scared it just hide others.

cc @AlCalzone maybe something could be done on driver to prevent stalled close?

AlCalzone:
I'm looking into eliminating sources of these driver stalls. Applications should not need to work around this.

adamjernst:
Cool, I'll take my efforts to https://github.com/zwave-js/zwave-js/pull/9249 in the hope that can help identify the causes of the stalls.

